PR #5, opened 2016-05-13
Added test environment using pytest

ogenstad:
As I mentioned in the chat I created a new testing environment with more thorough tests which tests more scenarios than the current tests. It's also quite a lot quicker and it's easier to see what is wrong when the tests fail.

I made some changes to tox.ini, since this isn't currently used by Travis it shouldn't impact anything else.

To test how this works you can just try this in the ntc-template root:
```bash
pip install tox
tox
```

Currently there are a few issues which these tests pick up that the current tests didn't. Such as the "show interfaces status" template which was fixed yesterday and the "show ip ospf neighbor" template. 

Also there are some duplicate lines in one of the reference files. Other parsed files are not in the same order as the actual data. In order to fix this the tests could sort the lists before comparing, or the parsed data could be changed.

I also made some changes to tox.ini to make it easier to run the Ansible tests without caring about the environment. You can just run tox and it will setup the requirements. However currently the Ansible tests are run after the pytest tests and since the pytest tests fail due to the incorrect data in the templates and parsed files the Ansible test is never run.

It this PR is accepted and it's desired to keep the Ansible tests I would propose that the code for that is changed so that it also uses pytest. Then the Ansible tests would be run even if the other tests fail.

Some tests are a bit redundant now, this is mostly to make it easier for contributors to spot what the actual error is. Another benefit of this is that it will be easier for contributors to run their own tests before they submit a PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -183,3 +183,5 @@ ENV/
 # Rope project settings
 .ropeproject
 
+# Cloned modules used while testing
+.ntc-modules/
--- a/tests/test_structured_data_against_parsed_reference_files.py
+++ b/tests/test_structured_data_against_parsed_reference_files.py
@@ -1,0 +1,137 @@
+"""Run tests against all the *.raw files."""
+import glob
+import pytest
+import yaml
+try:
+    import clitable
+except:
+    import textfsm.clitable as clitable
+
+
+def return_test_files():
+    """Return a list of all the *.raw files to run tests against."""
+    platform_dirs = glob.glob('tests/*')
+
+    platforms = []
+    for platform in platform_dirs:
+        platforms.append(glob.glob('%s/*' % platform))
+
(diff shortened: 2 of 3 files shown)


ogenstad:
Added a small fix:

rename tests/{test_structured_data_aginst_parsed_reference_files.py => test_structured_data_against_parsed_reference_files.py}

ogenstad:
Rebased to resolve merge conflicts.

GGabriele:
Great job @ogenstad !
Can you aso fix docstrings? Some of them seem to be duplicated.

ogenstad:
Thanks. :) Sure I'll take a look!

ogenstad:
@GGabriele, I've rebased to fix the doc strings. Let me know if you need anything else.

PR #19, opened 2016-06-13
New template: Cisco IOS XR - show configuration commit list

ogenstad:


ogenstad:
Rebase with changed names of sample files so that Ansible tests pass.

PR #21, opened 2016-07-07
Update Show Version Template

itdependsnetworks:
Please Note this changes serial and hardware to lists. Since there can be multiple devices, e.g. stackable switches or VSS pairs.


jedelman8:
re-submit this after we talk live (after current project)

PR #31, opened 2016-08-16
adding additional hp_comware templates and supporting files

netmanchris:


jedelman8:
heads up, this one is failing and by the way, can you do separate PRs for the docs?  Just use branches on your side :)

netmanchris:
I might back out the latest PR ( lldb_neighbor_verbose )  until you have a few minutes to look at it.  I THINK I might be hitting the first match vs. best match thing that we talked about before, but I’m not sure. I’ll hack away a bit at it this morning, but if it doesn’t work, I’ll move on to the other templates that don’t have any overlap.  

Looking at the output, it looks like the test is using the raw from the lldp-neigh command and not going to the lldp-neigh-verbose command.     LLDP-neigh test also appears first in the logs so I'm pretty sure I'm hitting the first match vs. best match, but don't understand the testing internals well enough to figure out why.

I'll pull out the LLDP-Neighbor-Verbose template and tests until you can make time for a webex to figure this out. No rush.

Thanks!

PR #33, opened 2016-08-17
Adding cisco_nxos_show_fex_id

GGabriele:


jedelman8:
It looks like you have this below `cisco_nxos_show_fex` but we need more specific commands to be above  others, etc.  I'll make the change, but just an FYI.

PR #36, opened 2016-10-10
Pinning ansible version

GGabriele:


jedelman8:
@GGabriele Why did you pin this?  We need to update the test script before this? :)  Current script still uses 1.9.2!

PR #40, opened 2016-10-20
Fix for "show ip ospf neighbor" and removal of old tests from tox.ini

ogenstad:
The current template for Cisco IOS with the reference data doesn't work and the old tests using Ansible doesn't pick this up. I've added done a small change to the template and added a new pair of files for reference and parsed data. These won't be run with the old tests but the py.test one loads all of the files.

I'm not actually sure about the validity of the data in the old reference file with two spaces after `/` here: `FULL/  -`. I just left that part as it was. Instead I edited the parsed reference which didn't have the extra space. It could be that the reference data was edited before it was submitted. Let me know if you want me to change the reference data instead.

Ps. As a next step I would vote for removing the Ansible tests as it's easier to see what's going wrong with the new tests, it's possible to test more reference files and they run much quicker.


ogenstad:
I see that the current tests fail because of how it tries to find the data for the test. 

What do you say about changing .travis.yml so it uses pytest instead?

jmcgill298:
```
ogenstad [4:54 PM]
One remaining issue is that the tests now fail due to the fact that the test framework can’t automatically find the tests for vyos. This looks like it’s due to the wildcard at the bottom of the index file for vyatta_vyos. Should that wildcard be there?


mcgoo298 [4:30 PM]
I'm assuming the theory is that there are several vyos implementations that should still apply regardless of the vendor using it. But that is something I should look into.
Thanks for the rebase and updates. I will try to take a look by the weekend.

ogenstad [4:54 PM]
Ok, for now I just added that one and a few others to a list of checks that are ignored. That part is only for a check to validate that test cases exists for all entries in the index file. The actual tests are still run against those. Can try to come up with some other solution to verify that tests exist for things like the vyos one. Can perhaps to a file search if there’s a `*` character in the platform instead of doing an exact match.
```

PR #42, opened 2016-11-01
Add Alcatel Nokia SROS show service id base template

scamallnetworks:


GGabriele:
Hi @scamallnetworks, thanks for this PR.

Can you modify it to include all your test files into a new directory called `alcatel_ros` in the `tests` dir?
You should also update the `templates/index` file.

scamallnetworks:
Hi Gabriele

I don't see the folder, am I to create it?

Keith

GGabriele:
Yes, feel free to create `alcatel_ros` yourself :)

scamallnetworks:
Oooh travis didn't like that much

GGabriele:
Maybe it's better to try to solve these issues before pushing more templates. I'd suggest to test it locally first one template at time by just running `py.test` on the main folder.

scamallnetworks:
Travis is rejecting legitimate output because there are unquoted : in the output.  This is part of the way the router outputs information.  Yesterday I changed port:tag in one template so that the output would consider port and tag different keys.  Thats fine in this context but VPNV4 BGP uses a route distinguisher which is value:value and splitting this is a kludge that shouldn't be done.  Route targets as well will cause this issue.  For many vendors they are target:value:value or simply value:value for others. Any suggestions?

GGabriele:
Here there are few other things that need to be fixed first:
- for commands with a `-`, all related files/directories must contain that `-`. For example, `show_router_bgp_routes_vpn_ipv4.template` should be `show_router_bgp_routes_vpn-ipv4.template` instead, since the command is `show router bgp router vpn-ipv4`
- tests should reside in `tests/<vendor>/<command>/`...here you have `tests/alcatel_sros/show_router_bgp_routes_vpn_ipv4.parsed` instead for example. Take a look at existing tests for a guidance.

You can test your templates (and related files) locally first to makes sure all tests pass. :)

scamallnetworks:
Hey Gabriele, excuse my ignorance. How would I test this myself?

GGabriele:
You can test all the templates by just running `py.test` on the main folder.

GGabriele:
Closing this as per #66

PR #76, opened 2017-01-12
Fixed test-template.yml to run in Ansible 2.3

### LINKED ISSUE #73: Issues with trying to use the test-template playbook
##### ISSUE TYPE

 - Bug Report

##### COMMAND 
```
ansible-playbook  test-template.yml
-- ansible 2.3.0 (devel 07a517aebd) last updated 2017/01/09
```

##### SUMMARY

I was working on some template improvements and wanted to add the tests in.  I ran into a couple of issues.

1) test-template.yml does not work out of the box.  Perhaps I'm missing something, but it seems to not work with ansible 2.3 dev branch.  Starting with I can't see where TEST_DIR was defined, | split('_') balks, and with_items on the last task seemed to be in the wrong spot.  I was able to work through these and get it running with the diff below (Also simplified the prompting a bit with new var setup).

```diff
--- a/test-template.yml
+++ b/test-template.yml
@@ -7,30 +7,27 @@
   vars_prompt:
     - name: template_name
       prompt: "Enter the name of the template to verify:"
-      default: "cisco_nxos_show_ip_route"
-      private: no
-
-    - name: platform
-      prompt: "Enter the platform"
-      default: "cisco_nxos"
+      default: "cisco_ios_show_snmp_community"
       private: no

+  vars:
+    top_dir: "{{ lookup('env', 'PWD') }}"
+    TEST_DIR: "{{ top_dir }}/tests/"
+    platform: "{{ template_name.split('_')[0:2] | join('_') }}"
+    command_variable: "{{ template_name.split(platform + '_') | last }}"
+    command: "{{ command_variable.split('_') | join(' ') }}"
+
   tasks:

-  - name: extract command from template name
-    set_fact: command_variable="{{ template_name | split(platform + '_') | last }}"
-
-  - name: extract command string from command_variable
-    set_fact: command="{{ command_variable | split('_') | join(' ') }}"
-
   - name: run TextFSM locally on test input
     ntc_show_command:
       connection: offline
       file: "{{ item }}"
       platform: "{{ platform }}"
       command: "{{ command }}"
+      template_dir: "{{ top_dir }}/templates"
     with_fileglob:
-      - "{{ TEST_DIR }}/{{ platform }}/{{ command_variable}}/{{ template_name}}*.raw"
+      - "{{ TEST_DIR }}/{{ platform }}/{{ command_variable }}/{{ template_name }}*.raw"
     register: ntc_result

   - name: read parsed sample files
@@ -39,7 +36,7 @@
     register: ntc_result

   - name: verify that parsed result is the same as expected
+    with_items: "{{ ntc_result.results }}"
     compare_dict:
       result: "{{ item.item.response }}"
       sample: "{{ item.ansible_facts.parsed_sample }}"
-    with_items: "{{ ntc_result.results }}"
```
2) In addition to the above, once I did get it working I was not able to do negative testing with incorrect values on the parsed vs raw samples.  After digging a little further I found that compare_dict was not doing the job 100%.  I'm not sure if things changed somewhere in ansible, but the dicts are being passed in as strings and it appears to be breaking the string into characters to compare the expected results contains the same set of ascii characters (not necessarily in the right order!)  I was able to work out a patch for this as well, but I wanted to mention it before hand.  As that code is also in a separate repo, probably should just discuss there.  test-templates.py has a similar compare function in this repo, but I wasn't able to get that to run either.

3) after I succeeded with the above, ran into a case where I wanted to introduce a new variable (for ios show snmp) as entries can optional have an ACL apply, but then I wasn't sure how the template verification should handle "optional" record variables.  I think gTextFSM probably shouldn't be returning them if they were not parsed out of the input (or at least differentiate undefined from ""), but it not clear to me if you can make vars optional.  I guess I can discuss this one separately later, but just wanted to mention it.

bdowling:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

test-template.yml

##### SUMMARY

Changed the variable setup so it only prompts for template name and
creates the platform from that argument.  Fixed TEST_DIR to default
to below the current working directory.  Assumes playbook run from
ntc-templates folder.

Closes #73

```

```

### DIFF
--- a/test-template.yml
+++ b/test-template.yml
@@ -10,36 +10,33 @@
       default: "cisco_nxos_show_ip_route"
       private: no
 
-    - name: platform
-      prompt: "Enter the platform"
-      default: "cisco_nxos"
-      private: no
+  vars:
+    top_dir: "{{ lookup('env', 'PWD') }}"
+    TEST_DIR: "{{ top_dir }}/tests"
+    platform: "{{ template_name.split('_')[0:2] | join('_') }}"
+    command_variable: "{{ template_name.split(platform + '_') | last }}"
+    command: "{{ command_variable.split('_') | join(' ') }}"
 
   tasks:
 
-  - name: extract command from template name
-    set_fact: command_variable="{{ template_name | split(platform + '_') | last }}"
-
-  - name: extract command string from command_variable
-    set_fact: command="{{ command_variable | split('_') | join(' ') }}"
-
-  - name: run TextFSM locally on test input
+  - name: "{{template_name}}: run TextFSM on test input"
     ntc_show_command:
       connection: offline
       file: "{{ item }}"
       platform: "{{ platform }}"
       command: "{{ command }}"
+      template_dir: "{{ top_dir }}/templates"
     with_fileglob:
-      - "{{ TEST_DIR }}/{{ platform }}/{{ command_variable}}/{{ template_name}}*.raw"
+      - "{{ TEST_DIR }}/{{ platform }}/{{ command_variable }}/{{ template_name }}*.raw"
     register: ntc_result
 
-  - name: read parsed sample files
+  - name: "{{template_name}}: read parsed sample files"
     include_vars: "{{ item.item | regex_replace('\\.raw$', '.parsed') }}"
     with_items: "{{ ntc_result.results }}"
     register: ntc_result
 
-  - name: verify that parsed result is the same as expected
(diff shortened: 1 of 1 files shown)


bdowling:
Well, tests pass in **Ansible 1.9.2** as well.  Didn't realize the CI was using this older version -- perhaps that was my problem in #73?  What are the plans to update?

itdependsnetworks:
There has been discussion and a pending PR to help with testing.

PR #108, opened 2017-03-28
Add GigX/X/X support and Duplex a-half interface support.

jasonbarbee:
Fix - Add GigX/X/X support - Catalyst Stack Support, like 2960X.
Fix - 2950s report half duplex as “a-half”.
Added variety of test results to the raw parser test.

### DIFF
--- a/templates/cisco_ios_show_interfaces_status.template
+++ b/templates/cisco_ios_show_interfaces_status.template
@@ -1,5 +1,5 @@
 Value PORT (([a-zA-Z])\w+/\d+|([a-zA-Z])\w+/\d+/\d+|mgmt\d+|Po\d+|Lo\d+|Vlan\d+)
-Value NAME (\S+((\s\w+|\w+'|\w+'\w+|'\w+|\w+\s+)+)?)
+Value NAME (.{0,18}?)
 Value STATUS (\w+)
 Value VLAN (\w+)
 Value DUPLEX (a-full|auto|half|a-half)
--- a/templates/cisco_ios_show_interfaces_status.template
+++ b/templates/cisco_ios_show_interfaces_status.template
@@ -2,7 +2,7 @@ Value PORT (([a-zA-Z])\w+/\d+|mgmt\d+|Po\d+|Lo\d+|Vlan\d+)
 Value NAME (\S+((\s\w+)+)?)
 Value STATUS (\w+)
 Value VLAN (\w+)
-Value DUPLEX (a-full|auto|half)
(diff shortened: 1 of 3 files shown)


GGabriele:
Just run `pytest` on the main folder and I saw that the current template is failing with your new raw data. Would you mind taking a look please?

jasonbarbee:
Sorry, I had fully run a test on it before a last minute change. At the last minute I decided about removing Fa0 management interface out of this PR. Which is not included in this PR.
The Parser does not include "fa\d+", I added then I I took it out but forgot to modify the raw file. But overall, it's not a common use case, and depending on the usage, may have unintended consequences, at least it did for me when I ran it. I'm open for more comment on that. It's easy to add in, but if people are using it, and bulk changing interfaces identified through this script it could break the management access.

GGabriele:
I think the problem is actually that you're modifying parsed and raw files but not fixing the template accordingly (or simply not including it in this PR :) )

I've just used this template and worked with and without the Fa0 interface

```
Value Required PORT (\S+)
Value NAME (\S+((\s\w+)+)?)
Value STATUS (\w+)
Value VLAN (\w+)
Value DUPLEX (a-full|auto|half|a-half)
Value SPEED (auto|\w+-\d+|\d+)
Value TYPE (.*)

Start
  ^${PORT}\s+${NAME}\s+${STATUS}\s+${VLAN}\s+${DUPLEX}\s+${SPEED}\s+${TYPE} -> Record
  ^${PORT}\s+${STATUS}\s+${VLAN}\s+${DUPLEX}\s+${SPEED}\s+${TYPE} -> Record
```

jasonbarbee:
I see the failure, I will try to run the test script on my side and post an update. I ran into a new change today and tried to work it in.
The current codebase failed to parse descriptions with ' in them, or at least in certain cases.

itdependsnetworks:
I think this may be a bit cleaner, let me know what you think?

```
Value PORT (\S+)
Value NAME (.{0,18}?)
Value STATUS (\w+)
Value VLAN (\w+)
Value DUPLEX (a-full|full|auto|half|a-half)
Value SPEED (auto|\w+-\d+|\d+)
Value TYPE (.*)

Start
  ^${PORT}\s+${NAME}\s+${STATUS}\s+${VLAN}\s+${DUPLEX}\s+${SPEED}\s+${TYPE}$$ -> Record
```

jasonbarbee:
Excellent - much more readable. I tested it locally, and have added it to the PR.

GGabriele:
Looks good to me, thanks @jasonbarbee !

PR #132, opened 2017-07-15
Updated IOS show-ip-route to handle additional cases

jamiecaesar:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_ios_show_ip_route.template, cisco_ios, show ip route

##### SUMMARY

Modifies the the Cisco IOS "show ip route" template to account for the following scenarios:
* Less common interface names that include: a dash (Port-channel123), a colon (Serial0/0/0:0), or a period (FastEthernet0/0.100)
* Recognizes routes that are split across lines, such as when the Protocol, Type, Network and Mask are on one line and the rest of the route is on one or more follow-up lines (examples added to tests, which were sanitized from a real route table).  Covers both single and ECMP cases.
* Made the Protocol regex more generic since not all possible values were being matched.
* Replaced [0-9] with \d since they are equivalent (and I was already changing other parts)

```

```

### DIFF
--- a/templates/cisco_ios_show_ip_route.template
+++ b/templates/cisco_ios_show_ip_route.template
@@ -1,31 +1,43 @@
-Value Filldown Protocol (C|S|R|B|D|O|i|L)
+Value Filldown Protocol (\w)
 Value Filldown Type (\w{0,2})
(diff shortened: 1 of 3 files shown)


jamiecaesar:
Found an additional issue with the uptime.  Sometimes it is in the 00:00:00 format, and sometimes in the 5w3d type format.   I'll re-submit after fixing that issue.

PR #133, opened 2017-07-16
Updated IOS show-ip-route to handle additional cases

jamiecaesar:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT
cisco_ios_show_ip_route.template, cisco_ios, show ip route

##### SUMMARY

Modifies the the Cisco IOS "show ip route" template to account for the following scenarios:

- Updated template Value names to be all caps like is mentioned in the instructions for contributions.
- Updated raw output file to NOT include prompt or command (as mentioned in contributions section)
- Match on less common interface names that include: a dash (Port-channel123), a colon (Serial0/0/0:0), or a period (FastEthernet0/0.100).  Modified test output to have at least one of each type of interface.
- Recognizes routes that are split across lines, such as when the Protocol, Type, Network and Mask are on one line and the rest of the route is on one or more follow-up lines (examples added to tests, which were taken from a real output and sanitized). Covers both single and ECMP cases.  Example also added below.
- Made the PROTOCOL regex more generic.  Older routers still have EGP as an option, newer routers have NHRP and LISP characters that weren't in the initial list.
- Replaced [0-9] with \d since they are equivalent and it shortens the Value string a bit.
- Updated the UPTIME value to recognize both the 00:00:00 format and the 7w3d type format.  Modfied a route in the test case to look for this.

```
Sanitized Route Entry that is broken across lines by IOS due to length:
O E2    10.63.184.0/23 
           [110/20] via 10.62.4.29, 12:55:19, TenGigabitEthernet1/15
           [110/20] via 10.62.3.29, 12:55:19, TenGigabitEthernet1/16

```


itdependsnetworks:
Since this is not backwards compatible, can you remove the slash form the cidr notation as well?

jamiecaesar:
Done (other than waiting on CI checks).  Anything else you'd like tweaked while this is open?

jamiecaesar:
Just added another update to handle some different types of null routes (both in the template and tests).  Such as:
```
B        11.1.0.0/17 [200/0], 2w0d, Null0
S        12.0.0.0/16 is directly connected, Null0
O        13.14.128.0/17 is a summary, 7w0d, Null0
```

itdependsnetworks:
Looks good to me, any inputs @GGabriele ?

GGabriele:
Nice job @jamiecaesar

PR #158, opened 2017-10-21
Alex up

itdependsnetworks:
@alexanderhofmann 

I took one of your larger commits and fixed a bunch what I saw to make tests pass. Not sure how to deal with the `-y` flag and the folder name, so I ignored that for now. 

A few notes:
 - Always run tests simply type `py.test` in the repo directory (have to have appropriate installed of course)
 - Order matters in the index
   - We ran into issues where `show ip ospf` matched for `show ip ospf database`, so watch for that in the tests
 - If a column doesn't match, it still produces a key, thus it needs to be included in the tests.
 - Not sure how/why you have some many different people's commit's in your PR, but I think best to start from the PR, and retro fit

I still may clean this up a bit, but like I said, if you are ok with this, I think it makes sense to rebase from here, and then add back all of your changes, testing as you go along.


alexanderhofmann:
Somehow i messed up the rebase...
But thank you very much for the tips.

PR #167, opened 2017-12-13
added show vrf and show ip arp for nxos

vaneuk:
added cisco_nxos_show_vrf and cisco_nxos_show_ip_arp

##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_nxos_show_ip_arp.template, cisco nxos, show ip arp
cisco_nxos_show_vrf.template, cisco nxos, show vrf

##### SUMMARY

```

```


itdependsnetworks:
well done!

I'm trying to make it a standard moving forward to add these to each template: 

```
  ^\s+$$
  ^.* -> Error "LINE NOT FOUND"
```

This way if something has not known, it will fail. Having a few use cases where not finding issues that happen around 1% of the time, and hard to realize there is even an issue. Do you mind adding?

vaneuk:
Hi Ken,
I think i got your point. Makes sense.

vaneuk:
@itdependsnetworks i tested it a little bit more - shouldn't we add 
`^\s*$$`
instead of
`^\s+$$`
?

otherwise there will be errors on empty newlines.

itdependsnetworks:
Yes, I would actually prefer:
```
^\s+$$
^$$
```
Just to be more explicit. 

On a side note, many of us hang out here: http://slack.networktocode.com/

if you are doing any network automation, it's a good resource.

PR #168, opened 2017-12-13
Handle out-of-bound case where IOS splits lines

dagwieers:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
templates/cisco_ios_show_cdp_neighbors.template

##### SUMMARY
So the output we got was this, which was not handled with the current textfsm template.
```
Capability Codes: R - Router, T - Trans Bridge, B - Source Route Bridge
                  S - Switch, H - Host, I - IGMP, r - Repeater, P - Phone,
                  D - Remote, C - CVTA, M - Two-port Mac Relay
Device ID        Local Intrfce     Holdtme    Capability  Platform  Port ID
switch01.example.com
                 Ten 2/2/8         179             R S I  WS-C4510R Ten 6/1
switch02.example.com
                 Ten 1/2/8         169             R S I  WS-C4510R Ten 5/1
```

### DIFF
--- a/templates/cisco_ios_show_cdp_neighbors.template
+++ b/templates/cisco_ios_show_cdp_neighbors.template
@@ -1,15 +1,15 @@
 Value NEIGHBOR (\S+)
 Value LOCAL_INTERFACE (\S+\s\S+)
-Value CAPABILITY (([A-Z]\s)*?[A-Z]??)
+Value CAPABILITY (([A-Z]\s)*?[A-Z]?)
 Value PLATFORM (\S+\s\S+|\S+)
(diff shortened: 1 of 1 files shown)


itdependsnetworks:
@dagwieers can you update the tests to reflect this use case?

Also: 

I'm trying to make it a standard moving forward to add these to each template: 

```
  ^\s+$$
  ^.* -> Error "LINE NOT FOUND"
```

This way if something has not known, it will fail. Having a few use cases where not finding issues that happen around 1% of the time, and hard to realize there is even an issue. Do you mind adding?

jmcgill298:
#235

PR #177, opened 2018-02-11
Fix for VLANs with 4 digits

wmann:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_ios_show_mac-address-table

##### SUMMARY
This is fix for show mac address table on Cisco IOS. In the existing template it is looking for one or more spaces in front of the VLAN ID. But VLANs with 4 digits don't have any space before them.


wmann:
This pull request will be closed as it is handled in #179

PR #179, opened 2018-02-20
changes in cisco_nxos_show_mac_address-table and cisco_ios_show_mac-address-table

vaneuk:
##### ISSUE TYPE

 - Bugfix Pull Request
 - Additional Testing

##### COMPONENT

cisco_nxos_show_mac_address-table.template
cisco_ios_show_mac-address-table.template

##### SUMMARY

**i**
Changed textfsm variables in order for them to be consistent between cisco platforms.

cisco_ios_show_mac-address-table:
DESTINATION_PORT -> PORT
DESTINATION_ADDRESS -> MAC

cisco_nxos_show_mac_address-table:
PORTS -> PORT

**ii**
Now templates return error in case of unexpected output.

**iii**
Adjusted cisco_nxos regex to handle:
*) N3K and N5K periodically (every ~500 lines) show header (test case cisco_nxos_show_mac_address-table2.raw)
*) N4K regex is rather different from other nexus switches, added support for it (cisco_nxos_show_mac_address-table3.raw)

**iv**
Adjusted cisco_ios regex to handle newly added test cases cisco_ios_show_mac-address-table5.raw and cisco_ios_show_mac-address-table6.raw 
(almost the same as #177 )

```

```


wmann:
Great work. This pull request already fixes a problem which I noticed today. I will close my pull request #177

wmann:
I've done some tests last week on NX-OS, IOS, IOS-XE last week and some changes were necessary. 

# For NX-OS I had to do two adjustments:
### 1. On Nexus 5500/5600 running 7.1(4)N1(1) the legend looks like that:
```
Legend:
        * - primary entry, G - Gateway MAC, (R) - Routed MAC, O - Overlay MAC
        age - seconds since last seen,+ - primary entry using vPC Peer-Link
```

So I had to adjust your template from:
```
  ^.*age - seconds since first seen
```
to:
```
  ^.*age - seconds since
```

### 2. On Nexus 7000 running 7.2(2)D1(2) the header is changed and looks like that:
```
   VLAN     MAC Address      Type      age     Secure NTFY   Ports/SWID.SSID.LID
---------+-----------------+--------+---------+------+----+------------------
```
So I had to add the following line to your template:
```
  ^\s+VLAN/BD\s+MAC\s+Address\s+Type\s+age\s+Secure\s+NTFY\s+Ports -> TYPE1
```
So Maybe there is a better way in RegEx to add this to the existing statement but this one works.

# For IOS I had the following errors:
So far with not a real solution. I had to comment out the Error lines for TYPE2 and TYPE3 so the unmatched lines will be ignored and everything else got parsed.

### 1. On Catalyst 6500 running 15.1(2)SY6
```
*  ---  0002.abcd.1234    static  No           -   Router
*  ---  0000.0000.0000    static  No           -   Router
```
lines like that trigger an Error

### 2. On Catalyst 4500 running 12.2(46)SG
After the normal MAC addresses there will be Multicast Entries shown which also trigger an Error
```
Multicast Entries
 vlan    mac address     type    ports
-------+---------------+-------+--------------------------------------------
   1    ffff.ffff.ffff   system
 123    ffff.ffff.ffff   system Switch,Po1
 456    ffff.ffff.ffff   system Gi1/1,Gi1/2,Gi1/3,Gi1/4,Gi1/5,Gi1/6
                                Gi1/7,Gi1/8,Gi1/9,Fa2/1,Fa2/2,Fa2/3,Fa2/4
```

So maybe you have an idea how to handle this in the IOS template in a better way and also add the changes to the NX-OS templates.

vaneuk:
@wmann thanks for valuable inputs! will update the PR

i-m-d:
The template cisco_ios_show_mac-address-table.template DO NOT catch vlan id of 4 digits in my Cisco Catalyst (several versions 12.2.x) as there is NO space before vlan id and the template looks for one o more spaces (^\s+):

<pre>
TYPE3
  ^\s+${VLAN}\s+${DESTINATION_ADDRESS}\s+${TYPE}\s+\S+\s+${DESTINATION_PORT} -> Record

TYPE4
  ^\s+${VLAN}\s+${DESTINATION_ADDRESS}\s+${TYPE}\s+${DESTINATION_PORT} -> Record
</pre>

one of my switches as an example:

<pre>
#show mac address-table
          Mac Address Table
-------------------------------------------

Vlan    Mac Address       Type        Ports
----    -----------       --------    -----
 768    c2ea.e4cc.????    DYNAMIC     Po1
1383    0050.56ae.????    DYNAMIC     Po1
</pre>

The vlan id 1383 starts in the first column, so i have change my local file from "^\s+${VLAN}" to "^\s*${VLAN}" and now vlan id of 4 digits are catched correctly.

jmcgill298:
@itdependsnetworks what do you think about the renaming of the two capture groups?

PR #184, opened 2018-03-12
added template for cisco xr : "admin show environment fan" ,"admin show vm" and "show version"

rishabh5j:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_xr_admin_show_environment_fan.template
cisco_xr_admin_show_vm.template

##### SUMMARY
Added template cisco_xr_admin_show_environment_fan.template
Added template cisco_xr_admin_show_vm.template , 
Added template cisco_xr_show_version.template ,
Modified the index 
Added tests for all three

```

### CODE REVIEW

**Comment by rishabh5j** [comment]
@jedelman8  Hi Jason, can this be approved and merged?

File: templates/cisco_xr_admin_show_environment_fan.template [hunk 1]
```diff
@@ -0,0 +1,39 @@
+Value MODULE (0\/(FT|PM)\d)
```
itdependsnetworks: what is wrong with `/S+` here?

templates/cisco_xr_admin_show_vm.template
```diff
@@ -0,0 +1,21 @@
+Value Filldown LOCATION (\d\/(FC|RP|SC)?\d)
```
itdependsnetworks: Same here, what is wrong with `/S+`?

templates/cisco_xr_admin_show_environment_fan.template
```diff
@@ -0,0 +1,39 @@
...
+  ^\s+${FAN_SPEED} -> Continue
+  ^\s+\d+\s+${FAN_SPEED} -> Continue
+  ^\s+\d+\s+\d+\s+${FAN_SPEED} -> Continue
+  ^\s+\d+\s+\d+\s+\d+\s+${FAN_SPEED} -> Continue
+  ^\s+\d+\s+\d+\s+\d+\s+\d+\s+${FAN_SPEED} -> Continue
+  ^\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+${FAN_SPEED} -> Record
+  ^.*$$ -> PM
+  ^\s+$$
+  ^$$
+  ^.* -> Error "LINE NOT FOUND"
```
itdependsnetworks: Thanks!

templates/cisco_xr_admin_show_environment_fan.template
rishabh5j: @itdependsnetworks  I though the regex should be specific if possible. Is that not correct?

itdependsnetworks: To an extent yes, to an extent no :) 

Looking at the data, it doesn't seem like that would cover all future use cases. You are fairly well protected here in that you properly transition state to `FM`, after which I think it's a safe assumption that this is the only thing that would fit, and not be a false positive.

itdependsnetworks:
If you make these few updates, can get this one in as well.

jmcgill298:
https://github.com/rishabh5j/ntc-templates/pull/2 submitted to use "\S+"

PR #213, opened 2018-05-31
Add textFSM module for show tacacs in Cisco IOS

FedericoOlivieri:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_ios_show_tacacs.template

##### SUMMARY
Added show tacacs textFSM module for cisco IOS

```TASK [debug] ******************************************************************************************************************************************************************************************************
ok: [IE2BBRTR02] => {
    "show_tacacs": [
        {
            "TACACS": "10.42.9.140"
        }, 
        {
            "TACACS": "10.40.9.140"
        }
    ]
}

```

### DIFF
--- a/templates/cisco_ios_show_tacacs.template
+++ b/templates/cisco_ios_show_tacacs.template
@@ -1,0 +1,4 @@
+Value TACACS (\d+\.\d+\.\d+\.\d+)
+
+Start
+  ^\s+Server address:\s+${TACACS} -> Record
--- a/templates/index
+++ b/templates/index
@@ -145,6 +145,7 @@ cisco_ios_show_version.template, .*, cisco_ios, sh[[ow]] ver[[sion]]
 cisco_ios_show_ip_arp.template, .*, cisco_ios, sh[[ow]] i[[p]] a[[rp]]
 cisco_ios_show_ip_bgp.template, .*, cisco_ios, sh[[ow]] i[[p]] bgp
 cisco_ios_show_clock.template, .*, cisco_ios, sh[[ow]] clo[[ck]]
+cisco_ios_show_tacacs.template, .*, cisco_ios, sh[[ow]] tacacs
 cisco_ios_show_vlan.template, .*, cisco_ios, sh[[ow]] vlan
 cisco_ios_show_vrf.template, .*, cisco_ios, sh[[ow]] vrf
 cisco_ios_dir.template,  .*, cisco_ios, dir
--- a/tests/cisco_ios/show_tacacs/cisco_ios_show_tacacs.raw
+++ b/tests/cisco_ios/show_tacacs/cisco_ios_show_tacacs.raw
@@ -1,0 +1,24 @@
+Tacacs+ Server -  public  :
+            Server address: 10.1.1.140
+               Server port: 49
+              Socket opens:     138084
+             Socket closes:     137992
+             Socket aborts:          0
+             Socket errors:          0
+           Socket Timeouts:         59
+   Failed Connect Attempts:         52
+        Total Packets Sent:     147753
+        Total Packets Recv:     147693
+
+
+Tacacs+ Server -  public  :
+            Server address: 10.2.1.140
+               Server port: 49
+              Socket opens:       2027
+             Socket closes:       2027
+             Socket aborts:          0
+             Socket errors:          0
+           Socket Timeouts:          0
+   Failed Connect Attempts:          2
+        Total Packets Sent:       2053
+        Total Packets Recv:       2053


templates/cisco_ios_show_tacacs.template
```diff
@@ -0,0 +1,4 @@
+Value TACACS (\d+\.\d+\.\d+\.\d+)
+
+Start
+  ^\s+Server address:\s+${TACACS} -> Record
```
jmcgill298: The last line in the captures should be `^.*` -> Error`, so that we can ensure we know the data. I know that most of them don't have this, but it is the standard for all new templates. This also means that all of the lines that are not capturing data still need to be matched. They don't have to be elaborate, but something like `^\s+Socket` would match several lines and prevent from hitting the `Error` line.

templates/index
```diff
@@ -145,6 +145,7 @@ cisco_ios_show_version.template, .*, cisco_ios, sh[[ow]] ver[[sion]]
 cisco_ios_show_ip_arp.template, .*, cisco_ios, sh[[ow]] i[[p]] a[[rp]]
 cisco_ios_show_ip_bgp.template, .*, cisco_ios, sh[[ow]] i[[p]] bgp
 cisco_ios_show_clock.template, .*, cisco_ios, sh[[ow]] clo[[ck]]
+cisco_ios_show_tacacs.template, .*, cisco_ios, sh[[ow]] tacacs
```
jmcgill298: This should be moved above the previous entry as the command is one character longer and is the least preferred alphabetically from commands of the same length.

jmcgill298: what are the minimal characters required after `sh[[ow]]` to get tacacs command to execute?

tests/cisco_ios/show_tacacs/cisco_ios_show_tacacs.raw
```diff
@@ -0,0 +1,24 @@
+Tacacs+ Server -  public  :
```
jmcgill298: Need a parsed file to go with the raw file. The parsed file should be the parsed results from this data.
```
parsed_sample:
  - tacacs: 10.1.1.140
  - tacacs: 10.2.1.140
```
That would be as you have the template now; but you will need to adjust based on above comments for the template parsing.

templates/cisco_ios_show_tacacs.template
```diff
@@ -0,0 +1,4 @@
+Value TACACS (\d+\.\d+\.\d+\.\d+)
+
```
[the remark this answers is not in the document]
FedericoOlivieri: Agreed on this

templates/cisco_ios_show_tacacs.template
FedericoOlivieri: Not quite sure to understand this. If I add the line as you said, the script iterate over just the first tacacs output and is not catching the port

```Value TACACS (\d+\.\d+\.\d+\.\d+)
Value PORT (\d+)

Start
  ^\s+Server address:\s+${TACACS}
  ^\s+Server port:\s+${PORT}^.* -> Error```

TASK [debug] ******************************************************************************************************************************************************************************************************
ok: [IE2BBRTR02] => {
    "show_tacacs": [
        {
            "PORT": "", 
            "TACACS": "10.40.9.140"
        }
    ]
}

```

templates/index
FedericoOlivieri: Agreed on this

FedericoOlivieri: is `sh tac` so I believe should be sh[[ow]] tac[[acs]]

tests/cisco_ios/show_tacacs/cisco_ios_show_tacacs.raw
FedericoOlivieri: Not quite sure about this step. What this parser is required for?

templates/cisco_ios_show_tacacs.template
jmcgill298: Look at this one: https://github.com/jmcgill298/ntc-templates/blob/9c6df74b6a2f30c3427aa6f43aa55ba6e57f45b8/templates/hp_procurve_show_vlans.template

The last line should literally be `^.* -> Error`
```
  ^\s+Server address:\s+${TACACS}
  ^\s+Server port:\s+${PORT} -> Record
  ^$match3 #this needs to be updated to a regex line, and you will probably need multiple of these
  ^\s*$$
  ^.* -> Error
```

tests/cisco_ios/show_tacacs/cisco_ios_show_tacacs.raw
jmcgill298: This is required for our tox/travis tests. The parsed file is compared against the raw file to ensure that all parsed data would be derived from the corresponding raw file. Look at https://github.com/networktocode/ntc-templates/tree/master/tests/cisco_ios/show_interfaces/ directory and you will notice that the raw and parsed files have corresponding names. The parsed data is what textfsm would return after parsing the corresponding raw data.

templates/cisco_ios_show_tacacs.template
FedericoOlivieri: what  '^$match3' should match?

jmcgill298: you need to match every line in the raw data, or else it will hit the `Error` and raise an exception. If you just add ^.* -> Error to your current template and run it, you will see what I mean.

A line like `^\s+Socket` would account for four of the lines in the raw sample that you are currently not matching; that leaves 3 additional lines that you would need to match to avoid hitting the Error

templates/index
jmcgill298: that would be correct then

jmcgill298:
#344

PR #214, opened 2018-06-07
Find MAC addresses in cisco IOS show version.

kimoldfield:
##### ISSUE TYPE
 - Update Template Pull Request

##### COMPONENT
cisco_ios/show_version

##### SUMMARY
Recognise base MAC address from show version output.


itdependsnetworks:
The implementation seems fine, but I would prefer to not have to change this if possible, especially since it seems to only be in stack switches. Can you get this information from a show switch or other show command?

kimoldfield:
I'll try a new template for "show switch detail" as a separate pull request.

kimoldfield:
Some switches (such as our Cat 3560s) have a MAC address in "show version", but don't have the "show switch detail" command - so discovering MAC addresses in "show version" is still be required.
Having the MAC address often filled in is better than always having to use a different command which may only work on some switches.

itdependsnetworks:
It’s better from this one edge use case for sure, but better overall is a relative. Can you verify it’s not there with a show tech and some includes?

kimoldfield:
Given the large amount of information many devices have to generate with a "show tech", I feel "show tech" command is inappropriate (even with |include) for finding a MAC address which can be found with a simple "show version".
I'm not interested in further following up different commands which may give me what I can easily get with "show version". I'll continue to use my show_version.template.
If you choose not to incorporate my pull request them I'm OK with that.
I hope the repository continues to remain useful to me for other commands.

itdependsnetworks:
I was merely asking to find the commands via show tech, was not suggesting to use that as a solution.

kimoldfield:
On one of our C3650 switches the "show tech" output has the base MAC address in the following commands:
show version
show spanning-tree
show vtp status
The output from the latter two commands probably depends on the if those protocols are running, and the version of that protocol being run.

PR #234, opened 2018-07-26
Refine index order tests and added helper/sort-template-index

bdowling:
##### ISSUE TYPE

 - Additional Testing
 - Docs Pull Request

##### COMPONENT

index sorting

##### SUMMARY
I pulled forward some work I did some time ago on validating the index order.  Previously this had used the template filename, I changed it so it uses the command regex without the square brackets for validating the length.  This also pointed out that "show_ip_int_brief" for IOS was using a shortened filename, which was also renamed here.

Some of this is also just printing slightly nicer messages in the test and/or mentioning the OS so you can find the errant line easier.


jmcgill298:
@bdowling I incorporated some of the name completion changes and added a couple of more. I see that the index file included in this PR is not sorted properly, so I think the logic still needs a little work.

PR #241, opened 2018-08-15
Brocade fastiron show version

jmcgill298:
ISSUE TYPE
New Template Pull Request
COMPONENT
brocade, fastiron, show_version

SUMMARY
added new template
brocade_fastiron_show_version.template


jmcgill298:
@hilash I have updated to account for boot drive being `Primary` or `Seconfary`, and then I added a second commit to capture more of the data. What do you think of my suggestions in the second commit?

PR #288, opened 2018-11-13
Packaging

micahculpepper:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
Python packaging

##### SUMMARY
Made Python packaging simpler and more deterministic.

This PR:

1. Commits a symlink to git, so `ntc_templates/templates` will point to `templates`. This puts the templates inside the path of the Python project, which makes packaging tools happy. It also allows the `templates` directory to stay where it is, so existing PRs and tooling can still reference templates in the same location.
2. Removes unnecessarily complex code for finding the templates folder and version number


micahculpepper:
Hi @jmcgill298,

I use an import like in https://github.com/networktocode/ntc-templates/commit/71038c72f5e2677f106d80e4aa01b1e6022fe4f0 to handle gathering the version number in other projects I work on. Unfortunately they're not public for me to be able to link to them. It works, as long as the current working directory when `python setup.py sdist` is run contains the `ntc_templates` directory, or the current version of the project is installed in develop mode. [Reference](https://docs.python.org/3/tutorial/modules.html#the-module-search-path).

However, I can understand not wanting to change something that works. I'll go ahead and revert that change so as not to distract from the rest of the PR.

I'm totally in favor of the switch to `textfsm` and ready to rebase as soon as v1 of their package is available on PyPI.

micahculpepper:
I see `setup.py` on `master` has already been updated to require `textfsm`, so I went ahead and rebased.

It looks like they still only have version 0.4.1 on [PyPI](https://pypi.org/project/textfsm/#history), and I think this project will need at least version 1 for import compatibility.

jmcgill298:
This is a good page highlighting the options for versioning: `https://packaging.python.org/guides/single-sourcing-package-version/`. Option 6 is similar to what you proposed; if you look at the warning for that option, you will see the concern I had for importing from another file.

Thanks for pointing out the import issues with `parse.py`; I have updated it to prefer TextFSM as a package and then try how importing the `clitable` module.

micahculpepper:
@jmcgill298 is there anything outstanding on this PR?

PR #296, opened 2018-11-14
add juniper_junos_show_chassis_cluster_interfaces

jkraszewski:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
juniper_junos_show_chassis_cluster_interfaces.template, junos, show chassis cluster interfaces

##### SUMMARY
New juniper template


jkraszewski:
@jmcgill298 to my knowledge I only needed 2 pieces of information from this commands result so I was only grabbing those values. I could remove all of the comments and leave the lines that grab those values but the error line would break it. How should I proceed?

jmcgill298:
We have to match all of the lines regardless of whether we capture all of the possible data. This is the only way we can ensure we know all of the data was captured correctly.

PR #302, opened 2018-11-19
Modified cisco_asa_show_object-group_network.template

Jaydubya35:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_asa_show_object-group_network.template, cisco_asa, show object-group network
cisco_asa_show_object-group_network.template, cisco_asa, show running-config object-group network

##### SUMMARY
On the ASA, if you run `show object-group network` or `show running-config object-group network` you get the same output __except__ for the colon after the word `description`. I have modified the existing cisco_asa_show_object-group_network.template to accept the colon as an optional character for it to work for both commands.

I plan to add the `show running-config object network` template soon, and didn't want the `show running-config object-group network` command to conflict.


Jaydubya35:
The problem with your suggestion is the difference between the minimum letters of the syntax `object-group` to run to differentiate object-groups from single objects when running it with `running-config`.

"show object-group network" = `sh ob n`
"show running-config object-group network" = `sh ru object- n`
"show running-config **object** network" = `sh ru o n`

As you can see object and object-group would conflict with your proposed single regex. If you would still like a combined regex for the single template, the logical OR would need to contain the full `object-` when combined with `running-config`. Maybe something like this: 
```
sh[[ow]] (ru[[nning-config]] object-[[group]]|ob[[ject-group]]) n[[etwork]]
```

Let me know if you would prefer it that way or the split way that I originally committed.

jmcgill298:
How were you testing that my proposal doesn't work? I tested with the ntc_show_command Ansible module, and it was able to read the index file and locate the template for both commands.

The branch as you currently have it does not have a template for `show running-config object-group network`, so it still won't be able to parse based on the command syntax.

Jaydubya35:
Sorry, I should have specified it has the potential to conflict with #306. In that PR I added `show run object network` to the index. As long as your new regex is **below** that one it works fine.

Working example with your solution:
```
cisco_asa_show_running-config_object_network.template, .*, cisco_asa, sh[[ow]] ru[[nning-config]] o[[bject]] n[[etwork]]
cisco_asa_show_object-group_network.template, .*, cisco_asa, sh[[ow]] (?:ru[[nning-config]] |)ob[[ject-group]] n[[etwork]]
```
But, if the order was reversed (since longer commands should come first in the index file according to your own guidelines) it would no longer work as `show run ob network` would trigger on the index line that you provided. My solution would work either above or below `show run object network`.

I have provided a new commit with this change. Please review and let me know your thoughts.

jmcgill298:
The order needs to be maintained, but this should work: `sh[[ow]] (?:ru[[nning-config]] object-[[group]]|ob[[ject-group]]) n[[etwork]]`

Jaydubya35:
Are you sure? You have `?:` still in the parentheses so the entire section is optional and `show network` will also match.

Jaydubya35:
Sorry, I mixed up non-capturing and optional. Yes, it will work with `?:`

PR #317, opened 2018-12-18
removed error when an empty line exists in command output

routetonull:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_wlc_ssh_show_ap_config_general.template

##### SUMMARY

the template generates an error when an empty line exists in the command output but blank lines are expected


jmcgill298:
@routetonull I looked at this today, and we account for blank lines in the template; can you provide the error message you receive?

FragmentedPacket:
@routetonull Removing the `^. -> Error` doesn't adhere to our standard as we're wanting to make sure every line is accounted for, but doesn't have to be captured. The error you're encountering means the template isn't taking into account some data from the raw output that needs to be accounted for or might be a bug.

I'm going to close this PR, but please open an issue with all the necessary details and we'll be more than happy to look into a resolution with you.

PR #229, opened 2018-07-17
Cisco ios show ip interface.template

FragmentedPacket:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_ios_show_ip_interface.template

##### SUMMARY

This is adding a template for show ip interface that structures the following data:
Interface
Link status
protocol status
ip address
mask
mtu
Inbound ACL
Outgoing ACL
<!---

PR #322, opened 2018-12-20
Adding cisco_ios_show_ip_interface

vladola:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT
 cisco_ios_show_ip_interface.template, cisco_ios, `show ip interface`

##### SUMMARY
I needed a way to get Interface name, ip address *and* VRF (if empty, it's in GRT). Clearly the command has a lot more info than this, and can be expanded in the future if needed.


jmcgill298:
@vladola I submitted some updates in vladola#1. The important update is that new templates should raise an `Error` when an unknown line is encountered. Please let me know what you think of my updates. If you decide to merge that PR, then I will go ahead merge this one.

FragmentedPacket:
@vladola I have #229 opened that I think covers some of this. I can add VRF support into it. Is that cool if I add the VRF support into that since more fields are supported at this point?

vladola:
@jmcgill298 Looks good. Sorry, I knew about the new requirements, I just had no time in the moment to spend on doing it, so I just posted anyway in case someone needed something similar. Also, unsure why I was not notified of your comment, sorry for the long reply.

PR #327, opened 2019-01-10
add cisco_xr_show_bgp_neighbors (IPv4 Only)

jtishey:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_xr_show_bgp_neighbors.template, Cisco IOS-XR, show bgp neighbors

##### SUMMARY

Adding template, tests, and updated index for "show bgp neighbors" on Cisco IOS-XR.  Currently only works for IPV4 Unicast neighbors


jmcgill298:
@jtishey Thanks for the PR.

We do have a standard for all new templates that requires each state to have a catchall match that raises an error; this way we can have reasonable confidence that we are aware of each line in the command output.
```
Start
  ...
  ^\s*$$
  ^. -> Error
```

This also means that you will need to provide regex lines to match every line in your raw file.

jtishey:
yikes, ok you can close this PR then if you want  might be a while before i can go through all the possible lines of output.

jmcgill298:
I have moved this into branch, `xr_show_bgp_neigh`

PR #341, opened 2019-01-21
CISCO XR SHOW BGP NEIGH: Add new template

jmcgill298:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_xr_show_bgp_neighbor

##### SUMMARY

Add new template

```

```

PR #346, opened 2019-01-22
Update XR show bgp neighbors template

jtishey:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_xr_show_bgp_neighbor

##### SUMMARY
Updates to PR #341 for Cisco XR show bgp neighbors and add an additional test

Added tests/cisco_xr/show_bgp_neighbors1.raw
Added tests/cisco_xr/show_bgp_neighbors1.parsed
Updated templates/cisco_xr_show_bgp_neighbors.templates

Value GR (.+?)
Value List DEFAULT_ORIGINATE (.+)
Value LAST_RESET_REASON (.*)

START
  ^\s+Remote\s+AS\s+${REMOTE_AS}\,\s+local\s+AS\s+${LOCAL_AS}\,\s+(?:no\-prepend,\s+)?(?:replace\-as\,\s+)?${TYPE}\s+link
  ^\s+Administratively\s+shut\s+down
  ^\s+BFD\s+[Ee]nabled
  ^\s+Configured hold time\:\s+\d+\,\s+keepalive\:
  ^\s+Graceful\s+restart\s+is\s+${GR}\s*$$
  ^\s+Restart\s+time\s+is
  ^\s+Stale\s+path\s+timeout\s+time\s+is
  ^\s+Graceful\s+Restart\s+\(GR\s+Awareness\)\:
  ^\s+Inbound\s+message\s+logging
  ^\s+Outbound\s+message\s+logging

AFI
  ^\s+Default\s+information\s+originate\:\s+${DEFAULT_ORIGINATE}
  ^\s+AF-dependent capabilities
  ^\s+Graceful\s+Restart\s+capability
  ^\s+Local\s+restart\s+time\s+is
  ^\s+Maximum\s+stalepath\s+time\s+is
  ^\s+Remote\s+Restart\s+time\s+is
  ^\s+Private\s+AS\s+number\s+removed
  ^\s+Community\s+attribute
  ^\s+Advertise\s+VPNv[46]\s+routes\s+(is\s)?enabled\s+with
  ^\s+AS\s+[oO]verride\s+is
  ^\s+Send\s+Multicast\s+Attributes

CONNECTION
  ^\s+Last\s+reset\s+${LAST_RESET}\,\s+due\s+to\s+${LAST_RESET_REASON}
  ^\s+Last\s+reset\s+${LAST_RESET}
  ^\s+External\s+BGP\s+neighbor\s+not\s+directly


templates/cisco_xr_show_bgp_neighbors.template
```diff
@@ -23,6 +24,7 @@ Value List PREFIXES_OUT_SUPPRESSED (\d+)
 Value List PREFIXES_OUT_WITHDRAWN (\d+)
 Value List MAX_PREFIXES (\d+)
 Value List MAX_PREFIXES_WARN_PERCENT (\d+)
+Value List DEFAULT_ORIGINATE (.+)
```
jmcgill298:
```suggestion
Value List DEFAULT_ORIGINATE (.+?)
```

templates/cisco_xr_show_bgp_neighbors.template
```diff
@@ -32,7 +34,7 @@ Value LOCAL_PORT (\d+)
 Value REMOTE_ADDRESS (.*)
 Value REMOTE_PORT (\d+)
 Value LAST_RESET (\S+)
-Value LAST_RESET_REASON (.+?)
+Value LAST_RESET_REASON (.*)
```
jmcgill298: Was this not working for your data? Follows for line 106

templates/cisco_xr_show_bgp_neighbors.template
```diff
@@ -78,6 +88,12 @@ AFI
   ^\s+BGP\s+neighbor\s+version\s+${VERSION}\s*$$
   ^\s+Update\s+group:
   ^\s+Route-Reflector\s+${ROUTE_REFLECTOR_ROLE}\s*$$
+  ^\s+Default\s+information\s+originate\:\s+${DEFAULT_ORIGINATE}
```
jmcgill298:
```suggestion
  ^\s+Default\s+information\s+originate\:\s+${DEFAULT_ORIGINATE}\s*$$
```

templates/cisco_xr_show_bgp_neighbors.template
jtishey: so with ```  Default information originate: default sent``` we want to match to the end of the line including whitespace or else it will only capture "default" instead of "default send", no?

jmcgill298: It will capture everything up to the end of the line excluding a space. If the line doesn't have a space, it will still capture all of the text (that is why we use `*` instead of `+`)

templates/cisco_xr_show_bgp_neighbors.template
jmcgill298: Without ending the line with `\s*$$` that will happen, but my update included `\s*$$`, so it will capture everything up to the end of the line.

jtishey: @jmcgill298   I see what the issue was.  Some routers output 
```  ^\s+Last\s+reset\s+${LAST_RESET}\,\s+due\s+to\s+${LAST_RESET_REASON}```
and some just output
```  ^\s+Last\s+reset\s+${LAST_RESET}```

So there is a need to match either or (or just omit the values) .  so the .+? failed with the first output scenario...  I'll look into a more proper solution.

jtishey: @jmcgill298   Can you check out the latest commit 60e6e56 and see if that makes sense?

PR #350, opened 2019-01-26
XR SHOW HSRP: Add new template

jmcgill298:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_xr_show_hsrp
##### SUMMARY

Add new template

```

```


jmcgill298:
@rbeumer can you verify this works in your environment? I add a test file I found on the internet.

Updates I made to your template:
* Added capture group fro `preempt` and change regex line to accept preempt being enabled/disabled
* Changed the `addr` captures to capture continues non-whitespace characters:
   - Allows for `local` to be in either "active" or "standby" positions
   - Allows for IPv6 capturing
* Capture all known lines of data, and record an error if a line is not matched

PR #358, opened 2019-01-27
XR SHOW CONTROLLERS FABRIC FIA ERRORS INGRESS: Add new template

jmcgill298:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_xr_show_controllers_fabric_fia_errors_ingress_location
##### SUMMARY

Add new template

```

```


jmcgill298:
@vaneuk My last commit to this branch makes some big changes. I think it makes sense to capture the data as lists in order to account for different architectures. Please let me know what you think of my data representation.

vaneuk:
@jmcgill298 please give me a moment to think about this.

jmcgill298:
@vaneuk have you had a chance to look at this more?

vaneuk:
@jmcgill298 i checked the changes. i do not like Lists in this case for the following reasons:

- Now some counters are integers and some a lists. It makes parsed data harder to use in monitoring systems. For example when all counters are integers you can easily sum all of the counters and create an alert based on the sum value. And when some counters are actually lists you will have to convert lists to integer first.
- Adding Lists does not make regex template really much more flexible. Output of command is platform dependent. The tests that i provided in my commit were made for Typhoon cards. And output for Trident cards is completely different.

please let me know if the points above make sense

jmcgill298:
@vaneuk I have reverted that commit

PR #362, opened 2019-01-30
added aruba_os_show_aaa_authentication_all

jsnyder81:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

aruba_os_show_aaa_authentication_all.template
aruba_os
show aaa authentication all

##### SUMMARY

Addition of template file
Addition of raw file
addition of parsed yaml file
update of index

```

```
If you guys need anything from me, just let me know.  Thanks for letting me contribute.

### DIFF
--- a/templates/aruba_os_show_aaa_authentication_all.template
+++ b/templates/aruba_os_show_aaa_authentication_all.template
@@ -1,0 +1,12 @@
+Value METHOD (Web|802.1x|Management|tacacs-accounting)
+Value SUCCESSES (\d+)
+Value FAILURES (\d+)
+
+
+Start
+  ^Auth Method Statistics -> AUTHMETHODSTATS
+
+AUTHMETHODSTATS
+  ^${METHOD}\s+${SUCCESSES}\s+${FAILURES} -> Record
+
+EOF
--- a/templates/index
+++ b/templates/index
@@ -53,6 +53,7 @@ arista_eos_show_vlan.template, .*, arista_eos, sh[[ow]] vl[[an]]
 
 aruba_os_show_ipv6_interface_brief.template, .*, aruba_os, sh[[ow]] ipv6 in[[terface]] b[[rief]]
 aruba_os_show_ip_interface_brief.template, .*, aruba_os, sh[[ow]] ip in[[terface]] b[[rief]]
+aruba_os_show_aaa_authentication_all, .*, aruba_os, sh[[ow]] aaa authen[[tication]] a[[ll]]
 
 avaya_ers_show_mac-address-table.template, .*, avaya_ers, sh[[ow]] mac-a[[ddress-table]]
 avaya_ers_show_mlt_all-members.template, .*, avaya_ers, sh[[ow]] ml[[t]] a[[ll-members]]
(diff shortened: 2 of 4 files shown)


templates/aruba_os_show_aaa_authentication_all.template
```diff
@@ -0,0 +1,12 @@
...
+Value FAILURES (\d+)
+
+
+Start
+  ^Auth Method Statistics -> AUTHMETHODSTATS
+
+AUTHMETHODSTATS
+  ^${METHOD}\s+${SUCCESSES}\s+${FAILURES} -> Record
+
+EOF
```
jmcgill298: you aren't using `Filldown`, so I don't believe the `EOF` is needed

templates/aruba_os_show_aaa_authentication_all.template
```diff
@@ -0,0 +1,12 @@
+Value METHOD (Web|802.1x|Management|tacacs-accounting)
```
jmcgill298: I think we should be safe to use `(\S+)` here, and that would allow for additional options or variations in formats (capitalization, hyphens, abbreviations).

templates/aruba_os_show_aaa_authentication_all.template
```diff
@@ -0,0 +1,12 @@
...
+Value SUCCESSES (\d+)
+Value FAILURES (\d+)
+
+
+Start
+  ^Auth Method Statistics -> AUTHMETHODSTATS
```
jmcgill298: I don't think the state change does a lot for us, but if you want to keep it, then that is fine. I would ask that it follows TextFSMs state conventions of CamelCase

templates/index
```diff
@@ -53,6 +53,7 @@ arista_eos_show_vlan.template, .*, arista_eos, sh[[ow]] vl[[an]]
 
 aruba_os_show_ipv6_interface_brief.template, .*, aruba_os, sh[[ow]] ipv6 in[[terface]] b[[rief]]
 aruba_os_show_ip_interface_brief.template, .*, aruba_os, sh[[ow]] ip in[[terface]] b[[rief]]
+aruba_os_show_aaa_authentication_all, .*, aruba_os, sh[[ow]] aaa authen[[tication]] a[[ll]]
```
jmcgill298:
```suggestion
aruba_os_show_aaa_authentication_all.template, .*, aruba_os, sh[[ow]] aaa authen[[tication]] a[[ll]]
```

templates/aruba_os_show_aaa_authentication_all.template
jsnyder81: Sure, in the limited context that you guys use the template.  However, I also use these to parse pieces out of “show tech-support” that also include these commands, need the state change to keep from matching very similar tables in output.  Same goes for /S.

jmcgill298: I don't see how the templates would be reliable if you are feeding the entire show-tech output. I think you would have to split the file into separate command entries or else you  will eventually end up including data from other commands, or miss subtle differences in command output between different OS versions and platforms.

FragmentedPacket:
@jsnyder81 Thanks for taking the time to submit this template. Do you mind addressing @jmcgill298 comments and the failed check? After those are addressed and all checks are passing, we can get this merged into master!

jsnyder81:
After multiple discussions on slack, I have decided that my requirements for these templates do not along with the requirements for NTC and will not be pursuing this.

FragmentedPacket:
@jsnyder81 Alright. Sorry to hear that. If anything ever changes we'd hope to see your contributions!

PR #363, opened 2019-01-30
 Add Cisco IOS show ipv6 neighbors 

kimoldfield:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Cisco IOS show ipv6 neighbors


jmcgill298:
https://github.com/kimoldfield/ntc-templates/pull/2

jmcgill298:
The simple regex is easier to read and is safe with the data provided. Tabular data can usually use generic capture groups as long as the table header is matched exactly.

You are right about the `+`, I did not pay close enough attention to that.

PR #364, opened 2019-01-30
Cisco IOS show switch detail

kimoldfield:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Cisco IOS show switch detail


jmcgill298:
Created https://github.com/kimoldfield/ntc-templates/pull/3 with suggested modifications. That PR is up for discussion as some changing the format of the capture groups to lists.

PR #365, opened 2019-01-31
Added missing PyYAML dependency in setup.py

Alan-R:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
setup.py

##### SUMMARY
Added missing dependency in setup.py: PyYAML


Alan-R:
Should not be here.

PR #370, opened 2019-02-04
Update cisco_nxos_show_interface_status.template

Pluppo:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_nxos_show_interface_status.template

##### SUMMARY
In NXOS 7.0.3.I7.5a there is a separate section with column names for the mgmt0 interface. This caused a "textfsm.TextFSMError" when the column name section appeared for the second time. This update fixes the issue.

```

```

### DIFF
--- a/templates/cisco_nxos_show_interface_status.template
+++ b/templates/cisco_nxos_show_interface_status.template
@@ -14,6 +14,7 @@ Start
 
 INTFS
   ^${PORT}\s+${NAME}\s+${STATUS}\s+${VLAN}\s+${DUPLEX}\s+${SPEED}\s+${TYPE}\s*$$ -> Record
+  ^[Pp]ort\s+[Nn]ame\s+[Ss]tatus\s+[Vv]lan\s+[Dd]uplex\s+[Ss]peed\s+[Tt]ype\s*$$
   ^-+\s*$$
   ^$$
   ^.*$$ -> Error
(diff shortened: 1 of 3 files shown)


Pluppo:
Hi, I just added two new test files:
cisco_nxos_show_interface_status2.raw
cisco_nxos_show_interface_status2.parsed

Let me know if you need me to do anything else.

jmcgill298:
@Pluppo looking at this now with the raw data, I think it would be better just to remove the state change, and put all of the matching under `Start`. Does that work when you do it?

Pluppo:
@jmcgill298 I'm sorry, but I'm not sure what you mean. My apologies, I'm quite new in dealing with Github and code collaboration.

I tried moving the `^[Pp]ort\s+[Nn]ame\s+[Ss]tatus\s+[Vv]lan\s+[Dd]uplex\s+[Ss]peed\s+[Tt]ype\s*$$` line from the "INTFS" section to the "Start" section, but that did not work.

The reason for this change is that in newer versions of nxos, there is a separate title section for the "mgmt0" interface, and a second title section for the remaining interfaces. Without telling the template to ignore the second occurrence of the title section, the following error appears:
`TextFSMError: State Error raised. Rule Line: 20. Input Line: Port          Name               Status    Vlan      Duplex  Speed   Type`

If you don't mind giving me an example of the code you think would be better to use, I'd by happy to test it.

jmcgill298:
@Pluppo I created https://github.com/Pluppo/ntc-templates/pull/1 which removes the state change. In TextFSM, `-> $SectionName`, is a state change. This can be used to segment up long output, or to distinguish between different expected output depending on Platform, OS, etc. The state change wasn't really doing anything in this template, and your updates made the `Start` state and the `INTFS` state exactly the same, so it just makes sense to only use `Start`.

PR #394, opened 2019-04-16
Feature/cisco ios show hosts

MatthiasGabriel:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Template: cisco_ios_show_hosts_summary
OS: Cisco IOS
Command: show hosts summary

##### SUMMARY
Add a new template for the cisco ios show hosts summary which parses the default_domain, name_servers, local cache entries and dynamic cache entries.
the name servers are a comma separated string instead a list, because i couldn't get it to work.


templates/cisco_ios_show_hosts_summary.template
```diff
@@ -0,0 +1,12 @@
+Value DEFAULT_DOMAIN (\S+)
+Value NAME_SERVERS ([\S ]+)
```
FragmentedPacket: Can you change the `NAME_SERVERS` regex to **.+**?

Also, can you please change any static spacing within the **Start** state to either **\s*** or **\s+**?

After those are addressed, we can merge this.

MatthiasGabriel: Thank you for pointing out the NAME_SERVERS.
I also updated the regex in the start state...

PR #406, opened 2019-04-24
YAMLLINT: Add yamllint to tox testing

jmcgill298:
* Rename all parsed files to use ``.yml`` extension
* Add ``.yamllint`` file
* Add test in ``tox.ini`` file to run yamllint
* Add cli tool to format YAML files using ``ruamel``
* Reformat parsed files to adhere to preferred YAML style


FragmentedPacket:
@jmcgill298 When do you want to implement these changes and #407 ?

jmcgill298:
I want to get near all of the open PRs merged in order to avoid conflicts and changing the file extensions

jmcgill298:
@FragmentedPacket this should be ready to go; do you mind reviewing; especially the README

README.md
```diff
@@ -235,41 +235,82 @@ $
...
+  * `-yd`: Takes a glob path to a directory or directories that will ensure all files ending in `.yml` adhere to the .yamllint settings
+  * `-c`: Takes the path to a `.raw` file, and generates the parsed data and saves the results adjacent to the `.raw` file following the defined standards in .yamllint.
+  * `-cd`: Takes a glob path to a directory or directories containing `.raw` files, and creates the appropriate parsed files in the appropriate directory.
+
+  The `-y` and `-yd` arguments are designed to allow developers to generate the expected parsed file how they want, and ensure that the formatting adheres to the defined standard for this project.
+
+  The `-c` and `-cd` arguments use `lib.ntc_templates.parse.parse_output()` to generate the parsed data; this means that you can use these arguments to auto-generate the test `.yml` file(s) for new templates; just be sure that the template's parsing behavior meets expectations. In order for the data to be parsed, the template must be places in `templates/` and the `templates/index` file must be updated to correctly point to the template file(s).
```
FragmentedPacket: **placed** instead of **places**

README.md
```diff
@@ -235,41 +235,82 @@ $
 
 Multiple `raw` and `parsed` files are supported per directory, and are encouraged, as there are differences depending on version, length, etc. Additional test files and more real life data helps ensure backwards compatibility is maintained as each template is updated and merged into the repo.
 
+All YAML files must adhere to the YAML standards defined in the `.yamllint` file in the root directory. Yamllint provides thorough documentation of their configuration settings [here](https://yamllint.readthedocs.io/en/stable/rules.html). 
+
+##### DEVELOPMENT HELPER SCRIPTS
+
+We have provided a cli utility to assist with properly building the parsed files. All arguments that can be passed to the script are mutually exclusive (i.e. you can only pass one argument). The file can be made executable with the `chmod +x development_scripts.py` command. The arguments are:
+
+  * `-y`: Takes the path to a YAML file and ensures that the file adheres to the .yamllint settings
```
FragmentedPacket: I'm not sure `-y` works exactly as intended. I didn't leave a space between a `#` and the words, but it kept them without a space which appears to be one of the restraints within the `.yamllint` file. If I'm misunderstanding, please let me know.

e.g. `#test` instead of the proper formatting `# test`

Maybe update the verbiage as well. I thought it would spit out something saying that the file wasn't in adherence to the `yamllint` file. Along that note, some kind of output would be great as well. "Formatting completed/No formatting required", etc.

development_scripts.py
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env python
```
FragmentedPacket: I think we should add tests for this as well since it is a utility we're providing and want to make sure everything works in it as intended.

README.md
jmcgill298: I think we would probably need to see if ruamel supports comment spacing, or see about adding a custom representer to handle that use case.

I don't mind adding a print statement to say it is complete, but would say that should be assumed if the process exited without errors. The actual scripts are simple, and just rewrite the existing YAML file regardless of the correctness of the file.

development_scripts.py
jmcgill298: I did refactor the functions last night with this in mind. Also need to add tests for `lib/ntc-templates/parse.py`

PR #408, opened 2019-04-26
Updated cisco_xr_show_version.template

reyrey16:
##### ISSUE TYPE

 - New Template Pull Request
    - Updated the cisco_xr_show_version.template 
 - Additional Testing
    - Added a CRS version RAW and parsed files

##### COMPONENT

##### SUMMARY
I updated the cisco_xr_show_version.template to incorporate the CRS output of the show version command, in addition to the NCS version that it already parses. See RAW and parsed files for the before and after outputs.

### DIFF
--- a/templates/cisco_xr_show_version.template
+++ b/templates/cisco_xr_show_version.template
@@ -1,28 +1,19 @@
-Value VERSION (\S+)
+Value VERSION (\S[^\[]+)
 Value UPTIME (.+)
 Value LOCATION (\S+)
-Value HARDWARE (\S+)
+Value HARDWARE (.+)
 Value BUILD_HOST (\S+)
 
 Start
-  ^.+UTC
-  ^Cisco.+Software.+Version.+
-  ^Copyright\s\(c\).+
+  ^Cisco IOS XR Software, Version ${VERSION} 
   ^Build\s+Information: -> Build_Info
-  ^[Cc]isco\s${HARDWARE}\s.+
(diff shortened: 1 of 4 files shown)


templates/cisco_xr_show_version.template
```diff
@@ -1,28 +1,19 @@
...
-  ^.+UTC
-  ^Cisco.+Software.+Version.+
-  ^Copyright\s\(c\).+
+  ^Cisco IOS XR Software, Version ${VERSION} 
   ^Build\s+Information: -> Build_Info
-  ^[Cc]isco\s${HARDWARE}\s.+
-  ^System\suptime\sis\s${UPTIME} -> Record
-  ^\s+$$
-  ^$$
-  ^.* -> Error "Line not found"
```
FragmentedPacket: Can you add back the error handling on here? It's a requirement to make sure we're accounting for all data.

When adding back the error handling, you'll probably need to add the other lines that were removed, and possible a few others. It doesn't have to be a match on the whole line, but the beginning of the lines would be great. I'd be more than happy to assist with any issue you have.

FragmentedPacket: Can you add back the error handling on here? It's a requirement to make sure we're accounting for all data.

When adding back the error handling, you'll probably need to add the other lines that were removed, and possible a few others. It doesn't have to be a match on the whole line, but the beginning of the lines would be great. I'd be more than happy to assist with any issue you have.

PR #418, opened 2019-05-30
Fixes #417 - Modified index sequence test to check the command not…

### LINKED ISSUE #417: Index sequence tests do not test command ordering correctly
##### ISSUE TYPE

 - Bug Report

##### SUMMARY

The test_index_order.py test does not test the commands correctly.  
The template index rules state:
```
OS in alphabetical order
Command in length other
When Length is the same, use alphabetical order
Keep space between OS's
```
However, the test splits the template file name into vendor_os and command, but the file name is not always the command.  The command is the last field on the line.  there could potentially be two commands pointing to the same template file.  For example,  I have the following in my templates:
```
arista_eos_show_ip_bgp.template, .*, arista_eos, sh[[ow]] i[[p]] bg[[p]] ne[[ighbors]] (\S+) ad[[vertised-routes]]
arista_eos_show_ip_bgp.template, .*, arista_eos, sh[[ow]] i[[p]] bg[[p]] ne[[ighbors]] (\S+) rec[[eived-routes]]
```
It is important that the sequencing be correct, as the command is significantly longer than the template, and you need to sequence according to the actual command, but the output is equivalent to the basic `show ip bgp` command, so the same template should be used to avoid duplication.
To correct this, lines 117 and 118 should be changed to:
```
        os = row[2].strip()
        cmd = row[3].strip()
```
Line 116 should be deleted as it is now redundant.  
This IMHO is a more correct test, as it more accurately tests how the command regex is processed by TextFSM.  It is also simpler.
You might find that correctly testing the command results in a need to re-sequence the file.
Arguably, it might be desirable to strip the square brackets out and compare the command without them.
##### STEPS TO REPRODUCE
Run the tests

##### EXPECTED RESULTS
Compare the commands, not the template file name elements

##### ACTUAL RESULTS
template file name is compared.

robobeaver6:
…the file name and corrected index file item sequence to follow the rules.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

Index file tests
##### SUMMARY

Fixes #417 - The rules for the index file sequencing state the following:

1. OS in alphabetical order (The OS must be one of the netmko os types)
2. Command in length order
3. When Length is the same, use alphabetical order
4. Keep space between OS's

Currently the test gets the OS and Command from the first field of the line containing the Template filename.    The current test spits this field on the `_` allowing the first two elements as the OS and the rest as the command.  In my opinion this is wrong for several reasons.

1. TextFSM parses the file on a first match basis, using the Regex in the fourth field.  This can be different from the template file name, as two or more commands could conceivably point to the same template.  By not checking the actual command regex, you could mask a sequencing error, where one command masks another.
2. The OS Name and command have fields specifically dedicated to them, why not use them?
3. Splitting the template filename on `_` is inconsistent.  the `cisco_wlan_ssh` OS type is not correctly parsed, as it has three elements.  It incorrectly passes the test, due to it being concatenated in the list of accepted fields in the test, which is also incorrect.

For example I have three commands which use the same filter in my index:
```
arista_eos_show_ip_bgp.template, .*, arista_eos, sh[[ow]] i[[p]] bg[[p]] ne[[ighbors]] (\S+) ad[[vertised-routes]]
arista_eos_show_ip_bgp.template, .*, arista_eos, sh[[ow]] i[[p]] bg[[p]] ne[[ighbors]] (\S+) rec[[eived-routes]]
arista_eos_show_ip_bgp.template, .*, arista_eos, sh[[ow]] i[[p]] bg[[p]]
```
all three commands return the same format response, so they all use the same template.  without checking the actual command field, the test would fail as the template filename does not carry the entire command.  The test can't handle two template file names being the same, it throws an error saying the alphabetical sequence is wrong, no matter which way round they are.  It also insists on the lines being further down the file than they should be.

I think it would probably be sensible to strip the [[]] out of the command strings before comparison (which I haven't currently done in the PR, but can add).  This would help prevent comparison errors in where people decide to put the expansion brackets.


jmcgill298:
Originally there was a 1-to-1 mapping of templates to commands, and the template names were the commands. As you have pointed out, this doesn't actually work out:
1. The commands might contain variable syntax (such as an IP address)
2. The same template can be used for multiple commands

The solution you have come up with assumes that the 4th field that we define will map to the command (assumming that we could strip `[[]]`). However, this field is regex, and so doesn't necessarily represent the actual commands. I know of two examples in our index file where regex syntax is used (we define each template once, and use regex in the command syntax to match multiple commands)
```
brocade_netiron_show_monitor_actual.template, .*, brocade_netiron, sh[[ow]] mon[[itor]] (?:ac|co)
cisco_asa_show_object-group_network.template, .*, cisco_asa, sh[[ow]] (?:ru[[nning-config]] object-[[group]]|ob[[ject-group]]) n[[etwork]]
```

There is also the issue of using multiple templates for the same command. For example, I would need at least two templates (depends on model/OS, but could require more) to parse all of Cisco IOS "show vlan" with Textfsm:
```
cisco_ios_show_vlan.template:cisco_ios_show_vlan_type.template, .*, cisco_ios, sh[[ow]] vl[[an]]
```

I'm not really sure what the "correct" answer is at this point. I do prefer defining the template once, and leveraging regex, as that is how CliTable was designed to be used (I would actually want to extend this into the OS field and have less templates defined).

As for `cisco_wlc_ssh`, I believe that should be changed to `cisco_wlc` as that follows the other names, and is used by Netmiko.

jvanderaa:
@jmcgill298 is there anything we can do on this one?

jmcgill298:
I don't think there is anything left to do on this one but close it out

PR #422, opened 2019-07-03
Corrected Local IP and Local AS number line

abhimukherjee91:
Corrected the local IP and local AS Number line to fetch correct value

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

Local IP and Local AS line was not fetching the correct IP address and AS Number
##### SUMMARY

Changed the first line of the rule with the correct regular expression

```

```


jmcgill298:
@abhimukherjee91 I believe the issue was the first line was finding matches in the tabular data. I made that match more restrictive in #517; can you see if that fixed the issue you were running into?

Issue #390, opened 2019-03-18
New template cisco_ios_show_snmp_user.template


jifox:
##### ISSUE TYPE

 - Feature Idea

##### TEMPLATE USING

```
Value Required USER_NAME (\w+)
Value ENGINE_ID (\w+)
Value STORAGE_TYPE (\w+)
Value STORAGE_TYPE_ACTIVE (\w+)
Value AUTHENTICATION_PROTOCOL (\w+)
Value PRIVACY_PROTOCOL (\w+)
Value GROUP_NAME (\w+)

Start
  ^User\s+name:\s+${USER_NAME}$$
  ^Engine\s+ID:\s+${ENGINE_ID}$$
  ^storage-type:\s${STORAGE_TYPE}\s+${STORAGE_TYPE_ACTIVE}$$
  ^Authentication\s+Protocol:\s+${AUTHENTICATION_PROTOCOL}$$
  ^Privacy\s+Protocol:\s+${PRIVACY_PROTOCOL}$$
  ^Group-name:\s+${GROUP_NAME}$$ -> Record
```

##### SAMPLE COMMAND OUTPUT

```
TESTSW#show snmp user 

User name: user_snmp1
Engine ID: 80000009030000451DEC1085
storage-type: nonvolatile        active
Authentication Protocol: SHA
Privacy Protocol: AES128
Group-name: managerpriv

User name: user_snmp2
Engine ID: 80000009030000451DEC1085
storage-type: nonvolatile        active
Authentication Protocol: SHA
Privacy Protocol: AES128
Group-name: managerpriv
```

##### STEPS TO REPRODUCE

```yaml
- name: Read snmp users
  ios_command:
    commands:
      - show snmp user
  register: res_snmpuser
  changed_when: false

- set_fact:
    snmp_users: "{{ res_snmpuser.stdout[0] | parse_cli_textfsm(textfsm_template_dir ~ '/cisco_ios_show_snmp_user.template') }}"
  changed_when: false
  delegate_to: localhost

- debug:
    msg: "{{ snmp_users | to_nice_yaml() }}"
  delegate_to: localhost
  changed_when: false
  tags: [print_action]

```

##### ACTUAL RESULTS

```
    -   AUTHENTICATION_PROTOCOL: SHA
        ENGINE_ID: 80000009030000451DEC1085
        GROUP_NAME: managerpriv
        PRIVACY_PROTOCOL: AES128
        STORAGE_TYPE: nonvolatile
[...]

PR #429, opened 2019-07-21
New template cisco_ios_show_snmp_user.template

jifox:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_ios_show_snmp_user.template

##### SUMMARY

see issue #390


jifox:
@FragmentedPacket 
Implemented sugested changes

templates/cisco_ios_show_snmp_user.template
```diff
@@ -0,0 +1,17 @@
+Value Required USER_NAME (\w+)
+Value ENGINE_ID (\w+)
+Value STORAGE_TYPE (\w+)
+Value STORAGE_TYPE_ACTIVE (\w+)
```
FragmentedPacket: I think we should change the `STORAGE_TYPE_ACTIVE` to `ACCESS_LIST` to capture `d+` as that appears from the Cisco docs as to what it is referring to:

https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/snmp/command/nm-snmp-cr-book/nm-snmp-cr-s1.html#wp8434359780

And then change L12 to:
`^storage-type:\s${STORAGE_TYPE}\s+active$$`

And add the following line right above:
`^storage-type:\s${STORAGE_TYPE}\s+active\s+access-list:\s+${ACCESS_LIST}$$`

jifox:
@FragmentedPacket 
Suggested changes are implemneted. Raw data modified to reflect the access_list parameter.

PR #442, opened 2019-07-24
Update cisco_xr_show_version.template

mspiez:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT

##### SUMMARY
Typo:

```
- ^\s+Build\sHost\s+:\s+${BUILD_HOST}
+ ^\s+Built\sHost\s+:\s+${BUILD_HOST}
```


IceN9ne:
This should probably be updated to `Buil[dt]`. I compared many different cXR and eXR versions across ASR and NCS devices and only found that line to be on eXR devices and specifically `Built` only on 6.5.3 ASR and 7.0.1 NCS-6000. 6.5.2 on ASR has the `Build` string. Looks like this is a recent change in the "Built" process.

I found it interesting that 6.5.3 on NCS-5500 has the original `Build` and doesn't match the change on ASR. Also interesting is eXR 7.0.1 on NCS-6000 has the updated `Built` string though.

mspiez:
Both options should be supported. 
Template updated to support two version of output: `Build: ` and `Built: `
```
+ ^\s+Buil[dt]\sHost\s+:\s+${BUILD_HOST}
```

Also added test files to support second case.

PR #445, opened 2019-08-02
Update cisco_ios_show_ip_eigrp_topology.template

thomasbridge74:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_ios_show_ip_eigrp_topology.template

##### SUMMARY

When I used the show ip eigrp topology template for the first time, I ran into errors.   There were two cases where the current template failed:

1. Where the route had an inaccessible FD rather than a number, the template choked on the input because it was matching FD on a number.  
2. More significantly, the script would also throw an error on routes that had been entered into the local EIGRP topology database via redistribution - it failed where the via was via Rstatic or Redistributed.

I've update the raw input file and the expected parsed input to test for these scenarios.    The result of this updated is an additional field (source) at the end that would be null for ordinary routes but will indicated whether the route is learned via Rstatic or Redistributed.


thomasbridge74:
> @thomasbridge74 Thanks for your submission!
> 
> Instead of changing the original **raw** and **parsed**, could you please add new ones that indicate your use case? This allows us to ensure backwards compatibility and have more (data) test cases

Hi - one of the "enhancements" I've made to the TextFSM template is to add an extra field (SOURCE) to parse if the next hop is learned through Static or Redistribution (and is blank if a normal EIGRP route).     I've reverted my repository to use the original raw file, but I will need to modify that parsed file to use the new field (which is blank for all the cases) as otherwise it will fail the unit testing.   Is this acceptable?

Thanks,

Thomas

PR #447, opened 2019-08-06
Update description and add Install/Usage

jmcgill298:


README.md
```diff
@@ -69,13 +132,15 @@ Start
 BGPTable
  ... omitted
 ```
+
 Each **state** should end with `^. -> Error`. This helps to ensure we're accounting for every line within the raw output for the command. This doesn't mean we have to capture all the data as a **Value**, but we do have to account for it.
```
FragmentedPacket: Should we format this a bit different to show that both are required:
```
^\s*
^. -> Error
```

README.md
```diff
@@ -87,109 +152,111 @@ Start
...
-You can test your changes locally within your Git branch before submitting a PR. This can be done by installing **tox** within your environment using pip and then running **tox** within the **ntc-templates** root directory:
-```
-(main) ~/ntc-templates[update-readme !]:~$ tox
-GLOB sdist-make: /home/myohman/ntc-templates/setup.py
-py36 inst-nodeps: /home/myohman/ntc-templates/.tox/dist/ntc_templates-1.1.0.zip
+- Navigate to `./tests/{{ vendor_os }}/{{ command_name }}/`
+- Create new `.raw` and `.parsed` files within the directory, preferrably with a name identifying why the data is unique:
+  * Existing raw: `./tests/cisco_ios/show_clock/cisco_ios_show_version.raw`
+  * New raw: `./tests/cisco_ios/show_clock/cisco_ios_show_version_stack_platforms.raw`
+  * Existing parsed: `./tests/cisco_ios/show_clock/cisco_ios_show_clock.parsed`
```
FragmentedPacket: Should the existing/new be the same show command here?

README.md
```diff
@@ -87,109 +152,111 @@ Start
...
+$ cat tests/cisco_ios/show_clock/cisco_ios_show_clock.raw
+*18:57:38.347 UTC Mon Oct 19 2015
+$ 
 ```
 
-#### YAML file containing expected parsed dictionary
+##### YAML file containing expected parsed dictionary
 
-The parsed dictionary must be in a dictionary format. All keys in the dictionary should be in all lowercase.
+The parsed file should match the data that is returned from the `parse_output` function discussed in the begining. Dictionary keys should be in lowercase.
```
FragmentedPacket: Fix typo: begining -> beginning

PR #450, opened 2019-08-07
Update cisco_ios_show_interfaces_description.template

geo5555:
the above template gives me an exception with ios 15.7 2911 router. I saw that this was due to the first two lines that appear when you do a show int desc: When I removed them, the parsing worked.

Load for five secs: 17%/12%; one minute: 22%; five minutes: 21%
Time source is NTP, 10:59:42.831 EEST Wed Aug 7 2019

So I modified the template and it works now

##### ISSUE TYPE

 - Bugfix Pull Request

##### SUMMARY

```
textfsm.TextFSMError: State Error raised. Rule Line: 9. Input Line: Load for five secs: 17%/12%; one minute: 22%; five minutes: 21%

after removing the ^.-> works ok
[['Em0/0', 'admin down', 'down', '']
['Gi0/0', 'up', 'up', '']
['Gi0/0.1', 'up', 'up', 'To Data Vlan']]
```

### DIFF
--- a/templates/cisco_ios_show_interfaces_description.template
+++ b/templates/cisco_ios_show_interfaces_description.template
@@ -6,9 +6,9 @@ Value DESCRIP (\S.*?)
 Start
   ^Interface\s+Status\s+Protocol\s+Description\s*$$ -> Begin
   ^\s*$$
-  ^. -> Error
+
 
 Begin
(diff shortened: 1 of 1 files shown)


jmcgill298:
Fixed in #525

PR #462, opened 2019-08-30
show_snmp_user Fix

mzbroch:
The existing template disallows for use of some characters, like a dash resulting in an error


FragmentedPacket:
@mzbroch Do you mind adding additional `.raw` and `.parsed` files so we have more data to test on that covers the bug you discovered?

FragmentedPacket:
Fixed by #532

PR #464, opened 2019-09-04
HP Comware `display counters (inbound|outbound) interface`

ad8-bdl:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
`hp_comware_display_counters_bound_interface.template`
HP Comware
`display counters inbound interface`
`display counters outbound interface`

##### SUMMARY
New template for a HP Comware switch.

Per discussion on [#networktocode](https://networktocode.slack.com/archives/C0HHE1FFH/p1567571217039500): 
The command display counters outbound interface has identical output form to the display counters inbound interface command. So this PR re-uses the same template and tests for both. Ideally the command regex would be `(i[[nbound]]|o[[utbound]])` but it seems TextFSM doesn't support alternation in this way, so `(\S+)` it is.

The template contains three states (header, table body, footer). It allows for the `Overflow` and `--` field values and will return Error if any record has overflowed; a simpler one-rule version of the template would just silently skip such records. Whether that's desirable or not probably depends on your use case...


templates/hp_comware_display_counters_bound_interface.template
```diff
@@ -0,0 +1,17 @@
+Value INTERFACE (\S+)
+Value PACKETS_TOTAL (\d+)
+Value PACKETS_BROADCAST (\d+)
+Value PACKETS_MULTICAST (\d+)
+Value PACKETS_ERROR (\d+)
+
+Start
+  ^\s*$$
+  ^Interface\s+Total\s+\(pkts\)\s+Broadcast\s+\(pkts\)\s+Multicast\s+\(pkts\)\s+Err\s+\(pkts\) -> InterfaceTable
+
```
jmcgill298:
```suggestion
  ^. -> Error
```
Should error in case a different format of the table is produced by the command output

templates/hp_comware_display_counters_bound_interface.template
```diff
@@ -0,0 +1,17 @@
...
+InterfaceTable
+  ^${INTERFACE}\s+${PACKETS_TOTAL}\s+${PACKETS_BROADCAST}\s+${PACKETS_MULTICAST}\s+${PACKETS_ERROR} -> Record
+  ^\s*$$ -> InterfaceTableFooter
+  ^. -> Error
```
jmcgill298:
```suggestion
  ^\s*$$
  ^\s*Overflow -> InterfaceTableFooter
```

templates/hp_comware_display_counters_bound_interface.template
```diff
@@ -0,0 +1,20 @@
...
+Value PACKETS_ERROR (\d+)
+
+Start
+  ^\s*$$
+  ^Interface\s+Total\s+\(pkts\)\s+Broadcast\s+\(pkts\)\s+Multicast\s+\(pkts\)\s+Err\s+\(pkts\) -> InterfaceTable
+  ^. -> Error
+
+InterfaceTable
+  ^${INTERFACE}\s+${PACKETS_TOTAL}\s+${PACKETS_BROADCAST}\s+${PACKETS_MULTICAST}\s+${PACKETS_ERROR} -> Record
+  ^\s*$$ -> InterfaceTableFooter
```
jmcgill298:
```suggestion
```

PR #466, opened 2019-09-05
Paloalto panos arp fix

dainok:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
paloalto_panos_show_arp_all.template

##### SUMMARY
Template now supports incomplete ARP requests.


templates/paloalto_panos_show_arp_all.template
```diff
@@ -1,6 +1,6 @@
 Value INTERFACE (\S+)
 Value IP (\S+)
-Value MAC ([0-9a-fA-F:]+)
+Value MAC (\S+)
```
jmcgill298: I would prefer not to have the parenthesis captured. I think `(\(?\S+\)?)` should work

templates/paloalto_panos_show_arp_all.template
```diff
@@ -1,14 +1,15 @@
 Value INTERFACE (\S+)
 Value IP (\S+)
-Value MAC ([0-9a-fA-F:]+)
+Value MAC ([0-9A-Fa-f:]+|incomplete)
 Value PORT (\S+)
 Value STATUS (\S+)
 Value TTL (\d+)
 
 Start
+  ^interface\s+ip address\s+hw address\s+port\s+status\s+ttl
```
FragmentedPacket: L9 add the **-> Start_record** to the end of it, and the move L10 to the first line under **Start_record**. We're looking to start recording if the headers are correct so we're looking for the correct data.

dainok:
@jmcgill298  should be ok now, isn't it? I don't see any pending change request.

PR #478, opened 2019-09-27
Add template cisco_nxos_show_ip_interface_brief

mullaneywt:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_nxos_show_ip_interface_brief

##### SUMMARY
Adds in support for the command "show ip interface brief" on Cisco NX-OS.
This command is supported in IOS, and the IPv6 version of this command is supported in NX-OS.
Tested on Cisco Nexus 7k and 9k physical hardware.


templates/cisco_nxos_show_ip_interface_brief.template
```diff
@@ -0,0 +1,9 @@
+Value Required INTF (\S+)
+Value Required IPADDR ([a-zA-Z0-9./]+)
+Value STATUS (\S+-\S+)
+Value LINK (\S+-\S+)
+Value PROTO (\S+-\S+)
+
+Start
+  ^${INTF}\s+${IPADDR}\s+${STATUS}/${LINK}/${PROTO} -> Record
```
FragmentedPacket: I think ${STATUS} and ${PROTO} need to be swapped around, but other than that everything looks good.

mullaneywt:
Updated the template and the parsed filed as requested.

templates/cisco_nxos_show_ip_interface_brief.template
mullaneywt: Updated!

PR #489, opened 2019-10-10
Cisco NXoS template for "show forwarding ipv4 route"

nnaukwal:
##### ISSUE TYPE

 - New Template Pull Request
##### COMPONENT

cisco_nxos_show_forwarding_ipv4_route.template
##### SUMMARY

This change is to add a new template which can structurize the output of "show forwarding ipv4 route" from a Cisco NXoS platform. 
The use cases can be many:
* Programmatically check FIB for a Prefix
* Collect and compare the FIB status before and after making a change to the network, to validate assumptions.

##### TESTING
* Tox Passes
```
<snip>
tests/test_structured_data_against_parsed_reference_files.py::test_raw_data_against_mock[tests/cisco_nxos/show_forwarding_ipv4_route/cisco_nxos_show_forwarding_ipv4_route.raw] PASSED                                     [ 89%]
<snip>
  py35: commands succeeded
  congratulations :)
```
* FIB output collected from multiple nxos 3548 in the network to broaden the available test dataset.


templates/cisco_nxos_show_forwarding_ipv4_route.template
```diff
@@ -0,0 +1,22 @@
+Value PREFIX (\d+\.\d+\.\d+\.+\d+\/\d+)
+Value Required,List NEXTHOP (Drop|Receive|Attached|\d+\.\d+\.\d+\.\d+)
+Value List INTERFACE (\S+)
+
+Start
+  ^Prefix.*Partial\s+Install\s*$$ -> FIB
```
FragmentedPacket: Do you mind updating the header to match at least each word within the header?

nnaukwal: Modified to be more explicit.

PR #493, opened 2019-10-11
Add ruckus_fastiron show arp template

QuasarKid:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

Added ruckus_fastiron_show_arp.template

##### SUMMARY

Added ruckus_fastiron show arp command template, accounting for both the Layer 2 and Layer 3 software versions (numberless and 1 raw outputs respectively).

Tried to minimize the amount of lines used for different headers by catching both OS lines if syntax was similar. Had to create multiple lines for the actual data capture as Layer 3 OS does not report VLAN information.

MAC Address has the potential to be 'None', wanted to capture that information as well.


QuasarKid:
Noticing that perhaps the raw and parsed outputs numbered 1 have quite a few lines, could probably do to remove some of those.

templates/ruckus_fastiron_show_arp.template
```diff
@@ -0,0 +1,18 @@
...
+Value Required PORT (\S+)
+Value STATUS (\S+)
+Value VLAN (\S+)
+
+Start
+  ^All\s+ARPs:\s+\d+,\s+maximum\s+capacity:\s+\d+
+  ^Total\s+(number\s+of\s+|)ARP\s+[Ee]ntries\s*:\s+\d+
+  ^Entries\s+in\s+default\s+routing\s+instance:
+  ^No\.\s+(IP\s+Address|IP)\s+(MAC\s+Address|MAC)\s+Type\s+Age\s+Port\s+Status(\s+VLAN|)
+  ^${NUMBER}\s+${IP_ADDRESS}\s+${MAC_ADDRESS}\s+${TYPE}\s+${AGE}\s+${PORT}\s+${STATUS}\s+${VLAN} -> Record
```
FragmentedPacket: Thanks for your submission!

What separate out each expected output into their own states since there is a small difference in the output (one will provide VLAN and the other doesn't.

```
Start
  ^No\.\s+IP\s+MAC\s+Type\s+Age\s+Port\s+Status\s+VLAN -> Type1
  ^No\.\s+IP\s+Address|\s+MAC\s+Addres\s+Type\s+Age\s+Port\s+Status -> Type2

Type1
  xxxx

Type2
  xxxx
```

QuasarKid: Changes made, 0 changes to output.

templates/ruckus_fastiron_show_arp.template
```diff
@@ -8,11 +8,21 @@ Value STATUS (\S+)
...
-  ^Total\s+(number\s+of\s+|)ARP\s+[Ee]ntries\s*:\s+\d+
-  ^Entries\s+in\s+default\s+routing\s+instance:
-  ^No\.\s+(IP\s+Address|IP)\s+(MAC\s+Address|MAC)\s+Type\s+Age\s+Port\s+Status(\s+VLAN|)
+  ^All\s+ARPs:\s+\d+,\s+maximum\s+capacity:\s+\d+ -> Layer2
+  ^Total\s+number\s+of\s+ARP\s+entries\s*:\s+\d+ -> Layer3
+  ^\s*$$
+  ^. -> Error 
+
+Layer2
+  ^No\.\s+(IP\s+Address|IP)\s+(MAC\s+Address|MAC)\s+Type\s+Age\s+Port\s+Status\s+VLAN
```
FragmentedPacket: Oops. I would have the transition these within **Start** to transition to the different states since these are the headers and we want to make sure we're keying off the correct data.

QuasarKid: Okay, I figured we should key off of the earliest thing we could determine which version of the OS we were on to eliminate checking lines we would never encounter. I will correct.

QuasarKid: Updated changes. My understanding is that once it changes state it doesn't go back unless we tell it to, which is why I moved the one line to the end of Layer2 state as it is after the data. Let me know if this is incorrect or there is a better way to handle a footer.

PR #494, opened 2019-10-11
Fix netflow and wccp lines

targuan:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_ios_show_ip_interface.template

##### SUMMARY
On a 6500, ip interface are reporting some netflow configuration that breaks the parsing of the show ip interface.
On a 3750, the WCCP line does not start with IPv4.

### DIFF
--- a/templates/cisco_ios_show_ip_interface.template
+++ b/templates/cisco_ios_show_ip_interface.template
@@ -48,8 +48,8 @@ Start
   ^\s+Policy
   ^\s+Network\s+address\s+
   ^\s+BGP
-  ^\s+Sampled Netflow
-  ^\s+IP (Routed|Bridged) Flow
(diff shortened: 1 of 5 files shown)


templates/cisco_ios_show_ip_interface.template
```diff
@@ -48,8 +48,10 @@ Start
   ^\s+Policy
   ^\s+Network\s+address\s+
   ^\s+BGP
+  ^\s+Sampled Netflow
```
FragmentedPacket: Do you mind removing the static spacing and use either **\s*** or **\s+**. Once those are implemented, we can get this merged.

targuan: Should be fixed.

PR #496, opened 2019-10-19
Add arista eos show ip helper

targuan:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT
arista_eos_show_ip_helper-address.template
OS: arista EOS
command: show ip helper-address

##### SUMMARY
This template allow the collection of the IP helpers from ip interfaces.


templates/arista_eos_show_ip_helper-address.template
```diff
@@ -0,0 +1,9 @@
+Value Required INTERFACE (\S+)
+Value List IP_HELPER (\d+\.\d+\.\d+\.\d+|\S+)
+
+Start
+ ^\S -> Continue.Record
```
FragmentedPacket: Can you add `^DHCP` above L5 and `^\s+DHCP\s+Smart` above L7? And then add the following under L8?
```
 ^\s*$$
 ^. -> Error
```

tests/test_index_load.py
```diff
@@ -0,0 +1,15 @@
+try:
```
FragmentedPacket: Do you mind removing this? We would want this as a separate PR

targuan: This was just a test, it shouldn't have been there.

templates/arista_eos_show_ip_helper-address.template
targuan: This has been corrected

templates/arista_eos_show_ip_helper-address.template
```diff
@@ -0,0 +1,12 @@
+Value Required INTERFACE (\S+)
+Value List IP_HELPER (\d+\.\d+\.\d+\.\d+|\S+)
+
+Start
+ ^DHCP
+ ^\S -> Continue.Record
```
FragmentedPacket: Can you change L6 to `^Interface -> Continue.Record`?

Once that is changed, I will merge it.

Thanks for your work on the template!

targuan: Uclvri kp LS

PR #504, opened 2019-10-22
Show 802 11a|b cleanair config

timjsmith24:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_wlc_ssh_show_802_11b_cleanair_config.template, .*, cisco_wlc_ssh, show 802.11b cleanair config

##### SUMMARY

```

```


jmcgill298:
This looks to be a superset of 802.11a output, does it make sense to have one template and fix the index to allow the command to support 80211.a|b?

timjsmith24:
I'm good with this. I have updated the template name and index to apply to both radios

timjsmith24:
I add the same fixes for the literal . in the index file to this template as well.

jmcgill298:
@timjsmith24 can you rebase this since I merged the other PR which added the fix in the test file? Otherwise this should be good.

jmcgill298:
Actually, I just realized that my updates were missing the EoL (`$$`) for matching blank lines, I updated the show 802.11ab template, but this one will also need to add that. I think this will also require adding more match lines to at least match the insignificant lines.

timjsmith24:
Not sure why this closed. But I reopened it. I am having a problem after rebasing. The other command 'show 802.11ab' keeps failing. I am not sure if it has to do with the '.' in the template name? If I remove the files for "cisco_wlc_ssh_show_802.11ab.template" or if I remove the files for "cisco_wlc_ssh_show_802.11ab_cleanair_config.template" it will pass but not when both templates and files are in.

jmcgill298:
Thanks @timjsmith24 The tests weren't really written to with templates being able to support multiple commands, so I renamed the directories and templates to just use `show 802.11a` but CliTable should still work with 802.11a and 802.11b.

PR #526, opened 2019-10-30
Adv 802dot11 channel

timjsmith24:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_wlc_ssh_show_advanced_802.11a_channel.template, .*, cisco_wlc_ssh, show advanced 802.11a channel

##### SUMMARY

```

```


templates/cisco_wlc_ssh_show_advanced_802.11a_channel.template
```diff
@@ -0,0 +1,50 @@
...
+  ^\s+DCA\sSensitivity\sLevel\:?\s?\.*\s${DCA_SENSITIVITY}s*$$
+  ^\s+DCA\s\S+\sChannel\sWidth\.*\s${DCA_WIDTH}s*$$
+  ^\s+DCA\sMinimum\sEnergy\sLimit\.*\s${DCA_MIN}s*$$
+  ^\s+Channel\sEnergy\sLevels
+  ^\s+Minimum
+  ^\s+Average
+  ^\s+Maximum
+  ^\s+Channel\sDwell\sTimes
+  ^\s+802\.11a\s5\sGHz\sAuto\-RF Channel Lists*$$
+  ^.*Allowed\sChannel\sList\.*\s${DCA_CHANNELS}s*$$ -> Channels
```
FragmentedPacket: This attempts to capture allowed channels on both 5 and 4.9Ghz.

I would say that we'd want to capture those in separate variables and then post process them in needed.

This would be to do state transitions on L37 instead of L38 and separate variables.

timjsmith24:
Ok I separated out the 5GHz channels. I didn't create separate variables for the 4.9GHz channels as those are not used

PR #539, opened 2019-11-11
show interface summary

timjsmith24:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

cisco_wlc_ssh_show_interface_summary.template, .*, cisco_wlc_ssh, show interface summary

##### SUMMARY

```

```


templates/cisco_wlc_ssh_show_interface_summary.template
```diff
@@ -0,0 +1,16 @@
+Value Filldown INT_COUNT (\d+)
+Value NAME (.+?)
+Value PORT (\w+|-|N\/A)
+Value VLAN (\d+|\w+|N\/A)
+Value IP_ADDR (([\d1-9]+\.?){4})
+Value TYPE (\w+)
+Value AP_MGR (\w+)
+Value GUEST (N\/A|\w+)
+
```
FragmentedPacket: Here is the changes to the template that I would make:
```
Value Filldown INT_COUNT (\d+)
Value Required NAME (\S+)
Value PORT (\S+)
Value VLAN (\S+)
Value IP_ADDR (([\d1-9]+\.?){4})
Value TYPE (\S+)
Value AP_MGR (\S+)
Value GUEST (\S+)

Start
  ^\s+Number\sof\sInterfaces\.*\s${INT_COUNT}s*$$
  ^Interface\s+Name\s+Port\s+Vlan\s+Id\s+IP\s+Address\s+Type\s+Ap\s+Mgr\s+Guest -> Type_One

Type_One
  ^-+\s
  ^${NAME}\s+${PORT}\s+${VLAN}\s+${IP_ADDR}\s+${TYPE}\s+${AP_MGR}\s+${GUEST}$$ -> Record
  ^\s*$$
  ^. -> Error
```
Changed most of the value's regex to `\S+` as it should suffice here and catch any non-whitespace on and not depend on if it's a word or not.

I also changed it to track the whole header and then transition state to `Type_One` if the header matches. This allows us to make sure we're capturing the the correct and expected output. If the headers or output changes, we would add another line to match it and transition to a different state to capture the changed output.

I also added `Required` to `Value NAME` to prevent us from having the non-filled in information within the `.parsed` file.

FragmentedPacket: Whoops. Could you add the following to the **Start** state:
```
  ^\s*$$
  ^. -> Error
```

PR #543, opened 2019-11-20
Enhancement: Use ".textfsm" extension for IDEs

### LINKED ISSUE #171: Using the .textfsm extension for templates
##### ISSUE TYPE
 - Feature Idea

##### TEMPLATE USING
all

##### SUMMARY
I think there's value in renaming the textfsm templates to use the `.textfsm` file extension as currently they cannot be syntax-highlighted in editors, or they are syntax-highlighted wrongly.

jmcgill298:
Fixes #171


jmcgill298:
Need to update the README, but would like to wait until #406 is merged

PR #554, opened 2019-12-05
New Template for Cisco IOS Log Messages

afoster213:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT

cisco_ios_show_logging, cisco_ios, show logging

##### SUMMARY
Created a template to parse the output from show logging.  The goal with this template is to be able to then review the parsed output for log messages that are warnings or higher.

PR #562, opened 2019-12-20
Cisco ASA Show Logging Template

afoster213:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_asa_show_logging, cisco asa, show logging

##### SUMMARY
Parses show logging output from Cisco ASAs.  Follows the standard set up on the Cisco IOS show logging template.


jmcgill298:
@afoster213 shouldn't this look more like #554 with different options depending on timestamp format?

afoster213:
The ASA doesn't have the option to add a syslog number or timezone.  Only options I'm seeing across my examples are with and without hostname.  Happy to add in additional cases if anyone has any additional examples.

afoster213:
@jmcgill298 I added another case for logs with no timestamps.  I'm not able to find any other cases though based on Cisco's documentation.  Here's the guide that I'm working off of.
https://www.cisco.com/c/en/us/support/docs/security/pix-500-series-security-appliances/63884-config-asa-00.html

jmcgill298:
This should still account for multiline messages using Continue.Record on new messages and then capturing the message contents on subsequent match lines.

afoster213:
@jmcgill298 Updated for multiline cases.

PR #568, opened 2020-01-03
New Template: ASA - Show arp

bourskov83:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT
cisco_asa_show_arp, cisco asa, show arp

##### SUMMARY
Parses show arp output from Cisco ASAs. Follows the standard set up on the Cisco IOS show ip arp template for shared fields.


bourskov83:
Hi all, this is my first Pull Request with Github, so hopefully I got it right... ;)

Otherwise, please let me know so I can learn from it...

/Bo

templates/cisco_asa_show_arp.textfsm
```diff
@@ -0,0 +1,11 @@
+Value Required INTERFACE (\S+)
+Value Required ADDRESS (\d+\.\d+\.\d+\.\d+)
+Value Required AGE (\S+)
+Value Required MAC (\S+)
+
+Start
+  ^\s*${INTERFACE}\s+${ADDRESS}\s+${MAC}\s+${AGE} -> Record
+  # Capture time-stamp if vty line has command time-stamping turned on
```
FragmentedPacket: How do you activate the time stamping on ASA? I've tried looking but unable to find the command?

Can you please also add `  ^\s*$$` above the `  ^. -> Error` line?

Also, the Load for and Time source entries, are these the ones we'd see if time stamping was enabled?

tests/cisco_asa/show_arp/cisco_asa_show_arp.raw
```diff
@@ -0,0 +1,5 @@
+        outside 1.2.3.4 444e.6df8.b97e 15
```
FragmentedPacket: If the other output for the time stamping is for the ASA, can you add another file with what that looks like to make sure we're able to capture the correct data with the two different formats?

FragmentedPacket:
Closed via #573 

@bourskov83 Thanks for your contribution!

PR #577, opened 2020-01-21
Bug Fix for cisco ios show mac address-table parser

kwrobert:
##### ISSUE TYPE

 - Bug fix

##### COMPONENT

**Template:** cisco_ios_show_mac-address-table.textfsm
**OS:** Cisco IOS
**Command:** `show mac address-table`

##### SUMMARY

Some versions of Cisco IOS (encountered running on a Cisco Catalyst
4510R+E stack in production) display two tables when running "show mac
address-table", one containing multicast entries and one containing
unicast entries. This commit just makes it so the parser doesn't fail
when hitting the multicast entries by skipping parsing of the multicast table entirely.

If parsing this would be of interest, it should be easy to add a new "multicast-type" state using similar rules as the other table types. If this is desirable, I'll gladly add it. 

The output doesn't actually change at all using this commit. It just prevents an exception from being raised by skipping attempts to parse the multicast table entirely.


kwrobert:
Ah, I see my tests are failing because I added an entry to the index file for parsing the `show vdc` command on Cisco NXOS devices for my own purposes. I don't have tests and everything for that yet, and didn't intend to include it in this pull request, I can fix that now.

templates/cisco_ios_show_mac-address-table.textfsm
```diff
@@ -18,6 +18,7 @@ TYPE1
   ^Displaying\s+entries
   ^Destination\s+Address\s+Address\s+Type\s+VLAN\s+Destination\s+Port
   ^\s*$$
+  ^Multicast Entries -> End
```
FragmentedPacket: Do you mind adding `\s+` between **Multicast Entries** rather than the static spacing. Once that is updated we can get this merged in.

kwrobert: Done!

PR #582, opened 2020-01-25
Jschulman cisco nxos allow missing transceiver type

jeremyschulman:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
Cisco NXOS show interface transceivers

##### SUMMARY
Resolves the following parse errors:

###### Interface missing a type value
````
Ethernet2/3
    transceiver is present
    type is
    name is Arista Networks
    part number is QSFP-40G-LR4-AR
    revision is --
    serial number is FOOBAZ
    nominal bitrate is 10300 MBit/sec
    Link length supported for SMF fiber is 10 km
    cisco id is --
    cisco extended id number is 196
````

###### Interface without a transceiver
````
Ethernet102/1/44
    transceiver is not applicable
````


jeremyschulman:
@jmcgill298 - thank you for the suggestion.  I've updated the template to make all values optional and included new tests.

jmcgill298:
Thanks for the fix @jeremyschulman

PR #583, opened 2020-01-27
New Template: Cisco IOS show ip bgp neighbor

kwrobert:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

Template: cisco_ios_show_ip_bgp_neighbor.textfsm
OS: Cisco IOS
Command: `show ip bgp neighbor

##### SUMMARY

Added support to parse output from `show ip bgp neighbor` because it contains some information not available in the output of `show ip bgp summary`. 

A lot more output could be retrieved from this command, I only took what I needed for now.


templates/cisco_ios_show_ip_bgp_neighbor.textfsm
```diff
@@ -0,0 +1,36 @@
+Value NEIGHBOR_IP (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
```
FragmentedPacket: Can you change some of these values to match for say `cisco_nxos_show_ip_bgp_neighbors`? Just so we can try and keep the values as close as possible independent of OS as much as we can.

tests/cisco_ios/show_ip_bgp_neighbor/cisco_ios_show_ip_bgp_neighbor.raw
```diff
@@ -0,0 +1,97 @@
+BGP neighbor is 20.30.255.14,  remote AS 1255, internal link
```
FragmentedPacket: Do you possibly have output with two BGP neighbors that we can use for tests? This just ensures we can properly parse the output with multiple neighbors. If not, you can always duplicate the neighbors, but change some of the information as well.

templates/cisco_ios_show_ip_bgp_neighbor.textfsm
kwrobert: Certainly, that makes a lot of sense. I should have thought to look at some other templates as an example!

Quick question, do you want me to amend my previous commit with these changes or create a new commit that adds them?

FragmentedPacket: Adding another commit is fine. We squash them when we merge.

tests/cisco_ios/show_ip_bgp_neighbor/cisco_ios_show_ip_bgp_neighbor.raw
kwrobert: Yup! Added a new test with a bunch of neighbors in the most recent commit.

templates/cisco_ios_show_ip_bgp_neighbor.textfsm
```diff
@@ -0,0 +1,36 @@
...
+Value REMOTE_ROUTER_ID (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
+Value BGP_STATE (\w+)
+Value LOCAL_HOST (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
+Value LOCAL_PORT (\d+)
+Value FOREIGN_HOST (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
+Value FOREIGN_PORT (\d+)
+
+Start
+  # Capture first line, which shows the BGP neighor and remote AS number
+  ^BGP\s+neighbor\s+is\s+${NEIGHBOR_IP},\s+remote\s+AS\s+${REMOTE_AS} -> Bgp_neighbor
```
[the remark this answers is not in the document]
kwrobert: Done

tests/cisco_ios/show_ip_bgp_neighbors/cisco_ios_show_ip_bgp_neighbors_multiple.raw
```diff
@@ -0,0 +1,1239 @@
...
+recorded by systems personnel.  In the course of monitoring individuals
+improperly using this system, or in the course of system maintenance, the
+activities of authorized users will also be monitored."
+-----------------------------------------------------------------------------
+HMH
+nsxe-core-rtr2
+=============================================================================
+BGP neighbor is 175.135.172.146,  remote AS 65180, internal link
+ Description: nsxe-core-rtr1
+ Member of peer-group HMCO-core for session parameters
```
FragmentedPacket: I would update any values that are using **\w+** and change them to **\S+**. This will capture the `peer_group: 'HMCO-core'` within the parsed file.

FragmentedPacket:
@kwrobert Just an FYI you can **pip install tox** and then just run **tox** command for it to run the tests locally before pushing up and waiting for Travis CI

FragmentedPacket:
Closed by #590 

Thanks for your work on this!

PR #584, opened 2020-01-28
BugFix: Cisco show switch detail with mac persistency wait time

msom:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_ios_show_switch_detail.textfsm,  Catalyst L3 Switch Software Version 16.9.4, show switch detail

##### SUMMARY
The command might output the mac persistency wait time in which case the existing template errored.


templates/cisco_ios_show_switch_detail.textfsm
```diff
@@ -7,6 +7,7 @@ Value STATE (.+?)
 
 Start
   ^Switch/Stack
+  ^Mac persistency wait time
```
FragmentedPacket: Just a really small ask, but instead of the static spacing, can you change to `\s+` between the words on the newly added line?

Once that is fixed, I will get this merged

msom: Done

PR #592, opened 2020-02-09
Fixed regex for fortinet bgp template and added new template standards

ccorvese:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT

* Name of template: `fortinet_fortios_get_router_info_bgp_summary.textfsm`
* OS: `Fortinet FortiOS`
* Command: `get router info bgp summary`

##### SUMMARY

This fixes an issue with some of the output returned from the `get router info bgp summary` command on `fortinet fortios`. Sometimes time returned output from the device has ":" in the UP_DOWN portion (i.e. 05:48:47). The previous regex did not account for the ":" character. I have updated that with this PR

Additionally, I have updated the template to account for all potential lines of output from the command, finishing on an error state if a line is presented without a regex match.

### DIFF
--- a/templates/fortinet_fortios_get_router_info_bgp_summary.textfsm
+++ b/templates/fortinet_fortios_get_router_info_bgp_summary.textfsm
@@ -1,9 +1,15 @@
 Value BGP_NEIGH (\d+?\.\d+?\.\d+?\.\d+?)
 Value NEIGH_AS (\d+)
-Value UP_DOWN (\w+)
+Value UP_DOWN ((?:\w|:)*)
 Value STATE_PFXRCD (\w+)
 
 Start
+  ^BGP router identifier.*$$
+  ^BGP table version is.*$$
+  ^\d*\s+BGP AS-PATH entries.*$$
+  ^\d*\s+BGP community entries.*$$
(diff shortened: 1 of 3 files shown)


templates/fortinet_fortios_get_router_info_bgp_summary.textfsm
```diff
@@ -1,9 +1,15 @@
 Value BGP_NEIGH (\d+?\.\d+?\.\d+?\.\d+?)
 Value NEIGH_AS (\d+)
-Value UP_DOWN (\w+)
+Value UP_DOWN ((?:\w|:)*)
```
FragmentedPacket: Would it be worth it to change to **\S+** to capture anything non-whitespace? This should hopefully prevent us from having to worry too much about any format changes for it.

templates/fortinet_fortios_get_router_info_bgp_summary.textfsm
```diff
@@ -1,9 +1,15 @@
...
 Value STATE_PFXRCD (\w+)
 
 Start
+  ^BGP router identifier.*$$
```
FragmentedPacket: Do you mind changing the static spaces to **\s+**?

ccorvese:
@FragmentedPacket, thank you for your feedback. I have made the requested changes. Please let me know if there's anything else that may need improvement.

FragmentedPacket:
@ccorvese Thanks for another contribution!

PR #594, opened 2020-02-15
add show arp for aruba_os

dmwcode:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

Template:  aruba_os_show_arp.textfsm

##### SUMMARY

add show arp command for aruba_os

```

```


templates/aruba_os_show_arp.textfsm
```diff
@@ -0,0 +1,6 @@
+Value IP_ADDRESS (\d+.\d+.\d+.\d+)
+Value HARDWARE_ADDRESS (\w+.\w+.\w+.\w+.\w+.\w+)
+Value INTERFACE (\w+)
```
FragmentedPacket: Do you mind aligning the values with other ARP related templates? Such as `cisco_ios_show_ip_arp.textfsm`

```suggestion
Value ADDRESS (\d+.\d+.\d+.\d+)
Value MAC (\S+)
Value INTERFACE (\S+)
```

templates/aruba_os_show_arp.textfsm
```diff
@@ -0,0 +1,6 @@
...
+
+Start
+ ^\s+Internet\s+${IP_ADDRESS}\s+${HARDWARE_ADDRESS}\s+${INTERFACE} -> Record
```
FragmentedPacket: We also want to track headers within the output to make sure we're capturing the correct data. If the headers were to ever change, we can easily create different state transitions to account for the changes in the data.
```suggestion
Start
 ^IPV4\s+ARP\s+Table
 ^-+
 ^Protocol\s+IP\s+Address\s+Hardware\s+Address\s+Interface\s+Age
 ^\s+Internet\s+${IP_ADDRESS}\s+${HARDWARE_ADDRESS}\s+${INTERFACE} -> Record
 ^\s*$$
 ^. -> Error
```

FragmentedPacket:
No problem. I would change the MAC_ADDRESS regex match to something else that matches what some other templates are using to capture the mac addresses and change interface to `\S+` and then I think everything else looks good and you can add the changes to your branch and push a new commit up

tests/aruba_os/show_arp/aruba_os_show_arp.yml
```diff
@@ -0,0 +1,38 @@
+---
```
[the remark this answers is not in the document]
dmwcode: I think it should be all set.  Thank.

PR #595, opened 2020-02-17
Added Checkpoint GAIA show arp dynamic all template

dys152:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Template: checkpoint_gaia_show_arp_dynamic_all.textfsm

##### SUMMARY
New functionality as no previous template. Output from the Checkpoint GAIA O/S "show arp dynamic all" command is very limited and only shows IP and MAC address (it doesn't contain interface or age of entry unfortunately, even though this is available via the underlying Linux networking commands)


templates/checkpoint_gaia_show_arp_dynamic_all.textfsm
```diff
@@ -0,0 +1,8 @@
+Value ADDRESS (\d+\.\d+\.\d+\.\d+)
+Value MAC (\S+)
+
+Start
+  ^Dynamic\sArp\sParameters
+  ^IP\sAddress\s+Mac\sAddress
+  ^${ADDRESS}\s+${MAC} -> Record
+  ^. -> Error
```
FragmentedPacket: Looks great! If you don't mind changing the following and then we get this merged.
```suggestion
 ^\s*$$
 ^. -> Error
```

PR #596, opened 2020-02-19
Cisco XR - admin show environment power - template

hijm:
Cisco XR - admin show environment power - template

 - New Template Pull Request

##### COMPONENT
cisco_xr_admin_show_environment_power.textfsm
IOS XR
admin show environment power

##### SUMMARY
Inclusion of IOS XR admin show environment power for DC power supplies. There is a difference for AC power supplies. The number of columns are different.


templates/cisco_xr_admin_show_environment_power.textfsm
```diff
@@ -0,0 +1,8 @@
+Value POWER_MODULE (\d+\/PM\d+)
```
FragmentedPacket:
```suggestion
Value PS_MODULE (\S+)
```

So the value name matches an existing show environment power (`cisco_ios_show_environment_power_all.textfsm`)

Is there any other values for status other than **OK** or **Failed**? We can change the regex around for the `Parse` matching from all the space, non whitespace to just .+? and have the regex for status be `(OK|Failed)`.

hijm:
@FragmentedPacket. Do you have other change requests? Is it clear why I didn't accept your change request?

PR #597, opened 2020-02-20
add huawei vrp display version

lutfisan:
##### Huawei VRP - display version - template

 - New Template Pull Request

##### COMPONENT

huawei_vrp_display_version.textfsm
HUAWEI VRP
display version

##### SUMMARY

```

```


templates/huawei_vrp_display_version.textfsm
```diff
@@ -0,0 +1,9 @@
+Value VRP_VERSION (\S+)
+Value PRODUCT_VERSION (.+)
+Value MODEL (.+)
```
FragmentedPacket:
```suggestion
Value PRODUCT_VERSION (\S+)
Value MODEL (\S+)
```
Keeping uptime as **.+** makes sense, but I think the rest should just match non-whitespace so they shouldn't catch other things.

```suggestion
^HUAWEI\s${MODEL}\suptime\sis\s${UPTIME}$$
```
This should help prevent catching any additional whitespaces and shows we're expecting end of line for this

templates/huawei_vrp_display_version.textfsm
```diff
@@ -0,0 +1,9 @@
+Value VRP_VERSION (\S+)
+Value PRODUCT_VERSION (.+)
+Value MODEL (.+)
+Value UPTIME (.+)
+
+
+Start
+  ^.*software,\sVersion\s${VRP_VERSION}\s\(${PRODUCT_VERSION}\)
```
FragmentedPacket: Do you mind changing the each `\s` to `\s+` just in case they change any whitespace between versions and it won't break the template.

templates/index
```diff
@@ -1,368 +1,369 @@
-# First line is the header fields for columns and is mandatory.
```
FragmentedPacket: Can you please revert whatever formatting was done on this and just make sure your new template line is the only change?

lutfisan: sorry my bad

templates/huawei_vrp_display_version.textfsm
lutfisan: I use `.+` because `ATN 910C-B` in huawei_vrp_display_version2.raw have whitespace

templates/huawei_vrp_display_version.textfsm
lutfisan: done

templates/huawei_vrp_display_version.textfsm
FragmentedPacket: Sounds good. Thanks for the example.

PR #605, opened 2020-03-02
New Template: cisco_s300_show_version

wdennis:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_s300_show_version

##### SUMMARY
- Parses "show version" output from both version 1.x.y.z and 2.x.y.x Cisco Small Business NOS
- Both provide "Software version" value (`sw_version`) which is the "Firmware Version" on Cisco software download page
- ver 1.x.y.z firmware also provides separate "Boot version" (`boot_version`) and "Hardware version" (`hw_version`) values (present as tested on SG300-[##] switches), but not on ver 2.x.y.z firmware

### DIFF
--- a/templates/cisco_s300_show_version.textfsm
+++ b/templates/cisco_s300_show_version.textfsm
@@ -3,12 +3,12 @@ Value BOOT_VERSION ([1-9]\.[0-9]\.[0-9].[0-9]*)
 Value HW_VERSION (V0[1-9])
 
 Start
-  ^SW\sversion\s+${SW_VERSION}\s.* -> FWVER1
+  ^SW\s+version\s+${SW_VERSION}\s.* -> FWVER1
   ^Active-image:\s.* -> FWVER2
 
 FWVER1
-  ^Boot\sversion\s+${BOOT_VERSION}\s.*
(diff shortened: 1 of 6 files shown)


templates/cisco_s300_show_version.textfsm
```diff
@@ -0,0 +1,21 @@
+Value SW_VERSION ([1-9]\.[0-9]\.[0-9].[0-9]*)
+Value BOOT_VERSION ([1-9]\.[0-9]\.[0-9].[0-9]*)
+Value HW_VERSION (V0[1-9])
+
+Start
+  ^SW\sversion\s+${SW_VERSION}\s.* -> FWVER1
```
FragmentedPacket: Do you minding changing all the `\s` to `\s+` just in case Cisco changes whitespace between versions (not unheard of) and since we're expecting at least one space is the reason for the `+`

wdennis: Although I don't see why Cisco would increase the whitespace between "SW version", "Boot version" and "HW version" in the v1.x.y.z output, no harm covering the bases I guess :)

PR #606, opened 2020-03-02
New Template: cisco_ios_show_ip_ospf_database_router

ChristopherJHart:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_ios_show_ip_ospf_database_router

##### SUMMARY
* Parse relevant information from output of `show ip ospf database router` on Cisco IOS.
* Output pulled from an IOSv node running IOS 15.7(3)M3.

Let me know if any changes are needed - thank you!


ChristopherJHart:
@jmcgill298 Apologies! I've modified the TextFSM template states accordingly.

Thank you!

templates/cisco_ios_show_ip_ospf_database_router.textfsm
```diff
@@ -0,0 +1,48 @@
+Value Filldown ROUTER_ID (\d+\.\d+\.\d+\.\d+)
+Value Filldown PROCESS_ID (\d+)
+Value Filldown AREA (\d+\.\d+\.\d+\.\d+|\d+)
+Value Filldown LSA_AGE (\d+)
+Value Filldown LSA_OPTIONS (.*)
+Value Filldown LSA_TYPE (.*)
+Value Filldown LSA_ID (\d+\.\d+\.\d+\.\d+)
+Value Filldown LSA_ADV_ROUTER (\d+\.\d+\.\d+\.\d+)
+Value Filldown LSA_SEQ_NUMBER (\d+)
```
FragmentedPacket:
```suggestion
Value Filldown LSA_SEQ_NUMBER (\S+)
```

The raw data is: LS Seq Number: 800000D7
And the parsed data is: 800000

templates/cisco_ios_show_ip_ospf_database_router.textfsm
```diff
@@ -0,0 +1,48 @@
...
+Value Filldown LSA_CHECKSUM (\S+)
+Value Filldown LSA_LENGTH (\d+)
+Value Filldown LSA_NUM_LINKS (\d+)
+Value LS_LINK_TYPE ((a Stub Network)|(a Transit Network))
```
FragmentedPacket: Do you want to just capture whether it's a stub or transit rather than the whole `a Stub Network`?

tests/cisco_ios/show_ip_ospf_database_router/cisco_ios_show_ip_ospf_database_router.yml
```diff
@@ -0,0 +1,206 @@
+---
+parsed_sample:
```
FragmentedPacket: It appears the template still needs some work done. Comparing the `raw` and `yml` files, the LSA age appears to be off.

templates/cisco_ios_show_ip_ospf_database_router.textfsm
ChristopherJHart: I could be persuaded either way here, so I will defer to what the community thinks is best. 

Philosophically, I have always thought that a TextFSM template should convert *exactly* what the CLI says into structured output. So on one hand, I think "a Stub Network" and "a Transit Network" should be returned instead of "Stub" or "Transit".

On the other hand, I cannot think of a scenario where a user would be inconvenienced by the fact that this template returns "Stub", "Transit", or "point-to-point" (which I'm adding in a commit momentarily) instead of the full text. This could also reduce future LOCs that would filter out the unnecessary bits of what is returned.

It's an easy change either way, so I would say let me know which way you prefer, and I can modify it accordingly.

tests/cisco_ios/show_ip_ospf_database_router/cisco_ios_show_ip_ospf_database_router.yml
ChristopherJHart: Apologies - this should be fixed through 34dd6df.

I fixed this by introducing a stage change from the LSAs state to the LSAInfo state at the end of the LSAs state by triggering on and recording "LS age:". This is my first functional TextFSM template, and this feels like a bit of a hack to me. I am not sure if this is "acceptable" or not, so let me know if there's a better way to do this.

templates/cisco_ios_show_ip_ospf_database_router.textfsm
```diff
@@ -0,0 +1,48 @@
...
+Value Filldown LSA_LENGTH (\d+)
+Value Filldown LSA_NUM_LINKS (\d+)
+Value LS_LINK_TYPE ((a Stub Network)|(a Transit Network)|(another Router \(point-to-point\)))
+Value LS_LINK_ID (\d+\.\d+\.\d+\.\d+)
+Value LS_LINK_DATA (\d+\.\d+\.\d+\.\d+)
+Value LS_MTID_METRICS (\d+)
+Value LS_TOS_0_METRICS (\d+)
+
+Start
+  ^\s+OSPF Router with ID \(${ROUTER_ID}\) \(Process ID ${PROCESS_ID}\)
```
FragmentedPacket: Do you mind converting all the static spaces to `\s+`. This just helps to not break the template if the vendor changes the spacing.

tests/cisco_ios/show_ip_ospf_database_router/cisco_ios_show_ip_ospf_database_router.yml
FragmentedPacket: This should work

templates/cisco_ios_show_ip_ospf_database_router.textfsm
```diff
@@ -0,0 +1,48 @@
...
+Value Filldown LSA_AGE (\d+)
+Value Filldown LSA_OPTIONS (.*)
+Value Filldown LSA_TYPE (.*)
+Value Filldown LSA_ID (\d+\.\d+\.\d+\.\d+)
+Value Filldown LSA_ADV_ROUTER (\d+\.\d+\.\d+\.\d+)
+Value Filldown LSA_SEQ_NUMBER (\S+)
+Value Filldown LSA_CHECKSUM (\S+)
+Value Filldown LSA_LENGTH (\d+)
+Value Filldown LSA_NUM_LINKS (\d+)
+Value LS_LINK_TYPE ((a Stub Network)|(a Transit Network)|(another Router \(point-to-point\)))
```
FragmentedPacket: You can probably just change this regex to `.+?` and then on the line that captures `LS_LINK_TYPE` end it with `$$`. Just simplifies the regex a bit. And with this new output, we should just capture everything rather than `Stub/Transit` since that doesn't work for all the available output

templates/cisco_ios_show_ip_ospf_database_router.textfsm
```diff
@@ -0,0 +1,48 @@
+Value Filldown ROUTER_ID (\d+\.\d+\.\d+\.\d+)
+Value Filldown PROCESS_ID (\d+)
+Value Filldown AREA (\d+\.\d+\.\d+\.\d+|\d+)
+Value Filldown LSA_AGE (\d+)
+Value Filldown LSA_OPTIONS (.*)
```
FragmentedPacket: Can you change these from `.*` to `.+` since we're looking for one or more rather than 0 or more.

ChristopherJHart: Fixed via 22c9495

templates/cisco_ios_show_ip_ospf_database_router.textfsm
ChristopherJHart: Fixed via 87a1f37

templates/cisco_ios_show_ip_ospf_database_router.textfsm
ChristopherJHart: Opted to go with catching all text after support for P2P network types was added. Resolved via 87a1f37.

templates/cisco_ios_show_ip_ospf_database_router.textfsm
ChristopherJHart: Changes made via 87a1f37

PR #608, opened 2020-03-04
Fix regex: Update VyOS template to match addresses without netmask de…

sliddjur:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
vyos show interfaces

##### SUMMARY
Fix regex: Update VyOS template to match addresses without netmask de…


templates/vyatta_vyos_show_interfaces.textfsm
```diff
@@ -1,5 +1,5 @@
 Value INTERFACE (\S+)
-Value List IP_ADDRESS ((?:[A-Fa-f0-9:\.]+\/\d+)|-)
+Value List IP_ADDRESS ((?:[A-Fa-f0-9:\.]+)(\/\d+)?|-)
```
FragmentedPacket: Is this something we can simplify the regex for and use `\S+` instead?

Also, could you provide a new `.raw` for this use case and the accompanying `.yml`?

sliddjur: @FragmentedPacket ok I updated the regex to do `\S+` and also updated the raw and yml files.

PR #609, opened 2020-03-04
New Template: hp comware show ip interface

sliddjur:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
hp_comware_show_ip_interface, hp comware, show ip interface

##### SUMMARY
Add a new template.

### DIFF
--- a/templates/hp_comware_display_ip_interface.textfsm
+++ b/templates/hp_comware_display_ip_interface.textfsm
@@ -1,19 +1,42 @@
-Value INTERFACE (\S+)
-Value PHY (UP|DOWN|Administratively DOWN)
-Value PROTOCOL (UP(\(spoofing\))?|DOWN)
-Value List IP_ADDRESS (\S+)
+Value INTF (\S+)
+Value LINE_STATUS (UP|DOWN|Administratively DOWN)
+Value PROTOCOL_STATUS (UP(\(spoofing\))?|DOWN)
+Value List IPADDR (\S+)
 Value MTU (\d+)
 
 
 Start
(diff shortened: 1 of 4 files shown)


templates/hp_comware_display_ip_interface.textfsm
```diff
@@ -0,0 +1,19 @@
+Value INTERFACE (\S+)
```
FragmentedPacket:
```suggestion
Value INTF (\S+)
Value LINE_STATUS (.+)
Value PROTOCOL_STATUS (\S+)
Value List IPADDR (\S+)
Value MTU (\d+)


Start
  ^\S+\s+current\s+state -> Continue.Record
  ^${INTF}\s+current\s+state:\s+${LINE_STATUS}
  ^Line\s+protocol\s+current\s+state:\s+${PROTOCOL_STATUS}
  ^Internet\s+Address\s+is\s+${IPADDR}\s+Primary
  ^Internet\s+Address\s+is\s+${IPADDR}\s+Sub
  ^The\s+Maximum\s+Transmit\s+Unit:\s+${MTU}\s+bytes
  ^\s*$$
  ^. -> Error
```
I don't think I'd worry about a state transition for this template and use the above. I did change the values to align with templates of the same type such as `cisco_ios_show_ip_interface`.

Next you will need to add the other data to account for. For example:
```
  ^input\s+packets
  ^output\s+packets
```

sliddjur: ok, added all data and renamed the values :)

PR #612, opened 2020-03-06
Adding vlan name to access vlan and native vlan on NXOS

daanvdsanden:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_nxos_show_interfaces_switchport

##### SUMMARY
Adding the vlan name for the access vlan and native vlan. This way no extra command is needed when only vlan name of the vlans of access ports is needed.

Two extra field are available now: access_vlan_name and native_vlan_name


templates/cisco_nxos_show_interfaces_switchport.textfsm
```diff
@@ -3,7 +3,9 @@ Value SWITCHPORT (.+)
 Value SWITCHPORT_MONITOR (.+)
 Value MODE (.+)
 Value ACCESS_VLAN (\d+)
+Value ACCESS_VLAN_NAME (.*)
```
FragmentedPacket: Could you change the regex to `\.+` since I think we're expecting something between the brackets rather than possibly nothing.

daanvdsanden: Agreed will push the updates in a few minutes.

FragmentedPacket:
Awesome! Thanks for these enhancements!

PR #623, opened 2020-03-15
New Template: hp_comware_display_ip_routing-table & hp_comware_display_device_manuinfo

xdai555:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

1. hp_comware_display_ip_routing-table,hp_comware,display ip routing-table
2. hp_comware_display_device_manuinfo,hp_comware,display device manuinfo

##### SUMMARY

1. hp_comware_display_ip_routing-table  
Reference: cisco_ios_show_ip_route
2. display_device_manuinfo  
The following device types have been tested:  
    - Switches:
      - S125 series (Comware v7)
      - S6800 series (Comware v7)
      - S5560 (Comware v7)
      - S5500 (Comware v5)
      - S58 series (Comware v7 Simware)
    - Routers:
      - SR66 series (Comware v5 old version)
      - SR88 series (Comware v5)
      - MSR36 series (Comware v7 Simware)


xdai555:
The 'display ip routing-table' template needs to match each line, add '^. -> Error' is useless.
And 'display device manuinfo' update template to match every hardware even if it doesn't have a serial number.

jmcgill298:
> The 'display ip routing-table' template needs to match each line, add '^. -> Error' is useless.
> And 'display device manuinfo' update template to match every hardware even if it doesn't have a serial number.

This is not useless, as the intent is to catch output that is different from the known output and report back which line is not being captured. CLI output is known to vary between platforms, OS versions, and rare conditions that alter table data. Even knowing that this would account for all data today, it doesn't ensure that future updates will comply.

FragmentedPacket:
@xdai555 As Jacob said, those are something we require and it does help ensure when the output changes, it will throw an error and help locate the difference in output when trying to troubleshoot the template in the future.

Please update the templates to NTC Templates standard and then we'll get this merged in.

xdai555:
@FragmentedPacket @jmcgill298 I have some problems these days and now they have been solved. 
Sorry for keeping you waiting.

templates/hp_comware_display_device_manuinfo.textfsm
```diff
@@ -0,0 +1,21 @@
...
+  ^\s?Chassis\s+${CHASSIS_ID}
+  ^\s?${SLOT_TYPE}\s+${SLOT_ID}
+  ^\s?DEVICE_ID
+  ^\s?The\s+operation\s+is\s+not\s+supported\s+ -> Record Start
+  ^\s?DEVICE_NAME\s*:\s*${DEVICE_NAME}
+  ^\s?DEVICE_SERIAL_NUMBER\s*:\s*${DEVICE_SERIAL_NUMBER}
+  ^\s?MAC_ADDRESS\s*:\s*${MAC_ADDRESS}
+  ^\s?MANUFACTURING_DATE\s*:\s*${MANUFACTURING_DATE}
+  ^\s?VENDOR_NAME\s*:\s*${VENDOR_NAME} -> Record Start
+  ^. -> Error
```
FragmentedPacket, line 21: Do you mind adding `  ^\s*$$` above the error line?

Do you mind also changing the beginning `\s?` to `\s*` just in case they do change to more than one space at the beginning?

Those should be the last changes and then we can get these merged in.

xdai555: I will follow the format next time.
Practice makes perfect. ^_^

jmcgill298:
Thanks @xdai555

PR #626, opened 2020-03-18
Updates Cisco ASA `show vpn-sessiondb anyconnect`

jvanderaa:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

Cisco ASA - cisco_asa_show_vpn-sessiondb_anyconnect

##### SUMMARY

- Added the Session Index as a requirement to get the appropriate capture
- Updated to have a Start with the Anyconnect, then continuing onto the connection information as Anyconnect type is not repeated per new output in raw2
- Added raw2
- Updated syntax for raw1


tests/cisco_asa/show_vpn-sessiondb_anyconnect/cisco_asa_show_vpn-sessiondb_anyconnect.yml
```diff
@@ -24,4 +24,4 @@ parsed_sample:
     vlan_mapping: "N/A"
     vlan: "none"
     audt_sess_id: "ac1063fe003fa0005e715555"
-    security_grp: "none"
+    security_grp: "none                   "
```
jmcgill298: Lets fix captuing trailing spaces

tests/cisco_asa/show_vpn-sessiondb_anyconnect/cisco_asa_show_vpn-sessiondb_anyconnect.raw
```diff
@@ -12,4 +12,4 @@ Duration     : 2h:21m:21s
 Inactivity   : 0h:00m:00s
 VLAN Mapping : N/A                    VLAN         : none
 Audt Sess ID : ac1063fe003fa0005e715555
-Security Grp : none
\ No newline at end of file
+Security Grp : none
```
FragmentedPacket: I just checked with someone who has ASA and AnyConnect and they aren't seeing this whitespace in the raw output so I think we're safe to get rid of it the `.raw` files and you can change the regex back to what it was if you'd like.

`  ^\s*Security\s+Grp\s+:\s+${SECURITY_GRP}$$ -> Record`
Should suffice

jvanderaa: The one output does still have spaces in it. Not sure if that is a slack thing or not. Checked and the change to use a `\s*` for greedy capture should account for both cases (whitespace and no whitespace). 

Updated the whitespace to have a mix of one with and 2 without in the single test.

PR #632, opened 2020-03-26
Bugfix: cisco asa show nat

joewesch:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_asa_show_nat.textfsm, cisco_asa

##### SUMMARY
The service mappings in Manual and After-auto NAT do not require a protocol (tcp, udp, sctp) defined, but in the Auto NAT section they do. Previously, the `service_real` was getting set to this protocol and the `service_mapped` was getting set to the real service and the mapped service was getting dropped.


jmcgill298:
Thanks @joewesch

PR #654, opened 2020-04-20
Adds support for Cisco FTD

micahculpepper:
Adds template for parsing ping, and recycles the existing ASA template for parsing arp. See also: https://github.com/ktbyers/netmiko/pull/1681
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_ftd ping, and cisco_ftd show arp

##### SUMMARY

Adds support for cisco FTD ping and show arp commands. The show arp command shares the ASA's template.

tox output:
```
=================================================================== 846 passed in 19.76s ====================================================================
__________________________________________________________________________ summary __________________________________________________________________________
  py36: commands succeeded
  congratulations :)
```


FragmentedPacket:
I don't have an ASA to test on, but is the biggest thing between FTD and ASA just the paging and that's the reason for the new driver in Netmiko? Does the output differ? I'm wondering if we should go the route you did with the ARP command and just have the OS as either `cisco_(asa|ftd)` within the index file if that is the only difference.

micahculpepper:
The FTD CLI is a chimera of ASA and non-ASA commands and output. I think it makes sense to have a separate folder for FTD templates since it will surely need some of its own.

However, in the case of ping output, it does look just like ASA output, so the same template could be used for both. I was actually a bit surprised that there was no template for ASA pings to begin with. But I wanted to get FTD support off the ground without worrying about whether my new template would work for all ASAs, which have much broader support in ntc_templates already. In other words, doing it this way seemed less risky.

templates/index
```diff
@@ -132,10 +132,12 @@ cisco_asa_show_version.textfsm, .*, cisco_asa, sh[[ow]] ver[[sion]]
 cisco_asa_show_route.textfsm, .*, cisco_asa, sh[[ow]] ro[[ute]]
 cisco_asa_show_xlate.textfsm, .*, cisco_asa, sh[[ow]] x[[late]]
 cisco_asa_show_name.textfsm, .*, cisco_asa, sh[[ow]] nam[[e]]
-cisco_asa_show_arp.textfsm, .*, cisco_asa, sh[[ow]] arp
+cisco_asa_show_arp.textfsm, .*, cisco_(asa|ftd), sh[[ow]] arp
 cisco_asa_show_nat.textfsm, .*, cisco_asa, sh[[ow]] nat
 cisco_asa_dir.textfsm, .*, cisco_asa, dir
 
+cisco_ftd_ping.textfsm, .*, cisco_ftd, ping
```
FragmentedPacket, line 139: Do you mind just adding this command within ASA as well? And then using the same logic you did for `show_arp`?

templates/cisco_ftd_ping.textfsm
```diff
@@ -0,0 +1,18 @@
...
+Value Required RTT_AVG (\d+)
+Value Required RTT_MAX (\d+)
+
+
+Start
+  ^Type\s+escape\s+sequence\s+to\s+abort.
+  ^Sending\s+${SENT_QTY},\s+${SENT_TYPE}\s+to\s+${DESTINATION},\s+timeout\s+is\s+${TIMEOUT}\s+seconds:
+  ^${RESPONSE_STREAM}
+  ^Success\s+rate\s+is\s+${SUCCESS_PCT}\s+percent\s+\(${SUCCESS_QTY}/\d+\),\s+round-trip\s+min/avg/max\s+=\s+${RTT_MIN}/${RTT_AVG}/${RTT_MAX}\s+ms
+  ^. -> Error
```
FragmentedPacket, line 18: Do you mind adding `  ^\s*$$` right above the error line?

micahculpepper: Done.

templates/index
micahculpepper: Updated the line so it works for both platforms.

templates/index
```diff
@@ -132,10 +132,12 @@ cisco_asa_show_version.textfsm, .*, cisco_asa, sh[[ow]] ver[[sion]]
 cisco_asa_show_route.textfsm, .*, cisco_asa, sh[[ow]] ro[[ute]]
 cisco_asa_show_xlate.textfsm, .*, cisco_asa, sh[[ow]] x[[late]]
 cisco_asa_show_name.textfsm, .*, cisco_asa, sh[[ow]] nam[[e]]
-cisco_asa_show_arp.textfsm, .*, cisco_asa, sh[[ow]] arp
+cisco_asa_show_arp.textfsm, .*, cisco_(asa|ftd), sh[[ow]] arp
 cisco_asa_show_nat.textfsm, .*, cisco_asa, sh[[ow]] nat
 cisco_asa_dir.textfsm, .*, cisco_asa, dir
 
+cisco_ftd_ping.textfsm, .*, cisco_(asa|ftd), ping
```
FragmentedPacket, line 139: Sorry about this! I should have been clearer. Do you mind moving this up and then renaming to `cisco_asa_ping.textfsm`? I think we'll keep the naming convention the same, but introduce the `cisco_(asa|ftd)` into the index. The tests will have to be updated as well.

Let me know if you have any issues with this in our public Slack.

micahculpepper: OK, should be fixed now.

PR #676, opened 2020-05-03
Update arista_eos_show_interfaces.textfsm

Kosire:
Added support for subinterfaces like Ethernet49/1 for 100GE ports.

### DIFF
--- a/templates/arista_eos_show_interfaces.textfsm
+++ b/templates/arista_eos_show_interfaces.textfsm
@@ -10,7 +10,7 @@ Value MTU (\d+)
 Value BANDWIDTH (\d+\s+\w+)
 
 Start
-  ^${INTERFACE}\s+is\s+${LINK_STATUS}.*protocol\s+is\s+${PROTOCOL_STATUS} 
+  ^${INTERFACE}\s+|\d+is\s+${LINK_STATUS}.*protocol\s+is\s+${PROTOCOL_STATUS}
   ^\s+Hardware\s+is\s+${HARDWARE_TYPE}(.*address\s+is\s+${ADDRESS})*(.*bia\s+${BIA})*
   ^\s+Description:\s+${DESCRIPTION}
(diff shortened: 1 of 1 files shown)


templates/arista_eos_show_interfaces.textfsm
```diff
@@ -10,7 +10,7 @@ Value MTU (\d+)
 Value BANDWIDTH (\d+\s+\w+)
 
 Start
-  ^${INTERFACE}\s+is\s+${LINK_STATUS}.*protocol\s+is\s+${PROTOCOL_STATUS} 
+  ^${INTERFACE}\s+|\d+is\s+${LINK_STATUS}.*protocol\s+is\s+${PROTOCOL_STATUS}
```
FragmentedPacket, line 13: Please provide new `.raw` and `.yml` in the tests for this as well to make sure it's being properly captured and we know what the output looks like for this change as well.

Kosire: Yes, see https://github.com/networktocode/ntc-templates/pull/682

FragmentedPacket:
Do you mind bringing those into the original PRs and then closing out those PRs?

Each PR should include the following files that pertain to an issue/enhancement:
1. Changes to the template.
2. New `.raw` file that includes the raw output from the device and within the correct test directory.
3. New `.yml` file that includes the correctly parsed data using the changes from the template and within the correct test directory.

There is an included utility we provide that will parse the `.raw` file and provide the `.yml` (`development_scripts.py`) and more information regarding that can be found in our [README](https://github.com/networktocode/ntc-templates/blob/master/README.md). We recommend checking the auto-generated `.yml` file to make sure it does capture everything appropriately.

If you need help, you can find us on our Network to Code Slack.

FragmentedPacket:
@Kosire What issue are you having with this? Looking at the raw output you provided in #682, it parses fine except for link_status which is fixed in the PR below.

We can continue discussion on #711

PR #680, opened 2020-05-03
Update paloalto_panos_show_interface_hardware.raw

Kosire:
Updated the raw with an additional state for the speed column "ukn/ukn/down(autoneg)".


FragmentedPacket:
Closing and added to separate PR #710

PR #692, opened 2020-05-07
Cisco IOS 'show adjacency .* detail'

Yakuza-UA:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_ios_show_adjacency_detail.textfsm

##### SUMMARY
Introduces parsing logic to support 'show adjacency detail' command:
  - show adjacency [nexthop] [interface] detail
  - show adjacency [nexthop] detail
  - show adjacency [interface] detail
  - show adjacency detail

Includes test files from IOS-XE and legacy IOS platforms

Extracts
INTERFACE - interface name
ENDPOINT - nexthop's IP address, or name (e.g. point2point)
REWRITE_HEADERS - contains CEF-like headers in HEX format (can be used to extract VLAN ID, or destination IP address for GRE tunnels, etc)
RECURSIVE_INTERFACE - in case of recursive entry this shows the next interface
RECURSIVE_NEXTHOP - nexthop's IP address (for recursive)


templates/cisco_ios_show_adjacency_detail.textfsm
```diff
@@ -0,0 +1,36 @@
+Value Required INTERFACE (\S+)
+Value Required ENDPOINT (\S+)
+Value List REWRITE_HEADERS ([A-F0-9]+)
+Value RECURSIVE_INTERFACE (\S+)
+Value RECURSIVE_NEXTHOP (\S+)
+
+Start
+  ^Protocol\s+Interface\s+Address\s*$$ -> Entries
```
FragmentedPacket, line 8: Do you mind adding the following to each state? You have the error for two, but not the blank line.

```
  ^\s*$$
  ^. -> Error
```

Everything looks good besides those minor things.

Yakuza-UA: Thanks for your feedback!
Does it include 'Start' state as well?
How do I push new changes (sorry I am reasonably new to GitHub)? Do I just commit in my branch, then push to my forked copy and then do something within GitHub?
Regards

PR #693, opened 2020-05-07
Ciena sshkeystatus

georgesnow:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT

SSH server key status, ciena_saos, ssh server show key


templates/ciena_saos_ssh_server_show_key.textfsm
```diff
@@ -0,0 +1,13 @@
+Value USERNAME (\S*)
```
FragmentedPacket, line 1:
```suggestion
Value USERNAME (\S+)
Value KEY_STATUS (\S+)
```

templates/ciena_saos_ssh_server_show_key.textfsm
```diff
@@ -0,0 +1,13 @@
...
+Value KEY_STATUS (\S*)
+
+Start
+  ^\+\S*\s*\w+\s*\w+\s*\S*\s*-*\+
+  ^\|\s*\S*\s*\S*\s*\|\s*[Gg]enerated\s*\|
+  ^\|\s*\S*\s*[Ff]ingerprint\S*\s*\S*\s*\|\s*\S*\s*\|
```
FragmentedPacket, line 7: I think this template can be simplified a bit.

I would just account for the following:
```
Start
  # Accounts for tabular delineations and table names
  ^\+\-
  # Account for SSH Server Key block
  ^\|\s+[Kk]ey
  ^\|\s+[Uu]sername.*[Kk]ey\s+[Ss]tatus
  ^\|\s*${USERNAME}\s*\S*\s*\|\s*${KEY_STATUS}\s*\| -> Record
  ^\s*$$
  ^. -> Error
```

I haven't tested this, but I think it is more readable.

georgesnow: I thought you had to account for the every line as well as the whole line. however if this works. uh yea, it it way more readable. I will give it a go tomorrow. thanks!

templates/ciena_saos_ssh_server_show_key.textfsm
FragmentedPacket: Do you mind addressing this and I think we're good to get this merged in.

georgesnow: done.

PR #694, opened 2020-05-10
Cisco ios show ip cef

Yakuza-UA:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_ios_show_ip_cef.textfsm
Cisco IOS, IOS-XE: show ip cef, show ip cef detail

##### SUMMARY
This template understands two formats of show ip cef command:
   - short: show ip cef, show ip cef vrf 'name'
   - detailed: show ip cef detail, show ip cef vrf 'name' detail, show ip cef [prefix], etc

I have tested this template in live environment and executed it in on 500 devices, including Catalyst 3750, 3850, 9300 and 9500, ISR 2800, 2900, 3800, 3900, 4300, 4400, as well as VG204, VG224, etc

Couple issues have been identified and template adjusted. Please note, I am not a fan of 'capture all possible lines' approach, but because this is a requirement of NTC templates, I had to comply. This, however, means there's a risk this template might not be able to parse output from newer version of software. Generally speaking, fields that are IMPORTANT are identical across all IOS versions, except some very legacy IOS 12.x on ISR G1 (these are grouped as separate section in the template). However, after I successfully parsed output from 99% of my devices, I then faced an issue with the most recent IOS-XE software on 9500 (Gibraltar), where Cisco has added one line (QOS precedence), which is absolutely irrelevant for data collection, but broke parsing logic due to ^.* -> Error


itdependsnetworks:
Certainly understand where you are coming from, but do understand the long term benefits for the upfront efforts. Your scenario is actually very telling of the issue. A change was made to the output, now a human is made aware and can validate whether or not it relevant. In this case it was determined not to be, but if it were the opposite case you would have a %1 use case but that would be difficult to notice and could cause issues with components that rely on this data.

PR #695, opened 2020-05-10
New Template for Cisco IOS: show ip cef [detail]

Yakuza-UA:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
cisco_ios_show_ip_cef.textfsm
cisco_ios_show_ip_cef_detail.textfsm

Cisco IOS: show ip cef [vrf NAME]
Cisco IOS: show ip cef [vrf NAME] [PREFIX | INTERFACE] detail

##### SUMMARY
There are two templates, one understands brief list of CEF entries and can only parse `show ip cef`, or `show ip cef vrf NAME`

Other template can parse `show ip cef detail`, `show ip cef vrf NAME detail`, or `show ip cef INTERFACE detail`, etc as long as keyword `detail` is present. Otherwise there will be no match in index file.

I have tested this template in live environment and executed it in on 500 devices, including Catalyst 3750, 3850, 9300 and 9500, ISR 2800, 2900, 3800, 3900, 4300, 4400, as well as VG204, VG224, etc


Yakuza-UA:
Please do not merge yet. I have found one issue. Will let you know once I've committed all changes. Regards

Yakuza-UA:
Ok. Good to merge now. I've re-run all tests on ~500 devices.
It now supports ECMP, and it's good I've spotted it

itdependsnetworks:
I added a PR into your fork/branch. We do not generally treat multiple outputs as a single command. There is a fine line in when this applies, but a good rule of thumb would be if the output is modified by additional parameters, then a new template is required. If the output is filtered, we generally can reuse the same template.

templates/cisco_ios_show_ip_cef.textfsm
```diff
@@ -0,0 +1,43 @@
+#
+# This template is only compatible with the following command
+#   - show ip cef [vrf NAME]
+# This is enforced by INDEX file
+# Check cisco_ios_show_ip_cef_detail.textfsm for more details
+#
+Value Required PREFIX ((?:\d{1,3}\.){3}\d{1,3}\/\d{1,2})
+Value TYPE ((?:no\s)?(?:[a-z]+))
```
FragmentedPacket, line 8: We typically use the headers for the values as well. I don't think we should be capturing rows from the same column as different values (TYPE and NEXTHOP). They both are under the nexthop column.

templates/cisco_ios_show_ip_cef.textfsm
```diff
@@ -0,0 +1,43 @@
...
+
+Start
+  # >>> Parse EXCEPTIONS
+  # CEF is not enabled
+  ^%IPv4\s+CEF\s+not\s+running$$ -> End
+  # Invalid prefix
+  ^\s+Invalid\s+prefix/mask -> End
+  #
+  # >>> Parse HEADING
+  ^Prefix\s+Next\s+Hop\s+Interface$$ -> Entries
```
FragmentedPacket, line 24: Do you minding adding the ending lines to the `Start` state as well?

```
  ^\s*$$
  ^. -> Error
```

templates/cisco_ios_show_ip_cef_detail.textfsm
```diff
@@ -0,0 +1,126 @@
...
+#
+# This template is compatible with IOS v12.x and above,
+# including IOS-XE. There are two versions of output
+# with significant differences depending on IOS version.
+# Template is able to recognize both versions and captures
+# same structured data regardless of IOS version.
+#
+Value Required PREFIX ((?:\d{1,3}\.){3}\d{1,3}\/\d{1,2})
+Value List FLAGS ([^,\]\n]+)
+Value TYPE ((?:no\s)?(?:[a-z]+))
```
FragmentedPacket, line 15: This is the same thing in the above template as well.

Separating these out is more of a post-processing step and addresses your specific use case, but with maintaining a repository for most people, we need to solve the parsing problem and not the parsing/processing of specific use cases.

Other than those small adjustments, these looks good to merge once those are fixed and we really appreciate your effort and the level of detail you have taken!

templates/cisco_ios_show_ip_cef.textfsm
Yakuza-UA: I am don't think I understand this suggestion. I've read few times, but still blind.

templates/cisco_ios_show_ip_cef.textfsm
Yakuza-UA: I will do

templates/cisco_ios_show_ip_cef_detail.textfsm
Yakuza-UA: Again, I am not quite sure what is this suggestion about

templates/cisco_ios_show_ip_cef.textfsm
FragmentedPacket: The headers of the output are:
`Prefix               Next Hop             Interface`

I do not see type in there at all. You're saying there is either a "type" or "next hop" within the "Next Hop" column, but that is your assumption/use-case and not what the output is saying. You're deriving the type of not being an IP address and the next hop is an IP address.

This assumption/use-case should be handled in post-processing as that is an opinionated view of the data.

Your assumptions/use-cases are not wrong, but as stated above, it is opinionated and our repo attempts to stay as neutral as possible when it comes to parsing the data.

We should just stick to the headers and leave any use-cases to post processing once the data is structured.

Yakuza-UA: Ah, ok, I am with you now... so you're suggesting that there's no such thing as TYPE, but only NEXTHOP even though it is 'drop' or 'glean' or whatever. But in this case output from 'show ip cef' will not be compatible with 'show ip cef details' but if you want it this way I can do it

templates/cisco_ios_show_ip_cef_detail.textfsm
Yakuza-UA: I disagree as to this template. In detail view type is presented as SEPARATE value, it's not using same placement as NEXTHOP

templates/cisco_ios_show_ip_cef.textfsm
Yakuza-UA: This is resolved. Check last commit.

templates/cisco_ios_show_ip_cef_detail.textfsm
Yakuza-UA: Not going to resolve this, as it's different scenario, such as
```
  ^\s+${TYPE}(?<=receive)\s+for\s+${INTERFACE}
  ^\s+${TYPE}(?<=attached)\s+to\s+${INTERFACE}
  ^\s+${TYPE}(?<=drop)$$
  ^\s+${TYPE}(?<=no route)$$
  ^\s+${TYPE}(?<=receive)$$
  ^\s+${TYPE}(?<=multicast)$$
  ^\s+${TYPE}(?<=nexthop)\s+${NEXTHOP}\s+${INTERFACE}
```

Yakuza-UA:
Ok. I had to review my template from scratch to make sure it complies with all requirements
As result, I have simplified it and no longer capture any extra details, such as flags (recursive, connected, default route, etc). However, data structure is IDENTICAL to another template (show ip cef), even nexthop values are similar

PR #699, opened 2020-05-15
New Template for Cisco IOS: traceroute <destination> [options]

Yakuza-UA:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Cisco IOS, traceroute ip_address + keywords: cisco_ios_show_traceroute.textfsm

##### SUMMARY
This template can parse 'numeric' and 'non-numeric'
versions of the output, including extra details, such
as MPLS Exp bits, ASNs and MPLS labels (if available)

Supports up to 10 probes per hop
Captures RTT, or fault reason per probe/per hop

Excessive probes per hop (11th and above) are ignored


templates/cisco_ios_traceroute.textfsm
```diff
@@ -0,0 +1,44 @@
...
+#   - MPLS Exp Bits / Labels
+# This is captured into details field
+Value DETAILS (.*)
+
+Start
+  ^Type\s+escape
+  ^Tracing\s+the\s+route
+  ^VRF\s+info: -> Entries
+  ^\s*$$
+  ^.*$$ -> Error
```
FragmentedPacket, line 26:
```suggestion
  ^\s*$$
  ^. -> Error
```

templates/cisco_ios_traceroute.textfsm
```diff
@@ -0,0 +1,44 @@
+Value Required,Filldown HOP_NUM (\d+)
+Value ADDRESS ((?:\d{1,3}\.){3}\d{1,3})
+Value FQDN (\S+)
+# Captures RTT (in milliseconds), or faults, like
```
FragmentedPacket, line 4: Would it be worth adding a DESTINATION value to this? So then it is clear as to where the traceroute is attempting to go?

Yakuza-UA: Why? This comes from the command, i.e. if someone traceroutes to 8.8.8.8, they know the destination well in advance. It's pointless value IMHO as it is known in advance.

FragmentedPacket: My take is that the destination is also stored in the data structure and can be used in further processing.

It's not a big ask, but I'm not 100% tied to it so if you want to leave it out, then that is fine.

PR #640, opened 2020-04-05
Ciena show vlan

friday963:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

ciena_6x_vlan_show, ciena_6x, vlan_show

##### SUMMARY

Adding vlan show command template.

```

```

PR #707, opened 2020-05-19
BugFix: Ciena Saos: Added more use cases to vlan_show

FragmentedPacket:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

ciena_saos_vlan_show.textfsm
##### SUMMARY
Adds in more use cases from #640 

@georgesnow @friday963

The name does change in these, is that OK?


georgesnow:
Is this change you are referring to:

`Value VLAN_NAME ([0-9a-zA-Z_#]+)`

` Value NAME ([0-9a-zA-Z_#]+)`

FragmentedPacket:
Correct. That is a difference between both of y'all templates and is disruptive for you

georgesnow:
shouldn't effect me. I am not actually using the template for anything. I only added as baseline to basic output of the command.

FragmentedPacket:
@georgesnow Thanks for the quick response!

PR #724, opened 2020-05-23
Update to enforce double-quote

itdependsnetworks:
##### ISSUE TYPE

 - Additional Testing

##### COMPONENT

Yamllint
##### SUMMARY

This enforces double-quote. To me, we either have it, or remove it, otherwise the warnings at the start of tox testings is getting long


FragmentedPacket:
I completely agree. Thanks for making the issue.

PR #734, opened 2020-05-29
Fixes missing interfaces on down interfaces

### LINKED ISSUE #627: Cisco ASA Show interface does not catch unnamed interfaces
##### ISSUE TYPE

 - Bug Report

##### TEMPLATE USING

```
cisco_asa_show_interface
```

##### SUMMARY

When executing the show interface, even though the output has several interfaces, only those with a nameif assigned to them are being returned.

I have the start of a fix, but will also spend time to get the full cleanup done on the template to match what we have in place for newer templates.

jvanderaa:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_asa

##### SUMMARY

Fixes for interfaces that are in a down status

Fixes #627 

```

```


templates/cisco_asa_show_interface.textfsm
```diff
@@ -25,12 +25,14 @@ Value FIVEMIN_DROP_RATE (\d+)
 
 Start
   ^Interface.* -> Continue.Record
-  ^.*Interface ${INTERFACE} "${INTERFACE_ZONE}", is ${LINK_STATUS}.*protocol is ${PROTOCOL_STATUS}
```
FragmentedPacket, line 28: Can you remove all the static whitespaces in here and replace with `\s+` since we're expecting at least one space and it'll catch if they add more for whatever reason in a newer release.

templates/cisco_asa_show_interface.textfsm
```diff
@@ -25,12 +25,14 @@ Value FIVEMIN_DROP_RATE (\d+)
...
+  ^.*Interface ${INTERFACE} "${INTERFACE_ZONE}", is ${LINK_STATUS},.*protocol\s+is\s+${PROTOCOL_STATUS}
+  ^.*Interface ${INTERFACE}.*is\s+${LINK_STATUS},.*protocol\s+is\s+${PROTOCOL_STATUS}
```
FragmentedPacket, line 29: I would probably get rid of as many of the `.*` as possible in this line such as this:

```suggestion
  ^.*Interface ${INTERFACE}.*is\s+${LINK_STATUS},\s+line\s+protocol\s+is\s+${PROTOCOL_STATUS}
```

jvanderaa: Makes sense. I wanted to get something out there. Hopefully I can cut it back. We only have one other example thus far. Hopefully it doesn't break anything.

templates/cisco_asa_show_interface.textfsm
jvanderaa: Didn't want to adjust the original template that was out there. WIll get the updates in.

PR #744, opened 2020-06-04
Added interfaces to arista_eos_show_vrf template

JoeyG1973:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

artists eos show vrf

##### SUMMARY

Added interfaces assigned to the VRF.  Also, I corrected the table spacing in the existing raw file.

There are 2 command outputs here.  In my case, I only had 2 interfaces per line, but in your raw data, you folks showed 3 interfaces per line.  I coded to both scenarios.  Please see the 2 raw files under test/arista_eos_show_vrf


templates/arista_eos_show_vrf.textfsm
```diff
@@ -1,20 +1,98 @@
 Value VRF (\S+)
 Value RD (\d\S+|<.+>)
+# Value List PROTOCOLS (\S+)
```
FragmentedPacket, line 3: Any reason these are commented out?

tests/arista_eos/show_vrf/arista_eos_show_vrf.raw
```diff
@@ -1,14 +1,14 @@
...
-                                                                             Vlan2238, Vlan2239, Vlan2240
-   green            <not set>              ipv4,ipv6       v4:routing,       Vlan1015, Vlan1016, Vlan1017,
-                                                           v6:no routing     Vlan1018, Vlan1019, Vlan1020,
-                                                                             Vlan1021
-   yellow           10.125.253.15:4        ipv4,ipv6       v4:routing,       Vlan1009                      
-                                                           v6:no routing                                
-   red              10.125.253.15:6        ipv4,ipv6       v4:routing,                               
-                                                           v6:no routing                                  
-   black            999:999                ipv4,ipv6       v4:no routing     Management1
+  Vrf          RD                   Protocols      State             Interfaces
```
FragmentedPacket, line 1: Do you mind reverting this as well? I'd rather just keep the raw data we already have since I'm not sure if this is something that has changed between versions or not.

templates/arista_eos_show_vrf.textfsm
JoeyG1973: I was originally going to include them but since my customer doesn't need them I am not doing that now, I will remove them

tests/arista_eos/show_vrf/arista_eos_show_vrf.raw
JoeyG1973: Spacing is critical, without proper spacing in the table, none of this works.  The original raw data had spacing all over the place and was missing the critical \s$ at the end of each line for some of the lines.  The original template was also not concerned with collecting interfaces.  It appears as though the original raw file was input incorrectly as every version of eos we have tested does not care about the terminal width it always puts out a table exactly the same width with two interfaces max per line with proper spacing.  I only accounted for three interfaces per line in my update because that is what is in the existing raw file

tox.ini
```diff
@@ -9,7 +9,4 @@ whitelist_externals =
     git
     rm
 
-commands =
-    black ./ --diff --check
-    yamllint tests/
```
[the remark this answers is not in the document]
JoeyG1973: will do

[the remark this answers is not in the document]
JoeyG1973: My local copy doesn't show that at all
```
[tox]
envlist = py36,py37,py38
skip_missing_interpreters = true

[testenv]
install_command = pip install .[dev] {opts} {packages}

whitelist_externals =
    git
    rm

commands = pytest -vv
```

tests/arista_eos/show_vrf/arista_eos_show_vrf.raw
FragmentedPacket: Ok that's fair and hopefully doesn't become an issue in the future or we'll have to address those differences in spacing.

tox.ini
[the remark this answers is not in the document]
FragmentedPacket: I'm not sure why, but you can do the following:

```
git add upstream git@github.com:networktocode/ntc-templates.git
git checkout <this-pr-branch-name>
git checkout upstream master tox.ini
```

This should grab the file from upstream's master branch. If you look at our repo, it shows the correct stuff.

tests/arista_eos/show_vrf/arista_eos_show_vrf.raw
JoeyG1973: Murphy hates me so it will probably be me.

JoeyG1973:
changes completed

PR #757, opened 2020-06-10
Arista eos show port channel summary

JoeyG1973:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

arista_eos_show_port-channel_summary
##### SUMMARY

added parsing for artists eos show port-channel summary

```
see the raw file at tests/arista_eos/show_port-channel_summary/arista_eos_show_port-channel_summary.raw
```


templates/arista_eos_show_port-channel_summary.textfsm
```diff
@@ -0,0 +1,29 @@
+Value PORT_CHANNEL (Po\d+)
+Value PORT_CHANNEL_STATUS (\(\w+\))
+Value PROTOCOL (\w+)
+Value PROTOCOL_STATE ((\(\w+\*\)|\(\w+\)))
+Value List PHYS_IFACE (Et.+?)
+Value List PHYS_IFACE_STATUS (\(.+?\))
+
+
+Start
```
FragmentedPacket, line 9:
```suggestion
Value PORT_CHANNEL (Po\d+)
Value PORT_CHANNEL_STATUS (\S+)
Value PROTOCOL (\w+)
Value PROTOCOL_STATE (\S+)
Value List PHYS_IFACE (Et\S+)
Value List PHYS_IFACE_STATUS (\S+)
```
I would say to make the values like this and account for the parenthesis in the states instead of capturing them as well. You just capture the value in-between them.

```suggestion
Start
  ^\s*Flags
  ^-+
  ^\s+\S\s+-
  ^Number
  ^\s+Port-Channel\s+Protocol\s+Ports\s+$$ -> PCS
  ^\s*$$
  ^. -> Error
```

Just to be more verbose in what we're attempting to account for in the data.

Here are two other suggestions:
`  ^\s*Po\d+.+(Et.+?\(.+?\)\s){1}${PHYS_IFACE}\(${PHYS_IFACE_STATUS}\)(\s|$$) -> Continue`
instead of:
`  ^\s*Po\d+\(\w+\)\s+\w+(\(\w+\*\)|\(\w+\))\s+(Et.+?\(.+?\)\s){1}${PHYS_IFACE}${PHYS_IFACE_STATUS}(\s|$$) -> Continue` for each of those lines.

`  ^\s+(Et.+?\(.+?\)\s){1}${PHYS_IFACE}${PHYS_IFACE_STATUS}(\s|$$) -> Continue`
instead of:
`  ^\s+\s+(Et.+?\(.+?\)\s){1}${PHYS_IFACE}${PHYS_IFACE_STATUS}(\s|$$) -> Continue` for each of those lines.

JoeyG1973: I copied and modified the cisco_nxos_show_port-channel_summary so that captures the parenthesis as well.  I will change it in mine but please note that this is the way it is is in Cisco NXOS so you guys should probably change that as well.

abahtiti:
I think this line is not correct
`^\s+Port-Channel\s+Protocol\s+Ports\s+$$ -> PCS
`
it should be 

`^\s+Port-Channel\s+Protocol\s+Ports\s*$$ -> PCS`

PR #758, opened 2020-06-11
New Template: cisco_nxos_show_interface_description

network-dave:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

Cisco NXOS - show interface description

##### SUMMARY

N/A


templates/cisco_nxos_show_interface_description.textfsm
```diff
@@ -0,0 +1,16 @@
+Value PORT (\S+)
```
FragmentedPacket, line 1:
```suggestion
Value PORT (\S+)
Value TYPE (\S+)
Value SPEED (\w+)
Value DESCRIPTION (.+)

Start
  ^-+
  ^Interface\s+Description -> Type1
  ^Port\s+Type\s+Speed\s+Description -> Type2
  ^\s*$$
  ^. -> Error

Type1
  ^-+
  ^Port\s+Type\s+Speed\s+Description -> Type2
  ^${PORT}\s+${DESCRIPTION}$$ -> Record
  ^\s*$$
  ^. -> Error

Type2
  ^-+
  ^Interface\s+Description -> Type1
  ^${PORT}\s+${TYPE}\s+${SPEED}\s+${DESCRIPTION} -> Record
  ^\s*$$
  ^. -> Error
```

Here is a template that should work and uses the state changes to capture the necessary output for each header in the output.

network-dave: This doesn't work at 100% because it catches "Interface          Description" while already in the Type1 state. The new version would be substantially more complicated for the same result. I will commit a very slightly updated version. 

Open to discuss this though.

PR #760, opened 2020-06-11
New Template: cisco_nxos_show_hsrp_all

network-dave:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

Cisco NXOS - show hsrp all

##### SUMMARY

Note: some items will stay empty depending on which is the active router.


templates/cisco_nxos_show_hsrp_all.textfsm
```diff
@@ -0,0 +1,55 @@
...
+  ^${INTERFACE}\s+-\s+Group\s${GROUP_NUMBER}\s+\(HSRP-V${VERSION}\)\s\(IPv[46]\)\s*$$
+  ^\s+Local\s+state\s+is\s+${HSRP_ROUTER_STATE},\s+priority\s+${PRIORITY}\s+\(Cfged\s+${CONFIGURED_PRIORITY}\),\s+${PREEMPT}\s*$$  
+  ^\s+.*lower:\s+${LOWER_FWD_TRESHOLD}\s+upper:\s+${UPPER_FWD_TRESHOLD}\s*$$
+  ^\s+Preemption\s+Delay.*$$
+  ^\s+Hellotime\s+${TIMERS_HELLO_SEC}.*holdtime\s+${TIMERS_HOLD_SEC}.*$$
+  ^\s+Next\s+hello\s+sent.*$$
+  ^\s+Virtual\s+IP\s+address\s+is\s+${PRIMARY_IPV4_ADDRESS}\s+.*$$
+  ^\s+Secondary Virtual IP address is ${SECONDARY_IPV4_ADDRESS}\s*$$
+  ^\s+Active router is ${ACTIVE_ROUTER}\s*$$
+  ^\s+Active router is ${ACTIVE_IP_ADDRESS}\s?, priority ${ACTIVE_PRIORITY} expires in ${ACTIVE_EXPIRE}.*$$
```
FragmentedPacket, line 43: Do you mind replacing the static spaces with `\s+` all through the template?

Also, I would get rid of the record on whitespace and add the follow right below `Start`:
```suggestion
Start
  ^Vlan -> Continue.Record
```

Also:
```suggestion
  # Catch unexpected non-empty lines (secondary IP's at the end of the block for example)
  ^\s+\d+\.\d+\.\d+\.\d+$$
```

network-dave: This went under the radar, my bad. Will do ASAP !

network-dave: ^Vlan -> Continue.Record

This won't work because the interface name can be anything, not only "Vlan". We could eventually "Continue.Record" on the same REGEX as the first line, but isn't it more convenient to Record at the end ?

FragmentedPacket: I don't see these changes either :(

network-dave: The "^Vlan -> Continue.Record" does not work as intended, what do you suggest ? Let me know if you want to discuss this.

FragmentedPacket: You can just leave it as is right now, but push the white space changes up.

network-dave: Just pushed the changes to the branch, my bad.

PR #789, opened 2020-08-12
Detect hostname on NX-OS platforms

Yakuza-UA:
##### ISSUE TYPE
 - Template enhancements

##### COMPONENT
Cisco NX-OS: show version (cisco_nxos_show_version.textfsm)

##### SUMMARY
Added detection of hostname for show version on NX-OS platform, similar to what's already available for Cisco IOS


Yakuza-UA:
Let me know if you're ok to leave it as _hostname_ and not _device_name_ as seen in the CLI output. Mainly to align with Cisco IOS version.

FragmentedPacket (approved):

PR #798, opened 2020-09-08
Bugfix: cisco_ios_show_running-config_partition_access-list

jpobeda:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_ios_show_running_partition_access-list.textfsm
show running partition access-list

##### SUMMARY

Cisco IOS-XE outputs ACE line number on later versions. 
Added Value LINE_NUM and made it backwards compatible.
Introduced Value ICMP_TYPE to amend capturing the icmp type eg: 'echo=reply'


jpobeda:
I'll close and submit again, I screw it all and will be easier to just re do it than trying to fix the log and all.

PR #800, opened 2020-09-08
Bugfix: cisco_ios_show_ip_access_list

jpobeda:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_ios_show_ip_access-lists.textfsm

##### SUMMARY

- It wasn't capturing ICMP TYPES.
- Added a new Value for ICMP_TYPE and adjusted the record statement.
- Updated tests/cisco_ios/show_ip_access-lists/cisco_ios_show_ip_access-lists_fixed.raw with a few test cases as below.

**Note**: if you go by number rather than named icmp types, it seems to allow for a maximum of 2 on the same ace entry.
Eg: `permit icmp host 10.10.10.10 any 11 13`
I'm not capturing this atm, but was thinking _-who in the hell would go that way-_ to make ICMP_TYPE a list rather than a value if we want to have it. Happy to leave that to the next :D

```
Extended IP access list ntc-templates
    10 permit icmp host 10.10.10.10 any echo-reply
    20 permit icmp host 10.10.10.10 any administratively-prohibited
    30 permit icmp host 10.10.10.10 any unreachable log
    40 permit icmp host 10.10.10.10 any redirect
    50 permit icmp 10.10.10.0 0.0.0.255 any ttl-exceeded
    60 permit icmp 10.10.10.0 0.0.0.255 20.20.0.0 0.0.255.255 port-unreachable
    70 permit icmp 10.10.10.0 0.0.0.255 20.20.0.0 0.0.255.255 parameter-problem log
```


jpobeda:
Kind of goes together with https://github.com/networktocode/ntc-templates/pull/798

jmcgill298:
@jpobeda I believe you need to rebase these PRs to get the tests to pass

jpobeda:
I'll close and submit again, I screw it all and will be easier to just re do it than trying to fix the log and all.

PR #813, opened 2020-10-08
cisco_nxos show version - extract serial from "Processor board ID"

diepes:
##### ISSUE TYPE

 - Bugfix Pull Request
 - Additional Testing

##### COMPONENT

templates/cisco_nxos_show_version.textfsm

##### SUMMARY

1. extract serial# from show version (Also for existing 3 raw files, and added 4th with different syntax)
2. end on -> Error  
3. match additional stack switch serial, and add to list.

```

```


templates/cisco_nxos_show_version.textfsm
```diff
@@ -1,18 +1,71 @@
...
   ^\s+cisco\s+${PLATFORM}\s+[cC]hassis
   ^\s+cisco\s+Nexus\d+\s+${PLATFORM}
   # Cisco N5K platform
   ^\s+cisco\s+Nexus\s+${PLATFORM}\s+[cC]hassis
   ^\s+Device\s+name:\s+${HOSTNAME}$$
   ^\s+cisco\s+.+-${PLATFORM}\s*
-  ^Kernel\s+uptime\s+is\s+${UPTIME}
-  ^\s+Reason:\s${LAST_REBOOT_REASON} -> Record
+  # Nexus intel platform uses Board ID as serial/license
+  ^\s*Processor\s[Bb]oard\sID\s+${SERIAL}$$
```
FragmentedPacket, line 22: It looks like this is attempting to be used as Cat9K, but these are managed via different templates.

Please remove any IOS specific things as well as the new test file that is IOS-XE related.

I would suggest just capturing the serial number and not adding the `. ^. -> Error` since there are plenty of lines we don't care about and may make maintaining this template harder than other templates where most of the data needs to be captured.

FragmentedPacket (changes requested):

templates/cisco_nxos_show_version.textfsm
diepes: Thanks for the review and feedback, trying to implement/simplify/fix it now.

templates/cisco_nxos_show_cdp_neighbors_detail.textfsm
```diff
@@ -6,14 +6,14 @@ Value REMOTE_PORT (.*)
...
-Value CAPABILITIES (.*)
+Value CAPABILITIES (.*[^\s])
 
 Start
   ^Device ID:${DEST_HOST}
   ^System Name: ${SYSNAME}
   ^Interface address\(es\): -> GetInterfaceIP
   ^Mgmt address\(es\): -> GetIP
-  ^Platform: ${PLATFORM}, Capabilities: ${CAPABILITIES}
+  ^Platform: ${PLATFORM}, Capabilities: ${CAPABILITIES}\s?$$
```
jmcgill298, line 16:
```suggestion
  ^Platform: ${PLATFORM}, Capabilities: ${CAPABILITIES}\s*$$
```

This should allow 0 or more spaces

templates/cisco_ios_show_ip_eigrp_neighbors.textfsm
```diff
@@ -10,7 +10,8 @@ Value SEQ_NUM (\d+)
 
 Start
   ^.*\s+${ADDRESS}\s+${INTERFACE}\s+${HOLD}\s+${UPTIME}\s+${SRTT}\s+${RTO}\s+${Q_CNT}\s+${SEQ_NUM}\s*$$ -> Record
-  ^\s*(?:IP-|)EIGRP(?:-IPv(?:4|6)|)\s+[Nn]eighbors\s+for\s+(?:process\s+|AS\()${AS}(?:\)|)\s*$$
+  ^\s*(IP-|)EIGRP(-IPv[46](:\(\d+\))?|)\s+[Nn]eighbors\s+for\s+(process\s+|AS\()${AS}(\)|)\s*$$
+  ^\s*(IP-|)EIGRP(-IPv[46](:\(\d+\))?|)\s+[Nn]eighbors\s+for\s+(process\s+|AS\()${AS}(\)|)\sVRF default*$$
```
jmcgill298, line 14:
```suggestion
  ^\s*(IP-|)EIGRP(-IPv[46](:\(\d+\))?|)\s+[Nn]eighbors\s+for\s+(process\s+|AS\()${AS}(\)|)\s+VRF default\s*$$
```

Update capturing spaces

jmcgill298 (changes requested):
just a couple of issues on matching space characters. Thanks!

jmcgill298 (approved):

PR #815, opened 2020-10-25
Bugfix: ASA show interface - missing case handled for no values for duplex & speed

starlightdreamer:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
Cisco ASA show interface

##### SUMMARY
The line expected to have the format `...Duplex([parsed duplex value]), Auto-Speed([parsed speed value])`
sometimes does not have any parenthesis after `Duplex` and `Speed`.  When encountering this output, the current template fails.  This change accounts for the possibility and assigns `""` as the values.

Example of the expected duplex/speed line format:
```
	Auto-Duplex(Full-duplex), Auto-Speed(1000 Mbps)
```

Example of the unaccounted format:
```
        Full-Duplex, Auto-Speed
```


jmcgill298:
Thanks @starlightdreamer

PR #818, opened 2020-10-27
Bugfix: added 'Autostate exclude' to avoid error

abaretta:
IOS (15?) includes an optional 'Autostate exclude' switchport configuration option. Adding this prevents the filter throwing an error when the option is configured on an interface.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

Cisco IOS 'show interfaces switchport'

##### SUMMARY

The filter does not take into account the possibility of the 'Autostate exclude' option being configured on the switch port.

```
Before the filter throws an error when 'Autostate exclude' is configured on the interface, after it does not (otherwise no difference in output).
```


jmcgill298:
Thanks @abaretta

PR #843, opened 2020-12-09
Fixes#811 arista eos show ip route

diepes:
##### ISSUE TYPE

 - Bugfix Pull Request
 - Additional Testing

##### COMPONENT

arista_eos_show_ip_route

##### SUMMARY

1. handle tunnel interface in nexthop
2. Change nexthop IP/Int into a list.

before
```
    next_hop: "connected"
    interface: "Vlan105"
```
after
```
    next_hop:
      - "connected"
    interface:
      - "Vlan105"
```


jmcgill298 (approved):
Thanks @diepes I personally like this approach better, but it will be more difficult for users to consume. What do you think @FragmentedPacket @jvanderaa ?

FragmentedPacket:
I like this approach as well. I believe we've converted a few templates to this as well for ECMP related routes.

FragmentedPacket (approved):

jvanderaa:
It's not clear to me why the change to a list of items instead of just the value of the key being the text. I'm assuming that there is the possibility of multiple routes then? I'm not going to hold this up, but we should get an example of this into the tests as well to verify that it is handled.

Also, would this cause us to go to 2.x for the templates with the change?

jmcgill298:
@jvanderaa It is for multiple paths in the routing table, and the existing test cases have been updated and include scenarios for this scenario.

PR #848, opened 2020-12-24
cisco_nxos_show_ip_bgp_summary_vrf update to support 32-bits ASN outputs

burningnode:
##### ISSUE TYPE
 - Additional Testing & fixes

##### COMPONENT
cisco_nxos_show_ip_bgp_summary_vrf.textfsm, cisco_nxos, show_ip_bgp_summary_vrf

##### SUMMARY
I'd like to update this template because it fails to parse the output with 32-bit ASNs and line breaks/spaces relative to this output. I also added a separate entry for this template/command in the index.


burningnode:
Need to rework this

PR #863, opened 2021-01-28
Add Environment Variable Option for Custom Template Location

jeffkala:
##### ISSUE TYPE
 - Enhancements

##### COMPONENT
core update to `parse` to add ability to set custom template location via environment variable.

##### SUMMARY
Allow for custom templates directory to be specified and used.  This is helpful when you need to develop custom templates and usages within a project directory.

If the environment variable is not defined it reverts to existing logic to load from installed path.  I have tested the change.

### DIFF
--- a/lib/ntc_templates/parse.py
+++ b/lib/ntc_templates/parse.py
@@ -15,7 +15,7 @@ def _get_template_dir():
     env_dir = os.environ.get("NTC_TEMPLATES_DIR")
     if env_dir is not None:
         template_dir = os.path.join(env_dir, "templates")
-    if template_dir is not None:
+    if template_dir is None:
         package_dir = os.path.dirname(__file__)
         template_dir = os.path.join(package_dir, "templates")
         if not os.path.isdir(template_dir):
--- a/lib/ntc_templates/parse.py
+++ b/lib/ntc_templates/parse.py
@@ -12,10 +12,9 @@ except ImportError:
 
 
 def _get_template_dir():
-    env_dir = os.environ.get("NTC_TEMPLATES_DIR")
-    if env_dir is not None:
-        template_dir = os.path.join(env_dir, "templates")
-    if template_dir is None:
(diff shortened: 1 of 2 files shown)


lib/ntc_templates/parse.py
```diff
@@ -12,11 +12,13 @@
 
 
 def _get_template_dir():
-    package_dir = os.path.dirname(__file__)
-    template_dir = os.path.join(package_dir, "templates")
-    if not os.path.isdir(template_dir):
-        project_dir = os.path.dirname(os.path.dirname(os.path.dirname(template_dir)))
-        template_dir = os.path.join(project_dir, "templates")
+    template_dir = os.environ.get("NTC_TEMPLATES_DIR")
```
FragmentedPacket, line 15: I'm wondering if we should offer both an OS environment or pass in as an arg to `parse_output`.

Something like.

```
def parse_output(platform=None, command=None, data=None, template_dir=None):
    template_dir = os.environ.get("NTC_TEMPLATES_DIR") or template_dir
    if template_dir is None:
        template_dir = _get_template_dir()
```

And then keep `_get_template_dir()` the same.

jmcgill298: I would say it is good to have `_get_template_dir` do the environment lookup, but allow this function to override the environment by passing in the directory.

```
def parse_output(platform=None, command=None, data=None, template_dir=None):
    template_dir = template_dir or _get_template_dir()
```

I'm also okay with the PR like it is since that is more than is available today and a reasonable solution

jeffkala: @jmcgill298 @FragmentedPacket , just let me know either way.  I don't mind taking some extra time to implement the ability to set an arg.

FragmentedPacket: I'm with @jmcgill298 and his way is better than mine.

I'll approve it and let you decide :)

FragmentedPacket (approved):

jmcgill298 (approved):

PR #865, opened 2021-01-31
New template: yamaha

akira6592:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

- yamaha_show_environment
- yamaha_show_ip_route

##### SUMMARY

Add some templates for yamaha. yamaha is also supported by netmiko.


templates/yamaha_show_ip_route.textfsm
```diff
@@ -0,0 +1,16 @@
+# You need to set "console character ascci".
```
FragmentedPacket, line 1: Do you mind using looking at `cisco_ios_show_ip_route` and use some of the same names for those? We try and keep values the same between vendors to hopefully allow users to use the same keys for each vendor/command.

Other than that, I think this all looks good.

FragmentedPacket (changes requested):

templates/yamaha_show_ip_route.textfsm
akira6592: Thank you for your comment. I fixed it.
`additional_info` can contain various information, so I keep the same key name as the result of executing the show command.

templates/yamaha_show_ip_route.textfsm
```diff
@@ -0,0 +1,17 @@
...
+Value NETWORK (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|default)
+Value MASK (\d{1,2})
+Value NEXTHOP_IP (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|-)
+Value NEXTHOP_IF (\S+)
+Value PROTOCOL (\S+)
+Value ADDITIONAL_INFO (\S*)
+
+Start
```
FragmentedPacket, line 9:
```suggestion
Start
  ^Destination\s+Gateway\s+Interface\s+Kind\s+Additional\s+Info -> Routes
```

templates/yamaha_show_ip_route.textfsm
```diff
@@ -0,0 +1,18 @@
...
+  ^Destination\s+Gateway\s+Interface\s+Kind\s+Additional\s+Info -> Routes
+  ^Destination.* -> Routes
```
FragmentedPacket, line 11: Sorry I was hoping it would delete L11. Do you mind deleting L11 and then we should be good to merge this in.

akira6592: OK. Thank you.

PR #867, opened 2021-02-25
add test security-policy-match to palo alto

jeffkala:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

paloalto_panos_test_security-policy-match.textfsm, paloalto_panos, test security-policy-match <params>.

##### SUMMARY

Gives the ability to run the `test security-policy-match <params>` command and get structured data back.

```

```


templates/paloalto_panos_test_security-policy-match.textfsm
```diff
@@ -0,0 +1,36 @@
...
+Value DESTINATION_DEVICE (\S+)
+Value CATEGORY (\S+)
+Value APPLICATION_SERVICE (\[.*\])
+Value APPLICATION_SERVICE_IMPLICIT (\[.*\])
+Value ACTION (\S+)
+Value ICMP_UNREACHABLE (\S+)
+Value TERMINAL (\S+)
+
+Start
+  ^\"${NAME};\s+index:\s+${INDEX}\"\s+\{
```
FragmentedPacket, line 20: I think just grabbing the index should suffice rather than escaping the characters after, but you can leave it how it is as well. Just a nit pick on my part.

templates/paloalto_panos_test_security-policy-match.textfsm
```diff
@@ -0,0 +1,36 @@
...
+  ^\s+source-region\s+${SOURCE_REGION}; 
+  ^\s+to\s+${TO};
+  ^\s+destination\s+${DESTINATION};
+  ^\s+destination-region\s+${DESTINATION_REGION};
+  ^\s+user\s+${USER};
+  ^\s+source-device\s+${SOURCE_DEVICE};
+  ^\s+destinataion-device\s+${DESTINATION_DEVICE};
+  ^\s+category\s+${CATEGORY};
+  ^\s+application/service\s+${APPLICATION_SERVICE}];
+  ^\s+application/service\(implicit\)\s+${APPLICATION_SERVICE_IMPLICIT}];
```
FragmentedPacket, line 32: Do you have an example of the `application/service(implicit)` for the raw?

templates/paloalto_panos_test_security-policy-match.textfsm
```diff
@@ -0,0 +1,36 @@
...
+Value SOURCE (\S+)
+Value SOURCE_REGION (\S+)
+Value TO (\S+)
+Value DESTINATION ([\S+\s+]+)
+Value DESTINATION_REGION (\S+)
+Value USER (\S+)
+Value SOURCE_DEVICE (\S+)
+Value DESTINATION_DEVICE (\S+)
+Value CATEGORY (\S+)
+Value APPLICATION_SERVICE (\[.*\])
```
FragmentedPacket, line 13: I don't think APPLICATION_SERVICE is capturing as you expect from the `.raw` and `yml` so may need to revisit this Regex.

FragmentedPacket (changes requested):

FragmentedPacket:
Hit me up on Slack if you have any questions.

templates/paloalto_panos_test_security-policy-match.textfsm
jeffkala: Spot on on this one, I've fixed it.  I was using the regex from the running config that didn't work in this case.  Also the same one your previous IMPLICIT comment, that doesn't exist in the test command output.  Thanks for checking my work.

jeffkala:
> Hit me up on Slack if you have any questions.

Thanks, all great feedback.  I'm going back through each line now.  I went back into a test Palo and realized there is a few other 'or' options I need to add for some of these regex's.

templates/paloalto_panos_test_security-policy-match.textfsm
```diff
@@ -0,0 +1,36 @@
...
+  ^\s+user\s+${USER};
+  ^\s+source-device\s+${SOURCE_DEVICE};
+  ^\s+destinataion-device\s+${DESTINATION_DEVICE};
+  ^\s+category\s+${CATEGORY};
+  ^\s+application/service\s+${APPLICATION_SERVICE}];
+  ^\s+application/service\(implicit\)\s+${APPLICATION_SERVICE_IMPLICIT}];
+  ^\s+action\s+${ACTION};
+  ^\s+icmp-unreachable:\s+${ICMP_UNREACHABLE}
+  ^\s+terminal\s+${TERMINAL};
+  ^} -> Record
```
FragmentedPacket, line 36: Sorry I forgot about this until you said something, but we require the following lines:

```
  ^\s+$$
  ^. -> Error
```

You mentioned more use cases to account for, but you don't need to capture everything. We don't want to put that burden on you, and basically implement what works for you at this point, but you'll have to at least add part of the line to account for it.

jeffkala:
Should be good now, fixed all your concerns, along with fixing multiple of the regexs that wouldn't work in every scenario.

FragmentedPacket (approved):

Issue #869, opened 2021-02-27
cisco_nxos_show_ip_route errors out on Route Not Found


seanblundy:
##### ISSUE TYPE

 - Bug Report

##### TEMPLATE USING

```
Value Filldown VRF (\S+)
Value PROTOCOL ([\w-]+)
Value TYPE ([\w-]+)
Value Filldown NETWORK (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
Value Filldown MASK (\d{1,2})
Value DISTANCE (\d+)
Value METRIC (\d+)
Value NEXTHOP_IP (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
Value NEXTHOP_IF ([\w\./]+)
Value UPTIME ([\w:\.]+)
Value NEXTHOP_VRF (\S+)
Value TAG (\d+( \(\w+\))?)
Value SEGID (\d+)
Value TUNNELID (0x[a-f\d]+)
Value ENCAP (\w+)

Start
  ^IP\s+Route\s+Table\s+for\s+VRF\s+"${VRF}"\s*$$
  # Match the Network/Prefix Line
  ^\s*${NETWORK}\/${MASK}, ubest/mbest:
  #
  # Match VXLAN Route Entry
  ^\s+\*+via ${NEXTHOP_IP}(%${NEXTHOP_VRF})?, \[${DISTANCE}/${METRIC}\], ${UPTIME}, ${PROTOCOL}, ${TYPE}, tag ${TAG}\s+segid: ${SEGID}\s+tunnelid: ${TUNNELID}\s+encap: ${ENCAP} -> Record
  #
  # Match routes with no type, but with a tag (e.g. statics out SVIs with a tag)
  ^\s+\*+via ${NEXTHOP_IP}(%${NEXTHOP_VRF})?, ${NEXTHOP_IF}, \[${DISTANCE}\/${METRIC}\], ${UPTIME}, ${PROTOCOL}, tag ${TAG} -> Record
  #
  # Match standard dynamic route line (with or without tag)
  ^\s+\*+via ${NEXTHOP_IP}(%${NEXTHOP_VRF})?, ${NEXTHOP_IF}, \[${DISTANCE}/${METRIC}\], ${UPTIME}, ${PROTOCOL}, ${TYPE}(, tag ${TAG})? -> Record
  #
  # Match route without a type (e.g. direct, local)
[...]

PR #870, opened 2021-03-01
Fixs #869 cisco nxos show ip route not found

diepes:
##### ISSUE TYPE

 - Bugfix Pull Request
 - Additional Testing

##### COMPONENT

/templates/cisco_nxos_show_ip_route.textfsm

##### SUMMARY

add match  for "Route not found" resulting in empty return for the specific vrf

```

```


templates/cisco_nxos_show_ip_route.textfsm
```diff
@@ -15,10 +15,14 @@ Value TUNNELID (0x[a-f\d]+)
 Value ENCAP (\w+)
 
 Start
+  ^.+show\sip\sroute
```
FragmentedPacket, line 18: Do you mind removing this line and the line this accounts for in the output as well?

We typically don't have the command within the output.

FragmentedPacket (changes requested):

templates/cisco_nxos_show_ip_route.textfsm
diepes: Removed the command from the .raw and the match

diepes:
@FragmentedPacket please have a look again.  Removed the command from the raw template.

FragmentedPacket (approved):

PR #882, opened 2021-03-09
Migrate packaging to use poetry

jmcgill298:


.travis.yml
```diff
@@ -1,14 +1,29 @@
...
 python:
   - "3.6"
-
-sudo: false
+  - "3.7"
+  - "3.8"
 
 install:
-  - pip install tox
+  - "pip install virtualenv"
```
FragmentedPacket, line 12: Could we just use the builtin `venv` module here?

`python -m venv .venv`?

jmcgill298: I tried that at some point, and it did not like it. Travis should have the default pip with a version that supports installing cryptography soon, and this can go away.

jmcgill298:
@FragmentedPacket @jvanderaa are you good with these changes, or does this need more consideration?

FragmentedPacket (approved):

FragmentedPacket:
Looks good to me! Let's make sure to bump to major version for this and good readme notes.

PR #905, opened 2021-03-19
Template Change: cisco_ios, show archive

QuasarKid:
##### ISSUE TYPE

 - Template Addition Pull Request

##### COMPONENT

cisco_ios_show_archive.textfsm, cisco_ios, show_archive

##### SUMMARY

Added lines to capture filenames and most recent archive if archiving is enabled. Want to capture the index of the most recent to be able to reference which file in flash is startup-config. Relying on the implicit EOF -> Record in the case that the last line also has a filename in it. Otherwise we should record all the data the first empty index we match.


jvanderaa:
I think that this may be good to go, but @jmcgill298 take a quick look?

jmcgill298 (approved):

PR #943, opened 2021-06-29
cisco ios: fix empty and secure ports in mac addr-table

jacdavi:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

`cisco_ios_show_mac-address-table`, `cisco_ios`, `show mac address-table`
##### SUMMARY

I found a few cases that were breaking while trying things out for the first time on a handful of different switches.
In particular issues with static ports that had the "Secure" flag or no destination port set.

The changes address those two issues.
It may make sense to allow other values from the address-table legend (see raw output 9), but I haven't seen those so I left that as is.

Instead of adding a new test case, I appended to an existing one of the same format.
The new cases (along with all others) passed locally.

##### Input (truncated to specific entries):
```
   S     67 0000.0000.0004   static   No       -      Te1/5/11
 *      728 0000.0000.0003   static  Yes       -      
```

##### Before:
`"State Error raised. Rule Line: 32. Input Line:    S     67 0000.0000.0004   static   No       -      Te1/5/11"`

##### After):
```
...
	{
		"DESTINATION_ADDRESS": "0000.0000.0004",
		"DESTINATION_PORT": "Te1/5/11",
		"TYPE": "static",
		"VLAN": "67"
	},
	{
		"DESTINATION_ADDRESS": "0000.0000.0003",
		"DESTINATION_PORT": "",
		"TYPE": "static",
		"VLAN": "728"
	}
...
```

Still learning the ropes of both networking and TextFSM. Feel free to suggest changes.


FragmentedPacket:
@jacdavi Do you mind resolving the conflicts? There have been a few updates to the template that are conflicting with your change now.

jacdavi:
Ok should be good to go. #994 handled rows without a destination port by nature of switching to a list, so now this PR's only real change is to allow "Secure" (`S`) ports.

jmcgill298:
Thanks @jacdavi, unfortunately there was a change in the commit history, and so I have replaced this with #1688. It was hard to see what all you initially intended to change, but I have taken your examples and added to the test case and updated the template to parse those examples.

PR #945, opened 2021-06-29
Add ability to specify template dir as argument (and fallback to default dir)

jacdavi:
##### ISSUE TYPE

 - Python changes

##### COMPONENT

N/A

##### SUMMARY

This adds the ability to specify the templates directory in code via an argument to `parse_output`.
This option is more straightforward than setting an environment variable and adds increased flexibility in scripts.

Additionally, this PR adds a `try_fallback` option. When set, `parse_output` will first attempt to use the custom directory that was specified. 
If a matching template is not found in that directory, it will fallback to the default directory (which may be set by environment variable).
The purpose of this is that a user may want to maintain a small list of custom templates that are used first for parsing.
This is more convenient than having to copy the entire templates directory to a custom location, modifying the package files, or creating symlinks.

For example, I made a PR (#943) that makes changes to `cisco_ios_show_mac-address-table.textfsm`.
I'd like to use my new version, while maintaining the installed `ntc-templates` and using those for all other cases.
With these changes I can now create a directory as follows:
```
custom-templates
| - cisco_ios_show_mac-address-table.textfsm
| - index
```
Then utilize the custom template by calling:
```python
parse_output("cisco_ios", "show mac address-table" data, template_dir="custom-templates")
```

Additionally, I don't need to handle templates differently if wrapped in a function:
```
def my_parse(platform, command, data):
  return parse_output(platform, command, data, template_dir="custom-templates", try_fallback=True)

# uses custom template as above
my_parse("cisco_ios", "show mac address-table" data)
# silently uses default templtate
my_parse("cisco_ios", "show int status" data)
```

##### ADDITIONAL NOTES:

I made a few other small changes:
* Added documentation to method
* Instead of raising a generic `Exception` on a parsing error, the `CliTableError` is now re-raised the custom message (makes error handling easier)
* Updated readme to reflect changes

New changes have been run through `black`.
Feel free to suggest changes or close the PR if this is considered out-of-scope.


jmcgill298:
Thanks for this PR @jacdavi, as the history has changed since this PR, I have opened #1689 to replace it.

PR #982, opened 2021-09-04
fix parsing with int addresses = 0

dainok:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_nxos_show_cdp_neighbors_detail.textfsm

##### SUMMARY
If the output includes:

Interface address(es): 0

The parser fails.
The PR add verify non zero values:

^Interface address\(es\): ^[1-9]\d* -> GetInterfaceIP


jvanderaa (changes requested):
Please update the `ntc_templates/templates/cisco_nxos_show_cdp_neighbor_detail.textfsm` file instead of creating a new one in a new directory.

PR #995, opened 2021-09-22
Juniper Junos show interfaces extend capture

georgesnow:
extends the capture for junos on 

`show interfaces`

Capture descriptions if available on physical interfaces and capture local and destination IPs of logical interfaces


georgesnow:
I just notice I  need to clean up the commented out lines before this can be merged. I will fix that.

PR #997, opened 2021-09-24
Junos show chassis

georgesnow:
new template for junos show chassis hardware


georgesnow:
i should note this tempalte focuses on grabbing the part number, serial number, and description. it doesn't fully address position in the chassis. ie PIC/MIC/XCVR

georgesnow:
This now captures the FPC,MIC,PIC values for each and every Xcvr. Allowing for deriving what the interface is ie... xe-0/0/0. 

Also this contains the changes for show interfaces extension, which I can close if it is fine to roll all them up into one merge. Just let me know and I will close the other one.

ntc_templates/templates/juniper_junos_show_chassis_hardware.textfsm
```diff
@@ -0,0 +1,19 @@
...
+Start
+  ^\s+Hardware\s+
+  ^.*FPC\s+${FPC} -> Continue
+  ^\s+CPU.*
+  ^.*MIC\s+${MIC}\s+\S+\s+\S+\s+\S+.* -> Continue
+  ^.*PIC\s+${PIC}\s+\S+\s+\S+\s+\S+.* -> Continue
+  ^.*Xcvr\s+${XCVR}\s+(REV\s+\d+|)\s+${PART}\s+${SERIAL}\s+${DESCRIPTION} -> Continue.Record
+  ^\s*$$
+  ^(Power|Fan)\s+
+
```
jmcgill298, line 18: This is missing the EOF as defined at the end of this section: https://github.com/networktocode/ntc-templates#states

jmcgill298 (changes requested):

ntc_templates/templates/juniper_junos_show_chassis_hardware.textfsm
georgesnow: @jmcgill298 I just to clarify should this template and actually the other one (show interfaces) should like this:

```
^\s*$$
^(Power|Fan)\s+
^. -> Error
```
I didn't see the error line on the others and it passed the tests. So I had just matched to the others. Just let me know the above is correct and I will fix it.

jmcgill298: you will probably want to catch empty lines first.

```
...
  ^\s*$$
  ^. -> Error
```

This does require all lines to be matched (not necessarily captured), so if the template is ignoring some lines of output, it will need to be updated to account for those.

For the modifications of an existing template, we do not require it to be updated to have that line.

georgesnow: hmm ok. so there is a lot of variation in the starting lines between the platforms (MX, QFX, SRX, etc..discovered from testing locally). It worked against the MX and QFX in its current state. If that is a requirement I will need to add additional raw outputs and tests from the other platforms. I will be using against those platforms and if I add that error it will break the capture for QFXs.

jmcgill298: that is a requirement, as it is the only way we can have any assurance that the data that is intended to be captured is actually being captured. Otherwise it will ignore the relevant data and report back a subset of what is desired, and the end user would not be aware they had incorrect data

ntc_templates/templates/juniper_junos_show_interfaces.textfsm
```diff
@@ -2,14 +2,17 @@ Value Required INTERFACE (\S+)
...
+Value DESTINATION (\S+)
+Value LOCAL (\S+)
 Value MTU (\d+|Unlimited)
 
 Start
-  ^\s+Logical\s+interface\s+${INTERFACE}
   ^Physical\s+interface:\s+${INTERFACE},\s+${ADMIN_STATE},\s+Physical\s+link\s+is\s+${LINK_STATUS}
+  ^.*Description:\s+${DESCRIPTION}
   ^.*ype:\s+${HARDWARE_TYPE},.*MTU:\s+${MTU}.* -> Record
-  ^.*MTU:\s+${MTU}.* -> Record
```
jmcgill298, line 11: Was this intended to be deleted? If so, what is the reason to remove it?

jmcgill298 (changes requested):
I put in a PR to this branch to move the show interfaces to use `Continue.Record` as that is the appropriate pattern for that type of output. Per the tests in the branch, it results in the same output.

https://github.com/georgesnow/ntc-templates/pull/5

ntc_templates/templates/juniper_junos_show_interfaces.textfsm
georgesnow: yes, as it didn't appear to work in my tests against field gear it didn't work across multiple platforms.

georgesnow:
@jmcgill298  i think my changes made your change no longer needed. I added raw outputs for 4 different devices (MX10/80, MX104, MX204, and QFX5100).

and I tested the template against all of them and they appear to work.

georgesnow:
also i dont understand the assertion error. I tested the template using textfsm parser for raw3. There are 8 entries in the raw and the output. But the ntc-templates is only returning 7. 

I am not sure what's going on there. any thoughts? am I missing something?

ntc_templates/templates/juniper_junos_show_interfaces.textfsm
jmcgill298: I think that was because of the Record. My changes would make the Record deterministic on each new interface, so that should probably stay in as `^.*MTU\s+${MTU}`

jmcgill298:
@georgesnow what happens if you move `EOF` to the last state (end of the file)

georgesnow:
@jmcgill298 sweet that fixes it. i just pushed back the change after testing it

georgesnow:
@jmcgill298 i think all the outstanding issues are resolved and it's passing all the tests. is this good to merge?

PR #1001, opened 2021-10-13
Fixes cisco_wlc_ssh_show_interface_summary to match interfaces with other chars

AJatCDW:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_wlc_ssh_show_interface_summary, cisco_wlc_ssh, show interface summary

##### SUMMARY
Ran into an issue while using this template, apparently the WLC will accept a lot of different characters for the interface name. The only character that I found it didn't like was the single quote, it seems that all other characters are allowed. I modified the regex for the NAME value to include a negative lookaround so all of the characters in the interface name are matched but none of the trailing space until the Port number.

```

```


AJatCDW:
Seems that the Py3.7 build failed with 'ERROR: Could not install packages due to an OSError' - the other builds worked. Is there something up with 3.7 specifically?

PR #1005, opened 2021-10-20
Bugfix: Cisco ASA show failover: handle instance where no failover configured

ragzilla:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
bugfix: cisco_asa - show failover

##### SUMMARY
Current template does not handle instance where Failover LAN interface is not configured, or test for disabled failover. Added mock/test for unconfigured failover.


jmcgill298:
please open a pull request with only the intended files changed

PR #1070, opened 2022-02-02
Add Zyxel cfg lan get

elavaud:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Zyxel OS, cfg lan get


ntc_templates/templates/zyxel_os_cfg_lan_get.textfsm
```diff
@@ -0,0 +1,15 @@
...
+Value IPV4_DHCP (Enable|Disable)
+Value IPV6_ENABLE (true|false)
+
+Start
+  ^Group\sName\s+LAN\sIP\sAddress\s+IPv4\sDHCP\sServer\s+IPv6\sEnable\s*$$ -> LANTable
+  ^\s*$$
+  ^. -> Error
+
+LANTable
+  ^${NAME}\s+${IP}\s+${IPV4_DHCP}\s+${IPV6_ENABLE}\s*$$ -> Record
```
itdependsnetworks, line 12:
```suggestion
    ^${NAME}\s+${IP}\s+${IPV4_DHCP}\s+${IPV6_ENABLE}\s*$$ -> Record
    ^${NAME}\s+${IPV4_DHCP}\s+${IPV6_ENABLE}\s*$$ -> Record
```

ntc_templates/templates/zyxel_os_cfg_lan_get.textfsm
```diff
@@ -0,0 +1,15 @@
+Value NAME (.+?)
+Value IP (((\d+\.){3}\d+)?)
```
itdependsnetworks, line 2:
```suggestion
Value IP ((\d+\.){3}\d+)
```

itdependsnetworks: I think this works?? I don't think we should check for IP or nothing here. We should also standardize on a way to match an IP. 

@jvanderaa @jmcgill298 any thoughts?

elavaud: I do agree. And I was also thinking that I was using different ways to describe IPs. A standard would be welcome

elavaud:
The failure here seems unrelated to the code. Any way to retrigger the build?

elavaud:
Failure due to CI

PR #1072, opened 2022-02-02
Add Zyxel cfg nat addr map get

elavaud:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Zyxel OS, cfg nat addr map get


ntc_templates/templates/zyxel_os_cfg_nat_addr_map_get.textfsm
```diff
@@ -0,0 +1,19 @@
+Value INDEX (\d+)
+Value NAME (.+?)
+Value INTERFACE (\w+)
+Value TYPE (\S+?)
+Value LOCAL_START_IP ((\d+\.\d+\.\d+\.\d+)?)
```
itdependsnetworks, line 5: I don't think this pattern will actually work if the IPs are optional. LOCAL_START_IP will get conflated for LOCAL_END_IP as an example. If this actually exists in the real world, then would need tests, and likely another strategy to attack.

elavaud: Indeed. It should not be optional. Will correct

itdependsnetworks: I only commented on the one, but I meant it for all of them :)

elavaud: Ok, I fine tuned adding some tests with raw data I collected directly after manually configuring the address 
mapping through the UI. Let me know.
Sorry for the troubles. I am adding old work of a former colleague and I am new to networking.

jmcgill298 (commented):
needs rebase

elavaud:
Grouped in #1142

PR #1075, opened 2022-02-02
Add Zyxel cfg static dhcp get

elavaud:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Zyxel OS, cfg static_dhcp get


ntc_templates/templates/zyxel_os_cfg_static_dhcp_get.textfsm
```diff
@@ -0,0 +1,15 @@
+Value INDEX (\d+)
+Value STATUS (0|1)
+Value MAC_ADDRESS ((([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})?)
```
itdependsnetworks, line 3: Similar to others, if this doesn't exist in some cases, need to have tests and likely an alternative strategy.

elavaud: You are right again. It obviously should not be optional

itdependsnetworks (approved):

Issue #919, opened 2021-04-09
cisco ios show lldp neighbours details textfsm.parser.TextFSMError: State Error raised. Rule Line: 55. Input Line: Power via MDI:


RP-11:
##### ISSUE TYPE

 - Bug Report

##### TEMPLATE USING

```
Value LOCAL_INTERFACE (\S+)
Value CHASSIS_ID (\S+)
Value NEIGHBOR_PORT_ID (.*)
Value NEIGHBOR_INTERFACE (.*)
Value NEIGHBOR (.+?)
Value SYSTEM_DESCRIPTION (.*)
Value CAPABILITIES (.*)
Value MANAGEMENT_IP (\S+)
Value VLAN (\d+)
Value SERIAL (\S+)

Start
  ^.*not advertised
  ^.*Invalid input detected -> EOF
  ^.*LLDP is not enabled -> EOF
  ^Local\s+Intf:\s+${LOCAL_INTERFACE}\s*$$
  ^Chassis\s+id:\s+${CHASSIS_ID}\s*$$
  ^Port\s+id:\s+${NEIGHBOR_PORT_ID}\s*$$
  ^Port\s+Description:\s+${NEIGHBOR_INTERFACE}\s*$$
  ^System\s+Name(\s+-\s+not\s+advertised)\s*$$
  ^System\s+Name:?\s*$$
  ^System\s+Name(:\s+${NEIGHBOR})\s*$$
  ^System\s+Description -> GetDescription
  ^Time
  ^System\s+Capabilities
  ^Enabled\s+Capabilities:\s+${CAPABILITIES}\s*$$
  ^Management\s+Addresses
  ^\s+OID
  ^\s+[\d+\.]{8,}
  ^.*IP:\s+${MANAGEMENT_IP}
  ^Auto\s+Negotiation
  ^Physical\s+media
  # Removed \(\s+\) from the template - The line 'Other/unknown' would not be captured
  # Now looks for any text beginning with any space
  ^\s+.+\s*$$
  ^Media\s+Attachment
  ^\s+Inventory
  ^\s+Capabilities
  ^\s+Device\s+type
  ^\s+Network\s+Policies
  ^\s+Power\s+requirements
  ^\s+Location
  ^Time\s+remaining
  ^Vlan\s+ID:\s+(?:${VLAN}|-\s+not\s+advertised)\s*$$
  ^\s+\(\S+\)
  ^(?:PoE|\s+Power)
[...]

PR #1086, opened 2022-02-17
Improve cisco_ios show_lldp_neighbors_detail

lamiskin:
##### ISSUE TYPE

 - Additional Testing and improvements

##### COMPONENT

ntc_templates/templates/cisco_ios_show_lldp_neighbors_detail.textfsm
cisco_ios
show lldp neighbors details 

##### SUMMARY

 - Capture new power fields where available.
 - Capture chassis id's that contain spaces, but don't match `:` in the case another field is added on the same line to the ouptut in the future.
 - Tests added for both of these scenarios. 

```
  - local_interface: ""
    chassis_id: "0004.7a2a.bbbb"
    neighbor_port_id: "0004.7a2a.bbbb"
    neighbor_interface: "eth0"
    neighbor: "system2"
    system_description: "Dragonfly"
    capabilities: "S"
    management_ip: "10.132.14.52"
    vlan: ""
    serial: ""
    power_pair: "Spare"
    power_class: "4"
    power_device_type: "2"
    power_priority: "High"
    power_source: "PSE"
    power_requested: "25000"
```


ntc_templates/templates/cisco_ios_show_lldp_neighbors_detail.textfsm
```diff
@@ -1,5 +1,5 @@
 Value LOCAL_INTERFACE (\S+)
-Value CHASSIS_ID (\S+)
+Value CHASSIS_ID ([^:]*)
```
diepes, line 2: since matching anything not colon(:) maybe also exclude blank space ?
Should the match be nothing (*) or at least something (+) ?
e.g. ([^:\s]+)

ntc_templates/templates/cisco_ios_show_lldp_neighbors_detail.textfsm
```diff
@@ -63,6 +70,15 @@ IgnoreDescription
   ^\s*$$
   ^.* -> Error
 
+PowerMDI
+  ^\s+Power\s+Pair:\s+${POWER_PAIR}
```
diepes, line 74: As the variable match is at the end of the line, probably better to follow them with 
`\s*$$` to ensure no capture of spaces

lamiskin: Thanks @diepes I've pushed your suggested changes now.

ntc_templates/templates/cisco_ios_show_lldp_neighbors_detail.textfsm
lamiskin: @diepes Thanks for the feedback. I've pushed a commit now using `+` instead of `*`.

My example in the PR wasn't very clear for this case.
The output I get from a device is:
`Chassis id: S7-1200                   6DS7 212-1BH40-0XB0  S V-LCCC5820        10 V  4  3  1`

Do you think its valid to capture the output chassis-id as the following (including spaces)?:
`S7-1200                   6DS7 212-1BH40-0XB0  S V-LCCC5820        10 V  4  3  1`

mjbear:
resolves #919

PR #1102, opened 2022-03-11
Add Cisco IOS show power inline

Mathieu-Dirkx:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Cisco IOS - show power inline

##### SUMMARY
Add support for show power inline on Cisco IOS platform


ntc_templates/templates/cisco_ios_show_power_inline.textfsm
```diff
@@ -0,0 +1,28 @@
+Value Required INTERFACE (\S+)
+Value ADMIN (auto)
```
diepes, line 2: Is this the only value ADMIN can have ?  (auto)

ntc_templates/templates/cisco_ios_show_power_inline.textfsm
```diff
@@ -0,0 +1,28 @@
...
+# actual data
+  ^${INTERFACE}\s+${ADMIN}\s+${OPER}\s+${POWER}\s+${DEVICE}\s+${CLASS}\s+${MAX}\s*$$ -> Record
+  ^${INTERFACE}\s+${ADMIN}\s+${OPER}\s+${POWER}\s+${DEV_POWER}\s+${DEVICE}\s+${CLASS}\s*$$ -> Record
+# empty lines
+  ^\s*$$
+# Optional footer
+  ^Totals:\s+\d+\s+(on|off)\s+\d+\.\d+\s+\d+\.\d+\s*$$
+# Command unknown
+  ^% Invalid command .*$$ -> Error
+  ^.*$$
```
diepes, line 28: This "^.*$$" will capture anything/everything, you should try to end the state with "^. -> Error"

Quote form the README.md
"Each state should end with ^. -> Error. This helps to ensure we're accounting for every line within the raw output for the command. This doesn't mean we have to capture all the data as a Value, but we do have to account for it. In addition, it is also good to provide an expression to match blank lines, ^\s*$$"

diepes:
@Mathieu-Dirkx the PR should be for a single commit, you can re-base on the networktocode:master

ntc_templates/templates/cisco_ios_show_power_inline.textfsm
Mathieu-Dirkx: I have access to 434 Cisco IOS switches and (auto) is the only value observed. Probably other values are possible, but I cannot confirm or test.

ntc_templates/templates/cisco_ios_show_power_inline.textfsm
Mathieu-Dirkx: Template is updated as described

Mathieu-Dirkx:
I think I re-based my request.
Have never done this before, so can someone confirm it is okay?

PR #1105, opened 2022-03-14
Add Cisco IOS XR show install active

verbosemode:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT

- cisco_xr_show_install_active.textfsm, cisco_xr, show install active

##### SUMMARY

This command allows you to get a list of the current active packages per hardware component (route processor, line card, fabric module).

With the output you can verify that the installed packages on a give hardware type, like all line cards, is consistent. It usually should be the case, but I've currently got an issue on XR, where different versions were installed on different line cards. This is still subject to a TAC case however :-)

The test cases include the output from a 64-bit and from a 32-bit IOS XR router.


verbosemode:
The Travis CI task fails during "poetry install" and doesn't seem to be related to my changes. Could someone restart that CI job? Running tox locally worked without any issues.

jvanderaa:
Restart completed.

jvanderaa (approved):

PR #1106, opened 2022-03-21
Add Cisco IOS show power inline

Mathieu-Dirkx:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Cisco IOS - show power inline

##### SUMMARY
Add support for show power inline on Cisco IOS platform


jvanderaa:
@jmcgill298 I believe this is a start on a module set. We had a discussion at one point about what this template would look like. Since the output gives both interface level stats and the entire chassis/virtual switch details.

ntc_templates/templates/cisco_ios_show_power_inline.textfsm
```diff
@@ -0,0 +1,28 @@
...
+Value MAX (\d+\.\d)
+
+Start
+# Checking for modules headers
+  ^Module\s+Available\s+Used\s+Remaining\s*$$
+  ^Available:\d+(\.\d+)?\(w\)\s+Used:\d+(\.\d+)?\(w\)\s+Remaining:\d+(\.\d+)?\(w\)\s*$$
+  ^[\s(Wats)]*$$
+  ^[-\s]*$$
+  ^\d+\s+(n\/a|\d+\.\d+)\s+(n\/a|\d+\.\d+)\s+(n\/a|\d+\.\d+)\s*$$
+  ^Interface\s+Admin\s+Oper\s+Power(\(Watts\))?\s+Device\s+Class.*$$
```
diepes, line 17: You can use the heading for the table to jump to new parsing section `^match -> Table`
this keep the header matching and the table parsing separate, less chance of matching something by accident.

ntc_templates/templates/cisco_ios_show_power_inline.textfsm
```diff
@@ -0,0 +1,28 @@
...
+Value DEV_POWER (\d+\.\d+)
+Value DEVICE (.+?)
+Value CLASS (n\/a|\d)
+Value MAX (\d+\.\d)
+
+Start
+# Checking for modules headers
+  ^Module\s+Available\s+Used\s+Remaining\s*$$
+  ^Available:\d+(\.\d+)?\(w\)\s+Used:\d+(\.\d+)?\(w\)\s+Remaining:\d+(\.\d+)?\(w\)\s*$$
+  ^[\s(Wats)]*$$
```
diepes, line 14: Spelling mistake, will never match Watts with two t's

diepes:
@Mathieu-Dirkx please re-base the PR on the master so it becomes a single commit.

jmcgill298 (changes requested):
see @diepes comments

Mathieu-Dirkx:
I tried to re-base but nothing changed.
I do not know how to do what you want me to do.
I will no longer waste your time and close the pull-request.

PR #1116, opened 2022-04-06
Cisco ASA show access-list does not support mixed objects and fqdn

pertoft:
##### ISSUE TYPE

 - New Template Pull Request

##### COMPONENT

./ntc_templates/templates/cisco_asa_show_access-list.textfsm
show access-list

##### SUMMARY

Cisco ASA show access-list does not support mixed objects and fqdn
Example:
access-list test line 40 extended permit tcp object-group TEST-OBJ object host.example.com eq https (hitcnt=3228758) 0x3e3d87a5
  access-list test line 40 extended permit tcp host 10.0.1.10 fqdn host.example.com (unresolved) eq https 0xc6e4eca7

```

```


diepes (changes requested):
@pertoft rather than changing the existing test sample 
tests/cisco_asa/show_access-list/cisco_asa_show_access-list_update.raw

rather add a new sample .raw with corresponding .yml.
then run tox to ensure all the tests still pass.

Other option is to log a new issue, providing the ASA output for the show access-list that does not work, then we can add it as an additional test.

PR #1100, opened 2022-03-08
Fix cisco_ios show ip bgp neighbors advertised-routes (metric)

viktorkertesz:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
cisco_ios show ip bgp neighbors advertised-routes

testing is windows compatible now

##### SUMMARY
Metric can be 10 chars long and IOS router actually keep the static format by metric printing. This is a fix for supporting longer metric numbers.

When I tried the same with localpref, the router ruined the format (therefore new matching would be needed). I don't have use case for more than 6char long localpref, so I ignored that fact.

Fix testing path handling so windows can run the tests.

PR #1123, opened 2022-04-16
fixed metric width when it is wider than 6 chars

viktorkertesz:
##### ISSUE TYPE
 - Bugfix Pull Request

This is a split of #1100 

##### COMPONENT
cisco_ios show ip bgp

##### SUMMARY
Metric can be 10 chars long and IOS router actually keep the static format by metric printing. This is a fix for supporting longer metric numbers.

When I tried the same with localpref, the router ruined the format (therefore new matching would be needed). I don't have use case for more than 6char long localpref, so I ignored that fact.


tests/cisco_ios/show_ip_bgp/cisco_ios_show_ip_bgp6.raw
```diff
@@ -0,0 +1,15 @@
...
+RPKI validation codes: V valid, I invalid, N Not found
+
+     Network          Next Hop            Metric LocPrf Weight Path
+ *>  11.222.33.123/29 111.222.123.202    1762560    350      0 65000 ?
+ *>  11.222.33.123/29 111.222.123.202 4000000000    350      0 65000 ?
+ *>  11.222.12.0/23   11.222.1.202          3072    350      0 65000 ?
+ *>  123.233.111.222/24
+                       11.222.1.202         28416    350      0 65000 ?
+
+Total number of prefixes 4
```
jmcgill298, line 15:
```suggestion
Total number of prefixes 4

```

jmcgill298 (changes requested):

PR #1152, opened 2022-06-08
Mirkotik: Fix issue with flags on IP route print and long IPv6 addresses truncated

elavaud:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
Mikrotik router OS
ip route print without-paging

##### SUMMARY
* Flags were not well parsed. Trying first to trim the flags when parsing, but output makes it too difficult.
* Long IPv6 addresses where truncated, making the parsing not possible

-> Switching to command "ip route print terse" and updating parsing of the flags (trimming should be done after)

I updated the tests cases to include the flag issue as well as 2 different tests: one for IPv4 and one for IPv6.


elavaud:
Fix #1195

jvanderaa (approved):

PR #1159, opened 2022-06-14
fix(Mikrotik): Optional space before end of line

elavaud:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
Mirkotik router OS
* ip arp print without-paging
* ip firewall filter print all without-paging
* ip dhcp-server lease print without-paging

##### SUMMARY
Depending of the connection client used, the space at the end of the line might be trimmed. 
With this fix the space(s) at the end of the line become optional


jmcgill298:
Thanks @elavaud

PR #1189, opened 2022-09-05
Update cisco_ios_show_interfaces_description.textfsm

yevgeniy-voloshin:
All other 'show interface(s) description' templates used DESCRIPTION instead of short DESCRIP.
So I believe that the same key ('description') for all dictionaries much better than 'descrip' for cisco_ios.

P.S. also for arista_eos_show_interfaces_description.textfsm:Value DESCRIP (\S.*?)


yevgeniy-voloshin:
Dear Colleagues, I created the same request for Arista in #1190, but there are "All checks have failed"... I am confused
What am I doing wrong?

jvanderaa:
This will be a breaking change. We will need to account for this and plan for a major release.

jvanderaa:
Was also covered in another PR. Going to include reference to this pr.

PR #1197, opened 2022-10-06
Fix Zyxel sys atsh template when there message before data

k-ribot:
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT
Zyxel os, sys atsh

##### SUMMARY
Some time there is a message before expected data. So we need to bypass it because an error is raise


whitej6:
@k-ribot can you reach out to me via [jeremy.white@networktocode.com](mailto:jeremy.white@networktocode.com) This PR (and the several others you submitted) were considered part of hacktoberfest and Network to Code would like to show our appreciation for your contributions.

PR #1207, opened 2022-10-17
Fix cisco_xr_show_arp with Management/RP interface present

charlesmonson:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT
cisco_xr_show_arp

##### SUMMARY

Update cisco_xr_show_arp to  catch the presence of "RP" in the CPU Filldown variable when Management interfaces are in use.


jvanderaa:
Thank you @charlesmonson for the contribution!

PR #1222, opened 2022-11-15
Fix huawei display acl all - port numbers replaced by names and IPV6 ACL name position

elavaud:
ISSUE TYPE
- Bugfix Pull Request
- Additional Testing

COMPONENT
Huawei VRP
display acl all

SUMMARY

First commit:
Huawei will automatically replace well known ports by their corresponding names. For example 80 by "www", or 23 by "telnet"
I also improved the test by adding a range of ports

Second commit:
When available, the position of the name change between IPv4 and IPv6.

Third commit:
ACL rules can have descriptions

Fourth commit:
Always record first the ACL with empty rules values, and then its rules. Otherwise it was not possible to record ACL that has no rules


elavaud:
Hey @itdependsnetworks,

I noticed another problem with ACLs, and I am unsure how to best handle it:

There are two types of object fetched by this template: the ACL, and a rule of an ACL.
In most cases we can easily handle it after and separate the ACL from the rule.
However if the ACL does not have any rule, its information are not parsed.
For example in the current raw test file we have:
```
Advanced ACL qsdqsd 3999, 0 rule
Acl's step is 5
```
We don't get it parsed.
And although its has no rules, it is important to know that the ACL 'qsdqsd' at 3999 exists...

So I guess the broader problem is: Are we fetching ACL? Or only its rules? Or both (and how to do it) ? :)

itdependsnetworks:
Yes, I think you should capture it, even if the rule is empty

elavaud:
> Yes, I think you should capture it, even if the rule is empty

With the "rule description" being on an other line, the only solution I found was to first record each ACL name and number with empty rules values, then each rules.
Not super pretty but at least it is homogeneous. And again, handling 2 types of object with only one list/parsing is tricky...

itdependsnetworks (approved):

PR #1227, opened 2022-11-27
Added support for 7.0

klausmark:
Added fortigate 7.0 support for get system status.


ntc_templates/templates/fortinet_get_system_status.textfsm
```diff
@@ -70,7 +75,6 @@ Start
   ^Branch\s+point:\s+${BRANCH_POINT}
   ^Release\s+Version\s+Information:\s+${RELEASE_VERSION_INFORMATION}
   ^FortiOS\s+x86-64:\s+${FORTIOS_X86_64}
-  ^System\s+time:\s+${SYSTEM_TIME} -> Record
-  ^\s*$$
-  ^. -> Error
```
itdependsnetworks, line 75: This is there for a reason, can you add these two lines back?

klausmark: But it turns out that the "get system status", does not always end in the same order, so either we make a 7.0 template or we accept that EOF is the point where we record. 
On another note. If you look at many of the other templates, it's not custom that it throws an error if there are stuff we can't parse. It off cause was the main reason that i changed this template, but on the other hand, it would probably have work sufficiently for me with the old template, if it didn't raise.

jmcgill298: @klausmark there is no need to have a `Record` in this template, as there is an implicit `EOF -> Record` provided by TextFSM.

ntc_templates/templates/fortinet_get_system_status.textfsm
```diff
@@ -70,7 +75,6 @@ Start
...
-
+  ^System\s+time:\s+${SYSTEM_TIME}
+  ^Last\s+reboot\s+reason:\s+${LAST_REBOOT_REASON}
+  ^EOF -> Record
```
jmcgill298, line 80:
```suggestion
  ^\s*$$
  ^. -> Error
```

This works when I run the tests against the three test cases

ntc_templates/templates/fortinet_get_system_status.textfsm
jmcgill298: We want unmatched lines to error out, otherwise we do not have much confidence that the output is being parsed correctly. If additional lines are not being matched, then the template needs to be updated to account for those lines.

jmcgill298 (approved):

jmcgill298:
Thanks @klausmark

PR #1238, opened 2022-12-07
fixed asa_show_run_object_network_fqdn_issue

vsurresh:
#### ISSUE TYPE

Bugfix Pull Request

#### COMPONENT
cisco_asa, cisco_asa_show_running-config_object_network.textfsm

#### SUMMARY
The current template doesn't take FQDN into account therefore if you try to parse an output with FQDN, textfsm throws an error. 
```     raise TextFSMError('State Error raised. Rule Line: %s. Input Line: %s'
textfsm.parser.TextFSMError: State Error raised. Rule Line: 21. Input Line:  fqdn google.com
```


tests/cisco_asa/show_running-config_object_network/cisco_asa_show_running-config_object_network.raw
```diff
@@ -2,15 +2,13 @@ object network corp
  subnet 10.0.0.0 255.0.0.0
 object network internal
  subnet 192.168.0.0 255.255.255.0
-object network any
```
itdependsnetworks, line 5: Why is this being adjusted?

tests/cisco_asa/show_running-config_object_network/cisco_asa_show_running-config_object_network.raw
```diff
@@ -2,15 +2,13 @@ object network corp
...
- subnet 0.0.0.0 0.0.0.0
 object network host-10.10.11.1
  host 10.10.11.1
 object network block
  host 10.75.51.11
 object network dmz
- description dmz GW ip
```
itdependsnetworks, line 12: Why is this being adjusted?

tests/cisco_asa/show_running-config_object_network/cisco_asa_show_running-config_object_network.raw
vsurresh: ASA won't allow the creation of objects with the name 'any'  so, I had to remove it from my test. 

```
asa_test(config)# object network any
ERROR: This object name is reserved
```

tests/cisco_asa/show_running-config_object_network/cisco_asa_show_running-config_object_network.raw
vsurresh: The description of the object goes to the next line (after the host)

After:

```
object network dmz
 host 11.1.2.2
 description dmz GW IP
```

Before:
```
object network dmz
  description dmz GW IP
  host 11.1.2.2
```

jmcgill298:
@vsurresh it looks like there are some changes that aren't necessary and the yamllint check is failing. It doesn't really matter that `any` is reserved and can't be configured, it is just a fair representative of the type of data that would be there (a string within certain length constraints). Also, the template should be able to handle whatever order the `host` and `description` fields appear; imo, the test data would be better to have representation of both order (who knows if that is consistent across all platforms and all versions?).

I think if the changes just added an entry to match FQDN objects, that there will be much less change to the test data, and the tests will pass pretty easily.

jmcgill298:
replaced by https://github.com/networktocode/ntc-templates/pull/1752

PR #1247, opened 2022-12-19
Fix Cisco show interfaces speed and duplex

guillaume-mbali:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

cisco_ios_show_interfaces.textfsm, ios, show interfaces

##### SUMMARY
Correction of the regex allowing to recover the speed, the duplex and the media type


ntc_templates/templates/cisco_ios_show_interfaces.textfsm
```diff
@@ -9,7 +9,7 @@ Value IP_ADDRESS (\d+\.\d+\.\d+\.\d+\/\d+)
 Value MTU (\d+)
 Value DUPLEX (([Ff]ull|[Aa]uto|[Hh]alf|[Aa]-).*?)
 Value SPEED (.*?)
-Value MEDIA_TYPE (\S+(\s+\S+)?)
+Value MEDIA_TYPE (\S+[\s\S]*)
```
itdependsnetworks, line 12: What is this addressing? Can you add a test for that? 

Additionally, `\s` == white space and `\S` == non white space, I think can simply be 

`(\S+.*)`

guillaume-mbali: The value of media type can contain strings of characters alternated with spaces. I added some tests about this.

PR #915, opened 2021-03-31
fix for #914

armartirosyan:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT

`cisco_ios_show_ip_bgp_neighbors.textfsm` `Cisco IOS` `show ip bgp neighbors`

##### SUMMARY

This pull request should fix the bug described in #914 
The current problem with the template is that if the `show ip bgp neighbors` contains `vrf` in the output the `neighbor` and `remote_as` fields don't get matched.
I added `VRF` value and modified the regex to look for VRF if it is present.

**Before**
```
Value NEIGHBOR (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
Value REMOTE_AS (\d+)
Value PEER_GROUP (\S+)
Value REMOTE_ROUTER_ID (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
Value BGP_STATE (\w+)
Value LOCALHOST_IP (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
Value LOCALHOST_PORT (\d+)
Value REMOTE_IP (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
Value REMOTE_PORT (\d+)
Value INBOUND_ROUTEMAP (\S+)
Value OUTBOUND_ROUTEMAP (\S+)

Start
  # Capture first line, which shows the BGP neighor and remote AS number
  ^BGP\s+neighbor\s+is -> Continue.Record
  ^BGP\s+neighbor\s+is\s+${NEIGHBOR},\s+remote\s+AS\s+${REMOTE_AS}
  # Capture BGP peer group
  # Example: 'Member of peer-group RR_SERVERS for session parameters'
  ^\s*Member\s+of\s+peer-group\s+${PEER_GROUP}
  # 
  # Capture remote router ID
  # Example: ' BGP state = Established, up for 7w3d'
  ^.+remote\s+router\s+ID\s+${REMOTE_ROUTER_ID}
  #
  # Capture BGP state
  # Example: ' BGP state = Established, up for 7w3d'
  ^\s+BGP\s+state\s+=\s+${BGP_STATE}
  #
  # Capture Inbound/Outbound Route Maps
  # Example: 'Route map for incoming advertisements is BGP_Vendor_in'
  # Example: 'Route map for outgoing advertisements is BGP_Vendor_out'
[...]

PR #1250, opened 2022-12-19
Armartirosyan/cisco ios show ip bgp neighbors

jmcgill298:


jmcgill298:
replaces https://github.com/networktocode/ntc-templates/pull/915

PR #864, opened 2021-01-29
Add support for Arista EOS "show ipv6 bgp summary" and fix bug in "show ip bgp summary"

jsakofsky:
##### ISSUE TYPE
 - New Template Pull Request
 - Bugfix Pull Request

##### COMPONENT
new: arista_eos_show_ipv6_bgp_summary
fix: arista_eos_show_ip_bgp_summary

##### SUMMARY
Added support for "show ipv6 bgp summary" based on the existing show ip bgp summary template.
Decided to add a new template due to changes to the bgp neighbor regex capture. There was also a bug in the existing show ip bgp summary template that broke parsing when the state indicated "Idle(NoIf)" due to attempt to capture a space that doesn't exist after field. I made the space optional, in case there was a circumstance where a space might exist.

PR #1264, opened 2022-12-24
added arista ipv6 bgp sum template, fixed bug in ip bgp sum template …

itdependsnetworks:
…as well

Replace #864


itdependsnetworks:
FIxed already in #1232

PR #1272, opened 2022-12-29
Fix HP Comware displ inteface

dainok:
Adjust template with data coming from production switches (case insensitive, additional output dropped.
Fix:  ntc_templates/templates/hp_comware_display_interface.textfsm
Add: new tests


jmcgill298:
Thanks @dainok

PR #1282, opened 2023-01-12
Feature/cisco ios ping fail

KeithETruesdell:
adding a new template for ping command on the cisco ios with several tests included


ntc_templates/templates/cisco_ios_ping.textfsm
```diff
@@ -0,0 +1,23 @@
+Value Required SENT_QTY (\d+)
+Value Required SENT_TYPE (.*)
+Value Required DESTINATION (\S+)
+Value Required TIMEOUT (\d+)
+Value SOURCE (\S+)
+Value Required RESPONSE_STREAM ([\.\!]+)
```
jmcgill298, line 6: this should probably be a List as sending something like 100 pings could wrap to a second line

ntc_templates/templates/cisco_ios_ping_df.textfsm
```diff
@@ -0,0 +1,22 @@
+Value Required SENT_QTY (\d+)
```
jmcgill298, line 1: it looks like this is a duplicate of the above template

jmcgill298 (commented):
Thanks @KeithETruesdell, just a couple of comments to address

ntc_templates/templates/cisco_ios_ping_df.textfsm
KeithETruesdell: thank you, that was a duplicate, i was trying to test the DF function as part of a ping and forgot about this... i will remove

ntc_templates/templates/cisco_ios_ping.textfsm
KeithETruesdell: thank you for this...I did test things with 100 and 200, i probably should have a template for that to show, it did wrap visually on a second line but there was no line break so it still kept it in the variable...this could have also been due to how I was testing thing...
I will make this a list and add a test for that, thank you

KeithETruesdell: good call with the list
i added that, added a 200 ping test and adjusted the other tests
i don't suppose you know of a way to just run a single test?

KeithETruesdell:
updated the issues and re-requesting a review
thank you!

ntc_templates/templates/cisco_ios_ping.textfsm
jmcgill298: I think you would have to call pytest locally. https://docs.pytest.org/en/7.1.x/how-to/usage.html#specifying-which-tests-to-run

PR #1285, opened 2023-01-15
Fixed issue with timezones that have a / in them such as GMT/BST

sjhloco:
**ISSUE**
- Timezones that have  "/" in them such as GMT/BST caused the parsing to fail

**TEMPLATES**
- cisco_asa_show_failover
- cisco_asa_show_vpn-sessiondb_anyconnect


jmcgill298:
thanks @sjhloco

PR #1286, opened 2023-01-20
Update Cisco AireOS WLC show inventory

jversch6:
Some devices have the following lines present
`Power Supply 1................................... Present, OK
Power Supply 2................................... Absent`

And that's where it broke for me. This change doesn't record that information but at least it runs.


jmcgill298:
@jversch6 can you please update a test case to include an example line for this addition?

jversch6:
I think I just did.

PR #1307, opened 2023-02-17
NXOS show ip int fails without WCCP

dainok:
Template cisco_nxos_show_ip_interface.textfsm 

In some cases "WCCP Redirect" is not present, so I fixed template and added a test.


dainok:
Hello @jvanderaa sorry if I'm bothering you for this: anything I can do to make this approved?

jvanderaa:
Will see about getting a release out soon too, will try to process a bunch soon. These small and tested changes are *terrific*

dainok:
I totally agree. Let me know if I can help.

PR #1310, opened 2023-03-01
Added support for show_ip_bgp_all_summary_vrf

sjhloco:
I submitted [PR #1300](https://github.com/networktocode/ntc-templates/pull/1300) the other week to add a variable for the address family, however didn't realise at the time that this command (*show ip bgp summary vrf all*) will only ever return the IPv4 address families. To get multiple address families (IPv4 and L2VPN) we need to be running *show ip bgp ***all*** summary vrf all*.

Both command outputs are formatted exactly the same so are matched by this existing template. It seems pointless to create a new template for this so I added regex to the index so that it matches both commands and renamed the test file I added in the last PR to match the command. Let me know if you prefer to do this some other way, thanks.


jmcgill298:
replaced by https://github.com/networktocode/ntc-templates/pull/1765

PR #1314, opened 2023-03-06
Add Identifier as catch header for cisco ios

pierok13:
Some IOS switches use Identifier instead of Mac address as header. Added both headers. show authen ses

Interface    Identifier     Method  Domain  Status Fg Session ID
Gi1/0/25     1234.5678.9101  mab     DATA    Auth      0AF3B0050000001E00023F09


jvanderaa:
Please add some test data for this.

andershagman:
Is this solved?
I solved it like this.
Change:
^Interface\s+MAC Address\s+Method\s+Domain -> Catch

To:
^Interface\s+(?:MAC Address|Identifier)\s+Method\s+Domain -> Catch

jmcgill298:
replaced by https://github.com/networktocode/ntc-templates/pull/1766

PR #1315, opened 2023-03-08
Hp comware new displ vlan lldp

dainok:
Add templates for HP Comware:
- hp_comware_display_lldp_neighbor-information_list.textfsm
- hp_comware_display_vlan_all.textfsm

To avoid micro PR, I also patched:
- hp_comware_display_interface
- cisco_ios_show_vrf

WARNING: I added error management (^. -> Error) to cisco_ios_show_vrf. Tests are ok, let me know if you want I remove it.


jvanderaa:
Will get this one in, but would prefer the micro-prs for future as well so we can keep the conversation focused and to the individual items.

jvanderaa:
Thank you @dainok

PR #1323, opened 2023-03-14
Add Cisco IOS XR - show l2vpn mac-learning mac all location

stiltzkin10:
* Add Cisco IOS XR - show l2vpn mac-learning mac all location


jmcgill298:
replaced by https://github.com/networktocode/ntc-templates/pull/1767

PR #1324, opened 2023-03-14
Fix Cisco IOS show crypto ipsec sa detail - Including letters in SA_SIBLING_FLAGS.

nsnelson402:
ISSUE TYPE

Bugfix Pull Request

COMPONENT

template: cisco_ios_show_crypto_ipsec_sa_detail.textfsm

os: cisco ios

command: show crypto ipsec sa detail

SUMMARY

SA Sibling Flags can include letters. Causing the template to fail when they do. This fixes that.


jmcgill298:
replaced by https://github.com/networktocode/ntc-templates/pull/1768

PR #1346, opened 2023-04-05
Fix boot optimisation to match whitespace

danielno8:
Getting a state error from a Cisco 2960-XR

BOOT path-list      : flash:c2960x-universalk9-mz.152-7.E0a.bin
Config file         : flash:/config.text
Private Config file : flash:/private-config.text
Enable Break        : yes
Manual Boot         : no
Allow Dev Key         : yes
HELPER path-list    :
Auto upgrade        : yes
Auto upgrade path   :
Boot optimization   : disabled
NVRAM/Config file
      buffer size:   524288
Timeout for Config
          Download:    0 seconds
Config Download
       via DHCP:       disabled (next boot: disabled)


ntc_templates/templates/cisco_ios_show_boot.textfsm
```diff
@@ -40,12 +40,12 @@ Start
   ^Load\s+for\s+
   ^Time\s+source\s+is
   ^Boot\s*Mode\s*=\s*${BOOT_MODE}
-  ^Boot\soptimization\s:\s${BOOT_OPTIMIZATION}
+  ^Boot\s*optimization\s*:\s*${BOOT_OPTIMIZATION}
```
jvanderaa, line 43:
```suggestion
  ^Boot\s+optimization\s+:\s+${BOOT_OPTIMIZATION}
```

jvanderaa: Would just like to have the minimum one space caught.

danielno8: Thanks that also works

jmcgill298:
fixed in https://github.com/networktocode/ntc-templates/pull/1623

Issue #1135, opened 2022-05-18
cisco access list parse error


pertoft:
##### ISSUE TYPE

 - Bug Report

##### TEMPLATE USING

Latest master commit e72650d6c3508937e3e31ca79805cc336433e9a5
```
Value Required ACL_NAME (\S+)
Value ACL_TOT_ELEM (\d+)
Value ACL_NAME_HASH (0x\w+)
Value TYPE (standard|extended)
Value LINE_NUM (\d+)
Value REMARK (.+?)
Value ACTION (permit|deny)
Value PROTOCOL ([a-z]+)
Value SVC_OBJECT_GRP (\S+)
Value SVC_OBJECT (\S+)
Value SRC_INTFC (\S+)
Value SRC_OBJECT_GRP (\S+)
Value SRC_OBJECT (\S+)
Value SRC_HOST (\S+)
Value SRC_NETWORK (\d+\.\d+\.\d+\.\d+)
Value SRC_MASK (\d+\.\d+\.\d+\.\d+)
Value SRC_ANY (any[46]{0,1})
Value DST_INTFC (\S+)
Value DST_OBJECT_GRP (\S+)
Value DST_OBJECT (\S+)
Value DST_HOST (\S+)
Value DST_NETWORK (\d+\.\d+\.\d+\.\d+)
Value DST_MASK (\d+\.\d+\.\d+\.\d+)
Value DST_ANY (any[46]{0,1})
Value DST_PORT (\S+)
Value DST_PORT_GRP (\S+)
Value DST_PORT_OBJECT (\S+)
Value LOG_LEVEL ([a-z0-9]+)
Value LOG_INTERVAL (\d+)
Value STATE (inactive)
Value HIT_COUNT (\d+)
Value LINE_HASH (0x\w+)
Value ENTRY_PROTOCOL_ICMP (icmp)
Value ENTRY_PROTOCOL ([a-z\-]+)
Value ENTRY_SRC_FQDN (\S+)
Value ENTRY_SRC_RANGE_START (\d+\.\d+\.\d+\.\d+)
Value ENTRY_SRC_RANGE_END (\d+\.\d+\.\d+\.\d+)
Value ENTRY_SRC_HOST (\d+\.\d+\.\d+\.\d+)
Value ENTRY_SRC_NETWORK (\d+\.\d+\.\d+\.\d+)
Value ENTRY_SRC_MASK (\d+\.\d+\.\d+\.\d+)
[...]

PR #1351, opened 2023-04-11
Added new values for destination port lt/gt/range, IPv6, time-range o…

SaschaSchwarzKyndryl:
Added new values for destination port lt/gt/range, IPv6, time-range objects. Type and Action variable now also working for subrules/entries.
Parsing issues solved from issues:
cisco access list parse error #1135
cisco show access list parse error #2 #1137
cisco show access list parse error #3 #1138
Cisco ASA output of command "show access-list" is not properly parsed #1279


mjbear:
resolves #1135 
resolves #1137 
resolves #1138 
resolves #1279

PR #1368, opened 2023-05-05
Template and tests for cisco_nxos_show_ip_msdp_summary_vrf_all

alicegenkin:


jvanderaa:
@alicegenkin looks like we need some updates to the yaml format.

PR #1372, opened 2023-05-22
feat(cisco): add vrf template detail

guillaume-mbali:
ISSUE TYPE
Feat Template Pull Request

COMPONENT
cisco_ios_show_vrf_detail.textfsm, ios, show vrf detail

SUMMARY
Cisco adding vrf template


jvanderaa (commented):
I believe the only thing we need here yet is a `^. -> Error` for output that is not accounted for.

PR #1377, opened 2023-06-02
Fixs 1363 arista eos show interfaces - description quotes

diepes:
- Add new test.
- Update "ntc_templates/templates/arista_eos_show_interfaces.textfsm" DESCRIPTION to handle quoted string
- sort existing template (After running tests `invoke tests` )


jvanderaa:
Thanks @diepes

PR #1409, opened 2023-07-24
Add hp_procurve_show_ip_route

dainok:
Add hp_procurve_show_ip_route for HP Procurve switches.


dainok:
@jvanderaa template fixed using new naming convention

PR #1410, opened 2023-07-25
add support for bgp summary found on 7124SX

stesla:
The output in the raw sample here was collected from an Arista DCS-7124SX-R switch running EOS version 4.9.5.2.


jvanderaa (approved):

jvanderaa:
Thank you @stesla

PR #1413, opened 2023-07-27
some 7124sx do not output anything for pfxrcd

stesla:
Found another case with our 7124SX models that omit values for the PfxRcd column. Here's an update to my previous submission to handle that.


jmcgill298 (approved):

jmcgill298:
Thanks @stesla

PR #1414, opened 2023-07-31
Cisxo XR sh ipv4 int: ignore lines

dainok:
Ignore LACP and RPF lines on Cisco XR show ipv4 interfaces


dainok:
Hi @jvanderaa I missed the tests have failed. Now everything is fixed.
If you don't mind, please check 4 more PRs: https://github.com/networktocode/ntc-templates/pulls/dainok

ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
```diff
@@ -16,6 +16,7 @@ Start
   ^\s+MTU\sis\s${MTU}\s.*$$
   ^\s+Helper.*$$
   ^\s+Multicast.*$$
+  ^\s{4,}224\..*$$
```
jvanderaa, line 19: I'm not catching what this is doing in the template, if it is around multicast we have a few different numbers that this can be. (That is what I'm thinking of when I see `224`.

dainok: Exactly. In my tests, the line only include local multicast groups only (224.0.0.0/24). I could add all multicast addressees, but I don't have evidences.
Let me know.

jvanderaa: Is this line trying to capture it? `^\s+2(?:2[4-9]|3\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]\d?|0)){3}$$` and that just needs to be updated?? That doesn't look quite right.

jvanderaa: I think we may want to just move that line to this - `^(22[4-9]|23\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]\d?|0)){3}$$`

jvanderaa (approved):

PR #1438, opened 2023-08-12
Enhancement: mikrotik_routeros_ip_firewall_address-list

mjbear:
Enhancement: mikrotik_routeros_ip_firewall_address-list
- ADDRESS capture group > IP_ADDRESS
- split IP_ADDRESS so an optional PREFIX_LENGTH is separated out
- add IPV6_ADDRESS capture group and an extra record line to isolate instances of IPv6

Breaking change on account of the capture group changes


itdependsnetworks:
should we split ip from ipv6?

mjbear:
> should we split ip from ipv6?

Hello @itdependsnetworks 

There are a few templates that contain both IPv4 and IPv6 which each have their own capture group.

`ntc_templates/templates/checkpoint_gaia_show_interfaces_all.textfsm`
`ntc_templates/templates/linux_ip_address_show.textfsm`
`ntc_templates/templates/paloalto_panos_show_interface_management.textfsm`

If we split out the capture groups it is easier to "consume" those two pieces of structured data. Without this PR checking the IP address value via regex would be mandatory. (It would still be good practice to validate structured data values in the code using it. :grinning:)

For the bigger picture, as part of the end goal to normalization we will be able to use IP_ADDRESS for IPv4 addresses and IPV6_ADDRESS for IPv6 addresses.

I'm certainly open to further discussion as I want any changes to better the user experience and project.

PR #1443, opened 2023-08-14
add cisco template for show file systems

pedro-damasceno:


pedro-damasceno:
Fixed the issues previously captured

PR #1439, opened 2023-08-12
Standardize capture group IP_ADDRESS

mjbear:
Standardization/normalization instances of ADDRESS
Breaking change on account of the capture group changes

- ADDRESS > MAC_ADDRESS plus make the MAC_ADDRESS regex more concise for paloalto_panos show_inteface_hardware
- replace some instances of ADDRESS with IP_ADDRESS
- replace some instances of ADDRESS with MAC_ADDRESS
- split ADDRESS into its parts IP_ADDRESS and PREFIX_LENGTH (ex: arista show ip ospf interface)
- split IP_ADDRESS into its parts IP_ADDRESS and PREFIX_LENGTH
- regenerate yaml for fortinet get_system_interface and get_system_interface_physical
- replace IP with IP_ADDRESS for most replace IP with IP_STATE for alcatel_aos_show_vlan
- split IP into its parts IP_ADDRESS and PREFIX_LENGTH for aruba_aoscx_show_ip_route_all-vrfs
- replace capture group IPADDRESS with IP_ADDRESS
- replace capture group IP_ADDR with IP_ADDRESS
- replace capture group IPADDR with IP_ADDRESS

PR #1444, opened 2023-08-15
fix PR 1439 - incorrectly modifying ADDRESS to be IP_ADDRESS instead of MAC ADDRESS

mjbear:
fix PR #1439 - incorrectly modifying ADDRESS to be IP_ADDRESS instead of MAC ADDRESS as it should have been in cisco_nxos_show_interface.textfsm

Breaking change fixing my previous breaking change on account of capture group name changes


mjbear:
Huge thanks to @dainok for catching my mistake in PR #1439 regarding the cisco_nxos_show_interface.textfsm template.

itdependsnetworks (approved):

PR #1451, opened 2023-08-18
Add support for Arista EOS show ip bgp detail

tom0010:
Tested versions:

- 4.20.7M
- 4.23.3M
- 4.23.6M
- 4.26.3M
- 4.29.3M


ntc_templates/templates/arista_eos_show_ip_bgp_detail.textfsm
```diff
@@ -0,0 +1,42 @@
+Value LOCAL_PREF (\d+)
+Value WEIGHT (\S+)
+Value ORIGIN (.+)
+Value Filldown NETWORK (\S+)
```
jvanderaa, line 4: @tom0010 can we break out network into PREFIX and PREFIX_LENGTH? This is an updated data model that I need to add a PR shortly to add in.

jvanderaa (changes requested):

ntc_templates/templates/arista_eos_show_ip_bgp_detail.textfsm
tom0010: > 

@jvanderaa How can we do this in the case of a Filldown? What if the prefix is matched but not the subnet and vice versa? This may give us the wrong values.

jvanderaa: @jmcgill298 do you have any ideas on how to do a filldown where we want to capture two things?

jmcgill298: https://github.com/tom0010/ntc-templates/pull/1

tom0010: Thanks @jmcgill298.
How does it stop the filldown? I can see in the tests it does do that, and I've merged. Just wondering how for future.

jvanderaa (approved):

PR #1456, opened 2023-08-30
Cisco ios 5 new templates

PavloSkliarenko:
Added 5 new CiscoIOS templates for commands:

`show ip ospf neighbors detail`
`show mpls l2transport vc`
`show ip dhcp binding`
`show users [[all]]`
`show arp`


PavloSkliarenko:
Made separate PR for each template

PR #1457, opened 2023-08-30
Routeros 16 new templates

PavloSkliarenko:
Added 16 new templates for commands:

/routing ospf neighbor print terse without-paging
/routing bgp peer print status without-paging
/interface print terse without-paging
/log print detail without-paging
/ip address export verbose
/interface ethernet print
/ip neighbor print detail
/system resource print
/system identity print
/system clock print
/user active print
/ip address print
/tool speed-test
/tool profile
/user print
/ping


PavloSkliarenko:
Redundant files in commit

PR #1458, opened 2023-08-30
Routeros 10 new templates

PavloSkliarenko:
Added 10 new templates for commands:

`ip dhcp-server lease print terse without-paging`
`ip route print terse without-paging`
`ip arp print terse without-paging`
`routing bgp peer print status`
`routing ospf neighbor print`
`ip dhcp-server lease print`
`interface print detail`
`interface print brief`
`ip route print detail`
`ip arp print`


PavloSkliarenko:
Made separate PR wor each template

PR #1472, opened 2023-09-07
fortinet_get_router_info_ospf_status template + test

PavloSkliarenko:
Added `fortinet_get_router_info_ospf_status` template + test


ntc_templates/templates/fortinet_get_router_info_ospf_status.textfsm
```diff
@@ -0,0 +1,41 @@
+Value ROUTING_PROCESS_NAME (.*)
```
jvanderaa, line 1:
```suggestion
Value PROCESS_NAME (.*)
```

ntc_templates/templates/fortinet_get_router_info_ospf_status.textfsm
```diff
@@ -0,0 +1,41 @@
...
+Value ROUTING_PROCESS_ID (\S+)
```
jvanderaa, line 2:
```suggestion
Value PROCESS_ID (\S+)
```

jvanderaa (dismissed):
Reviewing the Arista and Cisco `show ip ospf database` we have the keys of `PROCESS_ID` and `ROUTER_ID` in there. I think that it would make sense to add ROUTING to this, but in the interim until the others are changed would like to keep things as similar as possible.

jmcgill298 (approved):

PR #1501, opened 2023-09-10
MikroTik RouterOS 'interface ethernet monitor name once' extend templatewith new values

PavloSkliarenko:
Template  + tests


PavloSkliarenko:
@jvanderaa Take a look at this fix, please. I`ve added new values

ntc_templates/templates/mikrotik_routeros_interface_ethernet_monitor_name_once.textfsm
```diff
@@ -1,23 +1,74 @@
 Value NAME (\S+)
-Value STATUS (\S+)
-Value AUTO_NEGOTIATION (\S+)
-Value RATE (\S+)
-Value FULL_DUPLEX (\S+)
-Value TX_FLOW_CONTROL (\S+)
-Value RX_FLOW_CONTROL (\S+)
```
jmcgill298, line 7: @PavloSkliarenko why are these being changed?

mjbear: @PavloSkliarenko 
Why are several of the capture group regexes being changed?

Rigid regexes result in breakage over time.

For example if Mikrotik decided to change the text for `RX_FLOW_CONTROL` from lowercase `yes` to `Yes` or `YES` this would break the entire template. In this case `\S+` is likely the ideal "long term" solution.

Even `SFP_LINK_LENGTH_9UM` need not be `\S+m` and could instead be simply `\S+`.

Only in rare cases where a regex is too greedy and matches too many characters do we make them more rigid. This is done as a last ditch effort as the "rigid" changes do tend to lead to template breakage over the evolution of a vendor's product/software.

PavloSkliarenko: Ok, fixed that

jmcgill298 (approved):

PR #1520, opened 2023-09-11
CiscoS300 'show vlan' template

PavloSkliarenko:
Template + test


PavloSkliarenko:
Fix value names according to new Data model

jvanderaa (approved):

PR #1523, opened 2023-09-11
CiscoS300 'show ip interface' template

PavloSkliarenko:
Template + test


PavloSkliarenko:
Fix value names according to new Data model

PR #1528, opened 2023-09-11
HP Comware 'display interface' template fix

PavloSkliarenko:
Add untagged vlan parsing

### DIFF
--- a/ntc_templates/templates/hp_comware_display_interface.textfsm
+++ b/ntc_templates/templates/hp_comware_display_interface.textfsm
@@ -10,7 +10,7 @@ Value BANDWIDTH (.*)
 Value DUPLEX (\S+)
 Value SPEED (\S+)
 Value PORT_LINK_TYPE (\S+)
-Value UNTAGGED_VLAN (\d+)
+Value UNTAGGED_VLAN_ID (\d+)
 Value VLAN_NATIVE (\d+)
 Value List VLAN_PASSING ([^,]+)
 Value List VLAN_PERMITTED ([^,]+)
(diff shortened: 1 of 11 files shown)


ntc_templates/templates/hp_comware_display_interface.textfsm
```diff
@@ -10,6 +10,7 @@ Value BANDWIDTH (.*)
 Value DUPLEX (\S+)
 Value SPEED (\S+)
 Value PORT_LINK_TYPE (\S+)
+Value UNTAGGED_VLAN (\d+)
```
jvanderaa, line 13: Let's update this one to being either VLAN_NAME or VLAN_ID to match other VLAN output. Also, can we get some example output that has the untagged wording in it?

jvanderaa (changes requested):

ntc_templates/templates/hp_comware_display_interface.textfsm
PavloSkliarenko: I've changed `UNTAGGED_VLAN` to `UNTAGGED_VLAN_ID`, but I'm not sure, that I can delete preffix `UNTAGGED_`. As I can see in examples [1](https://github.com/networktocode/ntc-templates/blob/d1c1a09a1ca4f4577ac4aab98ec4333892ab71ab/tests/hp_comware/display_interface/hp_comware_display_interface.raw), [2](https://github.com/networktocode/ntc-templates/blob/d1c1a09a1ca4f4577ac4aab98ec4333892ab71ab/tests/hp_comware/display_interface/hp_comware_display_interface2.raw), [3](https://github.com/networktocode/ntc-templates/blob/d1c1a09a1ca4f4577ac4aab98ec4333892ab71ab/tests/hp_comware/display_interface/hp_comware_display_interface6.raw) section with `Untagged VLAN ID :` appears when port in Access mode. But it's also has `Tagged   VLAN ID :` row then (And I don't know what for and why)

When port in Trunk mode. There is no such section at all. Passing and permited VLANs are in `VLAN Passing:` (`VLAN_PASSING` value) and `VLAN permitted:` (`VLAN_PERMITTED` value) rows as we can see [here](https://github.com/networktocode/ntc-templates/blob/d1c1a09a1ca4f4577ac4aab98ec4333892ab71ab/tests/hp_comware/display_interface/hp_comware_display_interface1.raw) for example

Unfortunately I'm unable to supply additional tests. So what you think if I'm leave `UNTAGGED_VLAN_ID` value. And should i rename `VLAN_PASSING` -> `VLAN_ID_PASSING` and `VLAN_PERMITTED` -> `VLAN_ID_PERMITTED` ?

jvanderaa (approved):

PR #1531, opened 2023-09-20
Aruba os show ap bss table details

afly007:
Adding additional version parsing to this template, the original template is based on ArubaOS version 8.9.0.0 or later as shown here: [show ap bss-table](https://www.arubanetworks.com/techdocs/CLI-Bank/Content/aos8/sh-ap-bss-table.htm).

This adds the ability to parse output from versions prior to 8.9.0.0.


jmcgill298:
Thanks @afly007 I have included updates to the existing test case in https://github.com/networktocode/ntc-templates/pull/1777

PR #1552, opened 2023-10-20
Added a template for cisco_ftd_show_vpn-sessiondb_anyconnect

manofcolombia:
A template specifically for cisco_ftd_show_vpn-sessiondb_anyconnect did not exist prior. I attempted to use cisco_asa_show_vpn-sessiondb_anyconnect since they should be relatively similar, however, FTD's output adds an additional field that is not present in the ASA versions I was able to test against. This field is ` ${TUNNEL_ZONE}`. 

I found it more appropriate to create a new template for FTD rather than account for this in the existing ASA template.


jvanderaa (approved):

jvanderaa:
Thanks @manofcolombia

PR #1561, opened 2023-11-10
fix(cisco_xr): Updating bgp vrf all ipv4 unicast summary template to add local as number

elavaud:
Local AS number was missing in Cisco XR on the "show bgp vrf all ipv4 unicast summary" template


jvanderaa (approved):

jvanderaa:
Thank you @elavaud

PR #1565, opened 2023-11-17
Update cisco_asa_ping.textfsm

jorlandobr:
Added U and Q response to RESPONSE STREAM

Cisco has four valid outputs for ping: . (fail) ! (success) U (unreachable) Q (quench)

U and Q were missing. Testing an output that carried these values would generate an error.


jvanderaa:
Can you add additional test cases?

jorlandobr:
Added test cases (yml and raw fles) in jorlandobr-patch-1

jvanderaa:
Looks great @jorlandobr! Thank you for the contribution.

jvanderaa:
I hope to have a new release out next week that will incorporate this.

PR #1564, opened 2023-11-17
ping template for Cisco IOS XR devices

jorlandobr:
Added a ping template for Cisco IOS XR devices. These devices begin the ping output with an extra line with a timestamp. That line would fail if tested against the cisco_ios_ping.textfsm template

PR #1566, opened 2023-11-17
Update index

jorlandobr:
Added index line for new template: cisco_xr_ping.textfsm


jvanderaa:
I'm not seeing a new template here.

jorlandobr:
My mistake. It should be added on PR #1564.

PR #1567, opened 2023-11-21
fix doc build

whitej6:


jvanderaa:
@whitej6 just a bump here.

whitej6:
@jvanderaa fixed, likely the pylint version in poetry.lock was updated when I relocked and a new test was added?

PR #1576, opened 2023-12-12
Rename extreme platform to extreme_exos

PavloSkliarenko:


jvanderaa:
@ktbyers we have been trying to keep alignment with Netmiko on the driver names with this. I think this makes sense but wouldn't mind a quick "Yep, this aligns" type thing.

ktbyers:
Yes, that works/is correct from a Netmiko perspective.

https://github.com/ktbyers/netmiko/blob/develop/netmiko/ssh_dispatcher.py#L215

PR #1581, opened 2023-12-23
Fixes parsing of the `ip` commands on busybox systems.

abates:
Embedded linux systems running busybox don't necessarily include the `iproute2` package, but instead leverage similar functionality from busybox. Unfortunately, the output from busybox `ip` commands appears to be a subset of the `iproute2` ip commands. This commit marks some fields as optional and also allows trailing whitespace (which busybox introduces) in order to parse `ip link show` and `ip address show` commands on busybox systems.


jvanderaa (approved):

jvanderaa:
LGTM.

PR #1588, opened 2024-01-10
MikroTik RouterOS '/interface print terse without-paging' fix

PavloSkliarenko:
Add missing field 'fast-path' for Mikrotik's command '/interface print terse without-paging'


jvanderaa (approved):
LGTM

PR #1603, opened 2024-01-18
enhance show inventory nxos template

jeffkala:
Enhance show inventory to be used for virtual N9K applicances


jvanderaa (approved):

jvanderaa:
Thank you @jeffkala

Issue #1317, opened 2023-03-10
cisco_nxos_show_interfaces_switchport adds additional space to monitor statement


Mattias3891:
Of unknown reason, I get an additional space in the "switchport_monitor" string. 
Thus, having to match "Enabled " and not just Enabled for example.

Example output:
`{'interface': 'Ethernet1/35', 'switchport': 'Enabled', 'switchport_monitor': 'Enabled ', 'mode': 'access', 'access_vlan': '1', 'access_vlan_name': 'default', 'native_vlan': '1', 'native_vlan_name': 'default', 'trunking_vlans': '1-4094', 'voice_vlan': ''}
{'interface': 'Ethernet1/36', 'switchport': 'Enabled', 'switchport_monitor': 'Enabled ', 'mode': 'access', 'access_vlan': '1', 'access_vlan_name': 'default', 'native_vlan': '1', 'native_vlan_name': 'default', 'trunking_vlans': '1-4094', 'voice_vlan': ''}
{'interface': 'port-channel22', 'switchport': 'Enabled', 'switchport_monitor': 'Not enabled ', 'mode': 'trunk', 'access_vlan': '1', 'access_vlan_name': 'default', 'native_vlan': '1', 'native_vlan_name': 'default', 'trunking_vlans': '10-20', 'voice_vlan': ''}
{'interface': 'port-channel23', 'switchport': 'Enabled', 'switchport_monitor': 'Not enabled ', 'mode': 'trunk', 'access_vlan': '1', 'access_vlan_name': 'default', 'native_vlan': '1', 'native_vlan_name': 'default', 'trunking_vlans': '10-20', 'voice_vlan': ''}
{'interface': 'port-channel24', 'switchport': 'Enabled', 'switchport_monitor': 'Not enabled ', 'mode': 'trunk', 'access_vlan': '1', 'access_vlan_name': 'default', 'native_vlan': '1', 'native_vlan_name': 'default', 'trunking_vlans': '10-20', 'voice_vlan': ''}`

To fix this, I've corrected the template to only allow Enabled/Disabled on switchport statement and Enabled/Not enabled on the switchport_monitor statement.

Value Required INTERFACE (\S+)
Value SWITCHPORT (Enabled|Disabled)
Value SWITCHPORT_MONITOR (Enabled|Not\senabled)
Value MODE (access|trunk)

Works well.

mjbear:
Hello @Mattias3891 

[...]

PR #1622, opened 2024-02-27
bug fix - modify 'switchport monitor' regex so a trailing space is not captured

mjbear:
bug fix - modify 'switchport monitor' regex so a trailing space is not captured to resolve issue #1317


jvanderaa (approved):
LGTM

PR #1631, opened 2024-03-08
New 'show vlans' template for cisco_ios

chadell:


jvanderaa (approved):

jmcgill298 (changes requested):
This will need to be updated to follow our standard for matching all lines and including an `-> Error` statement

chadell:
@jvanderaa I have a problem with the `test_index_ordering`.
As the index is in my PR, the parsing works, but if I swap the with the `show vlan` then the `show vlans` command uses the `show vlan` one, I guess because it's matching the string.

jmcgill298 (approved):

PR #1649, opened 2024-03-13
huawei_smartax_display_ont_info: New

evilmonkey19:
This is ready for code review :)


evilmonkey19:
I have changed the index ordering because with the last pull requests into the master the test didn't pass. Now they pass :)

jvanderaa:
Thank you @evilmonkey19 that is what I was thinking was the case. I'm having to do things a bit by hand on each of these, but I'd still take them in this route rather than the one big PR!! :D

evilmonkey19:
It is my pleasure!! I don't mind do a little bit extra effort if then it gets easier for everyone (even for myself). I'm learning how to do better PR thanks to you ;). Thanks!! :D

PR #1601, opened 2024-01-17
New template: cisco_xr_show_ip_interface

lamiskin:
(cherry picked from commit 5e28563941ff957163ae479ce09c2a8da31d4f50)

PR #1661, opened 2024-03-14
cisco_xr_show_ipv4_interface: Add new attributes, update index

lamiskin:
Update the index so "v4" as part of the command is optional.
"show ip interface" returns the same as "show ipv4 interface".

Add PREFIX_LENGTH, IP_MTU, MULTICAST_GROUPS.

Including relevant tests.

Related to #1601


lamiskin:
This would be a breaking change, as the IP_ADDRESS no longer includes the prefix length.

jvanderaa (approved):

PR #1662, opened 2024-03-14
cisco_show_ip_route: Remove vrf regex in index file.

lamiskin:
The current line in the index file is stopping some usage of the command, such as:
- show ip route static
- show ip route 0.0.0.0/0 

Basically anything other than specifying a vrf.

This PR would now make it the same as other platforms (like nx-os).


jvanderaa:
@itdependsnetworks would you agree with me here that this would be a breaking change? cc: @mjbear

mjbear:
@jvanderaa @itdependsnetworks @lamiskin 

So far based on my testing commands continue to parse. With no changes in the template itself, I don't believe this will "functionally" break anything.

We're fortunate that over the years IOS code for the specific routes (ex: `show ip route static`) prints the headers and gateway line so the state transition to `Routes` happens in the template. (Really old code didn't do that.)

Listing a specific route (`show ip route 0.0.0.0/0`) may not print headers+gateway though I should verify this against newer IOS.

The following commands parse fine with the index regex set to `sh[[ow]] ip r[[oute]]`:
`show ip route`
`show ip route vrf NAME`
`show ip route static`
`show ip route ospf`
`show ip route ospf PROC_ID`

jvanderaa (approved):

jvanderaa:
I'm going with this, knowing that we have expanded support and use that works.

lamiskin:
@mjbear @jvanderaa Thank you!

PR #1665, opened 2024-03-20
cisco_ios_show_interfaces_switchport: Edge case addition

lamiskin:
Had some failures because of the line 'Priority for untagged frame: 0'.

Fixed and added an appropriate test.


jvanderaa (approved):
LGTM

PR #1632, opened 2024-03-11
Add allied telesis templates

dainok:
The PR adds some templates for Allied Telesis AW+ devices:

- allied_telesis_awplus_show_arp.textfsm
- allied_telesis_awplus_show_etherchannel_summary.textfsm
- allied_telesis_awplus_show_interface.textfsm
- allied_telesis_awplus_show_interface_switchport.textfsm
- allied_telesis_awplus_show_ip_route.textfsm
- allied_telesis_awplus_show_lldp_neighbors_detail.textfsm
- allied_telesis_awplus_show_mac_address-table.textfsm
- allied_telesis_awplus_show_static-channel-group.textfsm
- allied_telesis_awplus_show_system.textfsm
- allied_telesis_awplus_show_vlan_all.textfsm

PR #1676, opened 2024-03-22
Add Allied Telesis show_system template

dainok:
Add Allied Telesis show_system template

Replace https://github.com/networktocode/ntc-templates/pull/1632


dainok:
Required by: https://github.com/dainok/netdoc/pull/122/

Issue #1179, opened 2022-08-11
cisco_ios_show_mac-address-table.textfsm error in cisco C6807


williamluotao:
##### ISSUE TYPE

 - Bug Report

##### TEMPLATE USING

```
Value DESTINATION_ADDRESS ([0-9a-fA-F]{4}\.[0-9a-fA-F]{4}\.[0-9a-fA-F]{4})
Value TYPE (\S+)
Value VLAN (\S+)
Value List DESTINATION_PORT ([^,\s]+)

Start
  ^Destination\s+Address\s+Address\s+Type\s+VLAN\s+Destination\s+Port -> TYPE1
  ^\s+vlan\s+mac address\s+type\s+learn\s+age\s+ports -> TYPE2
  ^\s+vlan\s+mac address\s+type\s+protocols\s+port -> TYPE3
  ^Vlan\s+Mac Address\s+Type\s+Ports -> TYPE4
  # Capture time-stamp if vty line has command time-stamping turned on
  ^Load\s+for\s+
  ^Time\s+source\s+is

TYPE1
  ^${DESTINATION_ADDRESS}\s+${TYPE}\s+${VLAN}\s+${DESTINATION_PORT}$$ -> Record
  ^-+\s+-+
  ^Displaying\s+entries
  ^Destination\s+Address\s+Address\s+Type\s+VLAN\s+Destination\s+Port
  ^\s*$$
  ^Multicast\s+Entries -> End
  ^. -> Error

TYPE2
  # Order of the group in brackets here matters
  ^\s*(\*\s+R|\*|R|\s)\s*(\d+|-+|[Nn]/[Aa]) -> Continue.Record
  # using 20 spaces should ensure that lines only match destination port flowing to next line
  ^\s{20}\s+${DESTINATION_PORT}(\s+|,\s*)\S -> Continue
  ^\s{20}\s+${DESTINATION_PORT},*\s*$$
  ^\s{20}\s+[^,\s]+(\s+|,\s*)${DESTINATION_PORT}(\s+|,\s*)\S -> Continue
  ^\s{20}\s+[^,\s]+(\s+|,\s*)${DESTINATION_PORT},*\s*$$
  ^\s{20}\s+([^,\s]+(\s+|,\s*)){2}${DESTINATION_PORT}(\s+|,\s*)\S -> Continue
  ^\s{20}\s+([^,\s]+(\s+|,\s*)){2}${DESTINATION_PORT},*\s*$$
[...]

PR #1688, opened 2024-04-12
ios show mac addr: fix match of legend

jmcgill298:
This comes from @jacdavi where show mac address table was only matching the "R" field, but the legend shows "S" and "D" are also options. This replaces #943 which is out of sync with the default branch.


jmcgill298:
This replaces #943

mjbear:
resolved #1179

mjbear:
resolves #1179

PR #1689, opened 2024-04-12
Add ability to specify template dir as argument

jmcgill298:
This allows specifying the template directory to the `parse` function directly instead of having to use an environment variable. In addition to looking in the custom template directory, this allows a failure to fallback to the default directory. This comes from @jacdavi in #945


jmcgill298:
This is a replacement of #945 that is affected be history rewrite

jvanderaa (approved):

PR #1701, opened 2024-05-24
Update cisco_asa_show_running-config_object_network.textfsm

wollmannbruno:
This update includes the case where an object of type network includes an FQDN entry.

It was tested and worked on a Cisco ASA 5545 running version 9.12(4)67


wollmannbruno:
I found a better way to do things and will submit a pull request once finalized.

PR #1711, opened 2024-06-06
Update template: huawei_smartax display_ont_info_fsp to display_ont_info_0_0_0

evilmonkey19:
Currently, there is a little problem with the template. In the Huawei SmartAX models instead of using the 0/0/0 for example, the syntax is `display ont info 0 0 0`. I have changed so it matches the exact command as the one you would send via Netmiko. Also I have updated the template because if you have an output like the following it would have crashed:
```
  -----------------------------------------------------------------------------
  F/S/P   ONT         SN         Control     Run      Config   Match    Protect
          ID                     flag        state    state    state    side
  -----------------------------------------------------------------------------
    0/ 0/0    0  NEBW3W2ABK434A01  active      -        -        -        yes
    0/ 0/0    1  NFNCFTKLQNERTWPT  active      -        -        -        yes
  -----------------------------------------------------------------------------
  F/S/P   ONT-ID   Description
  -----------------------------------------------------------------------------
  0/ 0/0       0   The equable treatment pushes.
  0/ 0/0       1   The vacuous statistic attaches.
  -----------------------------------------------------------------------------
  In port 0/ 0/0, the total number of ONTs are: 2, online: 0, offline: 0
  -----------------------------------------------------------------------------
  ```
  
  Also an extra test case has been added.

### CODE REVIEW

**Comment by itdependsnetworks** [comment]
is it possible it would be something like `display ont info 1 2 0` ?

**Comment by evilmonkey19** [comment]
Yees, you can running using something like `display ont info 0 0 1` or `display ont info 1 2 0`. The 0 0 0 it's just a name. In the index I use the (\d+\s\d+\s\d+) to catch the  1 2 0.

**Comment by evilmonkey19** [comment]
Hey! How is it going the review?

File: tests/huawei_smartax/display_ont_info_0_0_0/huawei_smartax_display_ont_info_fsp_3.raw [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+  0/ 0/0      20   The skinny counter-force pecks.
+  0/ 0/0      21   The aloof bacon throws.
+  0/ 0/0      22   The inquisitive robotics wends.
+  0/ 0/0      23   The wee comma uses.
+  0/ 0/0      24   The anxious peacock heals.
+  0/ 0/0      25   The youthful quince models.
+  0/ 0/0      26   The talented interest flows.
+  -----------------------------------------------------------------------------
+  In port 0/ 0/0, the total number of ONTs are: 27, online: 9, offline: 10
+  -----------------------------------------------------------------------------
```
jmcgill298, line 64:
```suggestion
  -----------------------------------------------------------------------------

```

ntc_templates/templates/index
```diff
@@ -587,7 +587,7 @@ hp_procurve_show_arp.textfsm, .*, hp_procurve, sh[[ow]] ar[[p]]
 hp_procurve_show_ip.textfsm, .*, hp_procurve, sh[[ow]] ip
 
 huawei_smartax_display_ont_info_summary_ont.textfsm:huawei_smartax_display_ont_info_summary_sn.textfsm, .*, huawei_smartax, di[[splay]] ont i[[nfo]] su[[mmary]] \S+ *$
-huawei_smartax_display_ont_info_fsp.textfsm:huawei_smartax_display_ont_info_description.textfsm, .*, huawei_smartax, di[[splay]] ont i[[nfo]] \S+ *$
+huawei_smartax_display_ont_info_0_0_0.textfsm:huawei_smartax_display_ont_info_description.textfsm, .*, huawei_smartax, di[[splay]] ont i[[nfo]] (\d+\s\d+\s\d+) *$
```
jmcgill298, line 590:
```suggestion
huawei_smartax_display_ont_info_0_0_0.textfsm:huawei_smartax_display_ont_info_description.textfsm, .*, huawei_smartax, di[[splay]] ont i[[nfo]] (\d+\s\d+\s\d+)\s*$
```

jmcgill298 (changes requested):

jmcgill298 (approved):

jmcgill298:
@evilmonkey19 is this different than https://github.com/networktocode/ntc-templates/blob/master/ntc_templates/templates/huawei_smartax_display_ont_info_0_1_2.textfsm ?

If so, something needs to change with the regex in the index file

PR #1714, opened 2024-06-12
New template: huawei_smartax_ont_add

evilmonkey19:
This template is used to check the ont_id whenever a new ont is registered to a port.


evilmonkey19:
I'm sorry for the force-pushed. I was working in this branch on another template without realizing. I moved to another branch (as i should)

jvanderaa (approved):

PR #1718, opened 2024-06-19
New template: cisco_nxos_show_ip_eigrp_neighbors

ryanmerolle:
match cisco IOS command and add the vrf given its in the nxos show command output.


jvanderaa:
@ryanmerolle would we be able to add the error line that we see on many in case the output changes that would break this?

ryanmerolle:
> @ryanmerolle would we be able to add the error line that we see on many in case the output changes that would break this?

Of course.  I should have caught that.  I'll update when back at a computer.

ryanmerolle:
@jvanderaa fixed

jvanderaa (approved):

Issue #1182, opened 2022-08-18
truncated time in cisco_nxos_show_ip_bgp_neighbors.textfsm


ajw1980:
For values less than 1 day this regex is incorrect for cisco_nxos_show_ip_bgp_neighbors.textfsm:
```
Value UPTIME (\w+)
Value LAST_RESET (\w+)
Value LAST_PEER_RESET (\w+)
```

It is HH:MM:SS for values less than 1 day:
`BGP state = Established, up for 20:03:51`

I think this change is sufficient:
```
Value UPTIME ([\w:]+)
Value LAST_RESET ([\w:]+)
Value LAST_PEER_RESET ([\w:]+)
```

mjbear:
@ajw1980 @itdependsnetworks 

I tinkered around with it starting out with `\S+` first only to find that when a peer is in the idle BGP state then the template picks up an extra comma. Easily remedied by modifying the State Rule to make the comma optional.

By using the pattern for everything but white space `\S+`, the date could change from 20:03:51 to 20-03-51 and it will still parse (yes, silly if that time format changed). My suggestion also requires less characters at the expense of the flexibility just mentioned (pro and con). **Whether this is the best solution I'll leave to the group here to decide.** Either way I'm game to work up and submit the PR for the fix to be included.

```
Value UPTIME (\S+)
Value LAST_RESET (\S+)
Value LAST_PEER_RESET (\S+)
  ^\s+BGP state = ${BGP_STATE}, \w+ for ${UPTIME},?
```

mjbear:
@ryanmerolle
I'd say it just comes down to somebody (I'll volunteer) pulling a PR together.

I had been chasing some of these and when there was no response I moved on to other priorities. :shrug: That's life.
Thank you

mjbear:
@ryanmerolle @jvanderaa 
I have a PR ... it's #1719 to fix capturing BGP uptimes less than a day

ryanmerolle:
@jvanderaa can we please close this issue with PR is closed.

PR #1719, opened 2024-06-21
fix issue#1182 - capture BGP times less than a day

mjbear:
Solve Issue #1182 with a fix for capturing BGP uptimes that are less than a day


jvanderaa (approved):

mjbear:
resolves #1182 
test to see if this auto-resolves the issue ticket :grin:

PR #1722, opened 2024-06-22
Junos show_chassis_hardware simplify regex

mjbear:
This is a follow on to PR #1721 (this is based against my branch for 1721)

There were several regexes of the same `\w+-\w+|\w+` pattern in junos show_chassis_hardware that could be simplified to `\S+` without causing parsing issues.


mjbear:
@collin-wicker Pinging you to get your input. It parses the same as before! :grinning:
Thank you!

jvanderaa (approved):

PR #1737, opened 2024-07-13
Enhance IOS sho ip route to support *

jmcgill298:
Also support capturing Route Flags


jmcgill298:
Replaces https://github.com/networktocode/ntc-templates/pull/1131

jvanderaa (approved):

PR #1742, opened 2024-07-15
resolves 1741 (bugfix) - ios show interface link - NAME should not match a space if there is no other text

mjbear:
resolves 1741 (bugfix)
- ios - show interface link - NAME should not match a space if there is no other text


mjbear:
resolved by #1742

jmcgill298 (approved):

PR #1743, opened 2024-07-15
Updated Huawei SmartAX display ont info files so it can match the cas…

jmcgill298:
…e of single (all information about one) or multiple ONTs (a list with little information).


jvanderaa (approved):

evilmonkey19:
Thanks for the rebase! :)

PR #1164, opened 2022-06-24
added juniper show bgp summary

ysanajong:
##### ISSUE TYPE
 - New Template Pull Request

##### COMPONENT
Juniper junos show bgp summary 

##### SUMMARY
Adds parsing for show bgp summary output on Juniper JunOS

PR #1744, opened 2024-07-15
Adding juniper show bgp summary command

jmcgill298:


jmcgill298:
replaces https://github.com/networktocode/ntc-templates/pull/1164

PR #1173, opened 2022-07-26
Update cisco_xr_show_interfaces.textfsm

Na-Cly:
Add optional value LINK_TYPE for SR, LR . . . etc for physical link type.

##### ISSUE TYPE
 - New Template Pull Request (Updated existing)

##### COMPONENT
cisco_xr_show_interfaces

##### SUMMARY
Add value LINK_TYPE to get SR vs LR etc physical link type

PR #1746, opened 2024-07-16
Cisco xr show interfaces add hardware_media

jmcgill298:


jmcgill298:
replaces https://github.com/networktocode/ntc-templates/pull/1173

jvanderaa (approved):

PR #1342, opened 2023-03-27
Update cisco_nxos_show_interface.textfsm

Ardeck:
adding additional parameters to be retrieved :
- PKT_INPUT_RATE
- PKT_OUTPUT_RATE
- BW_INPUT_RATE
- BW_OUTPUT_RATE

PR #1769, opened 2024-07-19
Add fields to IOS/NXOS show int

jmcgill298:


jmcgill298:
replaces https://github.com/networktocode/ntc-templates/pull/1342

jvanderaa (approved):

PR #1784, opened 2024-07-24
Fix ios dir when no files present

jmcgill298:


mjbear:
Thank you @jmcgill298 !

PR #1785, opened 2024-07-25
cisco nxos show interface description for nexus without mgmt port

Kortzy:
The previous template does not take into account Nexuses that are without management ports.


jmcgill298:
Thanks for the PR @Kortzy , can you please provide a new test case for this also?

jmcgill298 (approved):

jmcgill298:
Thanks @Kortzy

PR #1807, opened 2024-08-08
Fix Mikrotik's '/system resource print' template

PavloSkliarenko:


jmcgill298:
@PavloSkliarenko can you please update or add a test case so we can make sure not to break this in a future change?

PavloSkliarenko:
> @PavloSkliarenko can you please update or add a test case so we can make sure not to break this in a future change?

Unfortunately I've made this fix long time ago and I can't provide additional test cases. Should I close this PR?

jmcgill298:
> > @PavloSkliarenko can you please update or add a test case so we can make sure not to break this in a future change?
> 
> Unfortunately I've made this fix long time ago and I can't provide additional test cases. Should I close this PR?

If you are unable to find or make up an example, then we would need to close the PR. We cannot accept changes to Values without tests.

PR #1814, opened 2024-08-08
Fix template for Edgecore's command 'show interfaces status'

PavloSkliarenko:


ntc_templates/templates/edgecore_show_interfaces_status.textfsm
```diff
@@ -4,7 +4,7 @@ Value ADDRESS (.*)
 Value NAME (.*)
 Value PORT_ADMIN (.*)
 Value MDIX_MODE (.*)
-Value NEGOTIATION (Auto)
+Value NEGOTIATION (.*)
```
jmcgill298, line 7:
```suggestion
Value NEGOTIATION (.*?)
```

jmcgill298 (changes requested):
make sure we aren't capturing trailing whitespace

ntc_templates/templates/edgecore_show_interfaces_status.textfsm
PavloSkliarenko: Done

jmcgill298 (approved):

PR #1841, opened 2024-09-18
fix: cisco `show ip interface` failing if enabled turbo vector

gertzakis:
Addressing issue #1839 for `show ip interface` failing in Cisco IOS

PR #1855, opened 2024-09-24
Fix IOS interface link template uptime issue

paulbauer1:
Corrected issue with template not displaying uptime values correctly when there are hours in the uptime value. Added optional hours to the uptime regex statement.


itdependsnetworks:
Can you add tests as well? 

We follow a simple mock and verify value setup, you can see a recent example here: https://github.com/networktocode/ntc-templates/pull/1841/files and documentation of the process here: https://ntc-templates.readthedocs.io/en/latest/dev/dev_parser/#tests


Essentially we are looking for new output that shows your new condition, while ensuring that there is no regressions to other conditions. This also protects your update in the future, in case someone were to inadvertently revert your change.

paulbauer1:
I added a new test file, tests/cisco_ios/show_interface_link/cisco_ios_show_interface_link4.raw. It is basically the same as the existing first raw file just with a newly generated yaml file to go with it. The existing test file (tests/cisco_ios/show_interface_link/cisco_ios_show_interface_link.raw) already contained the necessary data to validate my changes. Although, now the existing raw and yaml files fail testing as the bug I fixed had created bad data in the yaml file so that now fails. The tests/cisco_ios/show_interface_link/cisco_ios_show_interface_link.yaml file is no longer a valid yaml file now that the bug has been resolved.  Should I close this PR and create a new one in order to get the new test files incorporated and what should be done with the existing test files that now fail? Thank you.

jmcgill298:
@paulbauer1 if the existing test cases are already symptomatic of the problem, then you just need to update the existing test cases with the correct values, and do not need to create a new test case

paulbauer1:
I regenerated the cisco_ios_show_interface_link.yaml file. The other files were fine. All tests passed. 

============================================================================================== 2340 passed in 16.03s ==============================================================================================
All tests have passed!

jmcgill298 (changes requested):
I created a PR back to you to fix the downtime not getting captured https://github.com/paulbauer1/ntc-templates/pull/1

jmcgill298 (approved):

jmcgill298:
thanks @paulbauer1

PR #1862, opened 2024-10-09
New template: Aruba Procurve `show version`

evilmonkey19:
Personally, I don't know the difference between ArubaOS and Aruba Procurve. However, in the device i'm testing against using Netmiko, aruba_procurve works fine and aruba_os doesn't. 😅


ntc_templates/templates/aruba_procurve_show_version.textfsm
```diff
@@ -0,0 +1,18 @@
...
+  ^\s*Aruba\s*Operating\s*System\s*Software
+  ^\s*ArubaOS\s*\(MODEL:\s*${MODEL}\),\s*Version\s*${VERSION}
+  ^\s*Website:\s*http://www.arubanetworks.com
+  ^\s*\(c\)\s*Copyright\s*\d+\s*Hewlett\s*Packard\s*Enterprise\s*Development\s*LP\.
+  ^\s*Compiled\s*on\s*${COMPILATION}
+  ^\s*FIPS\s*Mode\s*:\s*${FIPS_MODE}
+  ^\s*AP\s*uptime\s*is\s*${UPTIME}
+  ^\s*Reboot\s*Time\s*and\s*Cause\s*:\s*${REBOOT_CAUSE}
+  ^\s*$$
+  ^.*$$ -> Error
```
jmcgill298, line 18:
```suggestion
  ^.*$$ -> Error

```

jmcgill298 (commented):
Should this be hp_procurve?

evilmonkey19:
Yes! It is the same 👌
https://github.com/ktbyers/netmiko/blob/eab18842d38e5993f39ca010a970ce6687797d75/netmiko/ssh_dispatcher.py#L172

evilmonkey19:
It is fine to put into `hp_procurve`, but i want to use `aruba_procurve` as the platform. does it work in ntc_templates?

jmcgill298:
> It is fine to put into `hp_procurve`, but i want to use `aruba_procurve` as the platform. does it work in ntc_templates?

That can be done by changing the index file to use `(hp|aruba_procurve)`, but I suspect that might break are testing framework (should be fixable though)

jmcgill298:
@evilmonkey19 can you test https://github.com/evilmonkey19/ntc-templates/pull/2 to see if that works for you?

evilmonkey19:
During the weekend I will try to fix it :)

evilmonkey19:
In local development it works using the `aruba_procurve` platform if you do the little change in the index. Personally, I'm in favor of testing everything but in this case it is just a small change that would test the same result against the same platform. I don't have further testing should be required.

jmcgill298 (approved):

Issue #1863, opened 2024-10-09
Cisco IOS `show bfd neighbors details` failing when missing `last`


gertzakis:
##### ISSUE TYPE

 - Template Issue with error and raw data
##### TEMPLATE USING

```
Value OUR_ADDR (\S+)
Value NEIGHBOR_ADDR (\S+)
Value LOCAL_DISCRIM (\d+)
Value REMOTE_DISCRIM (\d+)
Value REMOTE_HEARD (\S+)
Value HOLDDOWN (\d+)
Value MULTIPLIER (\d{1,2})
Value STATE (\w+)
Value INTERFACE (\S+)
Value ECHO_FUNCTION ((?:\w+\s+)?\w+\s+echo\s+function)
Value ECHO_INTERVAL (\d+)
Value DIAGNOSTIC_BIT (\d)
Value DEMAND_BIT (\d)
Value POLL_BIT (\d)
Value MIN_TX_INTERVAL (\d+)
Value MIN_RX_INTERVAL (\d+)
Value RECEIVED_MIN_RX_INTERVAL (\d+)
Value RECEIVED_MULTIPLIER (\d{1,2})
Value HOLDDOWN_HITS (\d+)
Value HELLO_INTERVAL (\d+)
Value HELLO_HITS (\d+)
Value RX_COUNT (\d+)
Value RX_LAST (\d+)
Value TX_COUNT (\d+)
Value TX_LAST (\d+)
Value REGISTERED_PROTOCOLS (\S+(?:\s+\S+)*)
Value TEMPLATE (\S+)
Value UPTIME (\S+)
Value VERSION (\d+)

Start
  ^\s*IPv\d+\s+Sessions\s*$$
  ^\s*(?:\S+\s+)?NeighAddr.*$$ -> BFD
  #
  # Capture time-stamp if vty line has command time-stamping turned on
  ^Load\s+for\s+
  ^Time\s+source\s+is
  ^\s*$$
  ^. -> Error

BFD
  ^\s*${OUR_ADDR}\s*${NEIGHBOR_ADDR}\s+${LOCAL_DISCRIM}\/${REMOTE_DISCRIM}\s+${REMOTE_HEARD}\s+\d+\s+\(\s*\d+\s*\)\s+${STATE}\s+${INTERFACE}\s*$$
  #
[...]

PR #1864, opened 2024-10-09
fix: IOS 'show bfd neighbors details' failures

gertzakis:
Addresses issue #1863 regarding BFD neighbors details command output


jmcgill298 (approved):

jmcgill298:
@gertzakis if you have a fix, then you don't need to create an issue

PR #1894, opened 2024-11-08
fix: Corrected Cisco IOS show_access-list template

k-ribot:
Corrected source port returned with matches value.


ntc_templates/templates/cisco_ios_show_access-list.textfsm
```diff
@@ -24,8 +24,8 @@ Value OPERATOR_DESTINATION_PORT (eq|gt|lt|neq|range)
 #
 # This can be either a single port or a list.
 #
-Value SOURCE_PORT ([\s\S]+)
-Value DESTINATION_PORT ([\s\S]+)
+Value SOURCE_PORT (((?! \(\d+ matches\))[\s\S])+)
+Value DESTINATION_PORT (((?! \(\d+ matches\))[\s\S])+)
```
jmcgill298, line 28: These additions should be moved to the match lines in the State section

k-ribot: Hello, I tried another solution

jmcgill298 (approved):

PR #1908, opened 2024-11-22
Added template for Alcatel AOS: `show interfaces port`

evilmonkey19:

PR #1911, opened 2024-11-26
fix: Bring port_id inline with other templates (allow satelites for interface)

bennnnnnnn:
Make PORT_ID less restrictive and bring it in line with other templates


ntc_templates/templates/alcatel_sros_show_lag_port.textfsm
```diff
@@ -1,5 +1,5 @@
 Value LAG_ID (\d+\(\w\))
-Value PORT_ID (\d+/\d+/\d+)
+Value PORT_ID (\S+)
```
mjbear, line 2: In PR #1908, (another Alcatel platform) @evilmonkey19 and I put together the following somewhat specific, yet flexible pattern.

Worth a thought.
If you like this suggestion, click the Commit suggestion button to accept my edit. Thank you!

```suggestion
Value PORT_ID ((\d+\/?)+)
```

mjbear: I don't show any `A/1` in the raw test output as there had been for #1912 
Hmmm

evilmonkey19: I recommend using the `((\d+\/?)+)` as @mjbear suggests as it is more intuitive. It only works if you use `1/1` but `A/1` which would be rare as far as I know.

mjbear: It's also possible to update the regex to allow `A/1`.
I'll give it some thought and testing, hehe.

bennnnnnnn: Bear in mind we also need to support satellite ports like esat-1/1/1

mjbear: > Bear in mind we also need to support satellite ports like esat-1/1/1

Ok. Cool.

@bennnnnnnn
We [need test data](https://github.com/networktocode/ntc-templates/blob/master/docs/user/faq.md#i-simply-want-to-add-my-template-to-the-project-i-do-not-want-to-add-all-of-these-tests-and-controls-can-i-just-do-so) for this PR. This PR shows only a single file (the template).

Please provide raw cli output that contains:

* Normal ports (ex: `1/1/1`)
* Ports for slot A (ex: `A/1`)
* Satellite ports (ex: `esat-1/1/1`)
* :point_right: ***Are there any other types of ports we should have present in test data?***
  * Ideally if we can find output from a device that has all of these that would be great! Thank you!

Could you please add test data that includes those `esat` ports?
I'll modify my suggestion a while, but chances are it could need modified when more test data shows up. :wink::grin:

jmcgill298 (changes requested):
Thanks @bennnnnnnn we will need test data before we are able to proceed with this PR.

mjbear:
I just noticed a couple of lines that should be fixed up.
Since I can't edit those lines from the Review feature, I'm going to submit a PR against your branch and fork! :grin:

@bennnnnnnn
:dart: Please review bennnnnnnn/ntc-templates/pull/1 and accept before Jacob re-reviews this.
Thank you!

jmcgill298 (approved):

PR #1920, opened 2024-12-02
Juniper junos show system processes summary

jnicholson56:
Adding Juniper Junos command **show system processes summary** support


jnicholson56:
I am trying to send PR's per commit. This did not work as intended.

PR #1936, opened 2024-12-13
Adding template for Alcatel AOS (only R8): `show interfaces alias`

evilmonkey19:
It only works in R8 not in R6


ntc_templates/templates/index
```diff
@@ -11,9 +11,11 @@
 #
 Template, Hostname, Platform, Command
 
+
```
jmcgill298, line 14:
```suggestion
```

jmcgill298 (approved):

PR #1942, opened 2024-12-19
Updated Aruba AOSCX show_vsf_detail with additional values

ritkit:
I needed the Topology type, hostname, and serial numbers from members in the stacks. I also fixed the status value as it would only get the first word and there are statuses with multiple works. Such as "No Link".


ntc_templates/templates/aruba_aoscx_show_vsf_detail.textfsm
```diff
@@ -1,18 +1,24 @@
...
+  ^\s+Name\s+:\s+${NAME}
   ^Member.* -> Continue.Record
   ^Member\s+ID\s+:\s+${MEMBER_ID}
-  ^\s+MAC\s+Address\s+:\s+${MAC_ADDRESS}                                                                          
-  ^\s+Status\s+:\s+${STATUS}                                                                       
-  ^\s+CPU\sUtilization\s+:\s+${CPU}                                                                      
+  ^\s+MAC\s+Address\s+:\s+${MAC_ADDRESS}
+  ^\s+Status\s+:\s+${STATUS}
+  ^\s+Serial\sNumber\s+:\s+${SERIAL_NUMBER}
+  ^\s+CPU\sUtilization\s+:\s+${CPU}
```
mjbear, line 21: While there isn't more than one space today, adding the plus sign future proofs us (at the cost of an extra character).

```suggestion
  ^\s+Serial\s+Number\s+:\s+${SERIAL_NUMBER}
  ^\s+CPU\s+Utilization\s+:\s+${CPU}
```

ritkit: Fair, thanks.

mjbear:
> I also fixed the status value as it would only get the first word and there are statuses with multiple works. Such as "No Link".

@ritkit
If you have different output, you could also add a snippet to the test data.

While I don't see "No Such" link, there is "No Split" as a multi-word status already.

ntc_templates/templates/aruba_aoscx_show_vsf_detail.textfsm
```diff
@@ -1,18 +1,24 @@
 Value MEMBER_ID (\S+)
+Value TOPOLOGY (\S+)
+Value NAME (\S+)
 Value MAC_ADDRESS (\S+)
-Value STATUS (.*)
+Value STATUS (\S+.*)
```
mjbear, line 5: Do we need the extra regex characters?

There's already a `No Split` multi-word status.

https://github.com/networktocode/ntc-templates/blob/master/tests/aruba_aoscx/show_vsf_detail/show_vsf_detail.yml#L7

Open to discussion - thank you!

```suggestion
Value STATUS (.*)
```

ritkit: Ya I made this change for a reason. 
I am now trying to remember exactly what that state was. Let me see if I can find the scenario that broke it. I just remember seeing only one word on the output to yml at one point.

mjbear: > Ya I made this change for a reason. I am now trying to remember exactly what that state was. Let me see if I can find the scenario that broke it. I just remember seeing only one word on the output to yml at one point.

Sounds good.
Please share the raw output for that situation when you find it.
Thank you!

ntc_templates/templates/aruba_aoscx_show_vsf_detail.textfsm
```diff
@@ -1,18 +1,24 @@
...
   ^Member.* -> Continue.Record
   ^Member\s+ID\s+:\s+${MEMBER_ID}
-  ^\s+MAC\s+Address\s+:\s+${MAC_ADDRESS}                                                                          
-  ^\s+Status\s+:\s+${STATUS}                                                                       
-  ^\s+CPU\sUtilization\s+:\s+${CPU}                                                                      
+  ^\s+MAC\s+Address\s+:\s+${MAC_ADDRESS}
+  ^\s+Status\s+:\s+${STATUS}
+  ^\s+Serial\s+Number\s+:\s+${SERIAL_NUMBER}
+  ^\s+CPU\s+Utilization\s+:\s+${CPU}
   ^\s+Memory\sUtilization\s+:\s+${MEMORY}
```
mjbear, line 22: I was only looking at the changed lines earlier and not those that were unchanged. Oops.

We could change the whitespace regex for Memory Utilization to be future-proofed to accept one or more spaces instead of just one. Not only does this future-proof, but also follows the pattern of other rules using `\s+`.

```suggestion
  ^\s+Memory\s+Utilization\s+:\s+${MEMORY}
```

ritkit: Ya fair, I was curious why some items had \s+ and others not.

ntc_templates/templates/aruba_aoscx_show_vsf_detail.textfsm
mjbear: > @mjbear so I cannot find the exact output that broke. Though I actually think that using the \S+ was there to force it to detect the first word and then grab everything else in the line. With the .* it is not specific enough and could just grab everything after the `:` which means it might grab a space even with the \s+ preceding it.

Understood.
Hopefully whatever else is out there in terms of statuses can be "covered" by the test data that has the two words `No Split`. :grinning:

The rule regex for the Status line already has whitespace matched (after the colon) so that shouldn't be an issue. If there were to be whitespace _after_ a status word, that would be matched.

_(If we are/were concerned about trailing whitespace, several rules could be modified to account for trailing whitespace.)_

> I always have been of the opinion of having regex as specific as possible.

There was a point at which I had the same opinion. After working on these templates for a while I've seen how fragile specific regexes can be. (Simpler and flexible is better.)

:question: Is there ever an instance where Status is or could be empty?
(What if the switches are not stacked?)
:dart: **If you're able to, please provide raw AOS-CX output for VSF where there is only one switch.**

Based on the test data we have presently, there's not a functional reason to change `STATUS` from `.*`. My suggestion would be to locate test data support a regex change. (Without test data to support the change, a regression could be introduced.)

Thank you!

ritkit: First here is output from a single switch. I obviously changed the mac and serial.

```
VSF Stack
        MAC Address                  : 89:ab:cd:ef:01:23
        Secondary                    :
        Topology                     : standalone
        Status                       : No Split
        Split Detection Method       : None
        Software Version             : FL.10.10.1030
        Force Autojoin               : Disabled
        Autojoin Eligibility Status  : Not Eligible
        Autojoin Ineligibility Reason: Configuration changes detected
        Name                         : Aruba-VSF-6300
        Contact                      :
        Location                     :

Member ID                            : 1
        MAC Address                  : 89:ab:cd:ef:01:23
        Type                         : JL659A
        Model                        : 6300M 48-port HPE Smart Rate 1/2.5/5GbE Class 6 PoE and 4-port SFP56 Switch
        Status                       : Conductor
        ROM Version                  : FL.01.11.0001
        Serial Number                : SG1234ABCD
        Uptime                       : 4 weeks, 2 hours, 35 minutes
        CPU Utilization              : 23%
        Memory Utilization           : 35%
        VSF Link 1                   :
        VSF Link 2                   :
```

And aye aye, I will leave the Status regex the same. If I find an example that breaks it I will resubmit a pull request with the new test and change.

mjbear: @ritkit 
So it seems like whether stackable switches are stacked or not they have a specific status (non-blank)

Interesting, the new name is Conductor.
~I wonder what the member switches are called.~ :grinning:
[conductor, standby, and member](https://www.arubanetworks.com/techdocs/central/2.5.7/content/nms/aos-cx/sw-stk-cx/sw-stk-mon.htm)

If you would, please add the above raw output into a new raw test file (ex: show_vsf_detail2.raw) [in here](https://github.com/networktocode/ntc-templates/tree/master/tests/aruba_aoscx/show_vsf_detail) and run the test script to generate a new parsed yaml file. The idea is to keep the old test data and add new so there's coverage for several types of CLI output.

Any questions, please ask.

ritkit:
@mjbear I pulled the changes we discussed and added the second raw test with second yml output. Take a look and see if it looks correct.

ntc_templates/templates/aruba_aoscx_show_vsf_detail.textfsm
```diff
@@ -1,18 +1,24 @@
...
-  ^\s+CPU\sUtilization\s+:\s+${CPU}                                                                      
-  ^\s+Memory\sUtilization\s+:\s+${MEMORY}
+  ^\s+MAC\s+Address\s+:\s+${MAC_ADDRESS}
+  ^\s+Status\s+:\s+${STATUS}
+  ^\s+Serial\s+Number\s+:\s+${SERIAL_NUMBER}
+  ^\s+CPU\s+Utilization\s+:\s+${CPU}
+  ^\s+Memory\s+Utilization\s+:\s+${MEMORY}
   ^\s+VSF\s+Link\s+1\s+:${VSF_LINK_1}
-  ^\s+VSF\s+Link\s+2\s+:${VSF_LINK_2}
+  ^\s+VSF\s+Link\s+2\s+:${VSF_LINK_2}
```
jmcgill298, line 24:
```suggestion
  ^\s+VSF\s+Link\s+2\s+:${VSF_LINK_2}

```

tests/aruba_aoscx/show_vsf_detail/show_vsf_detail2.raw
```diff
@@ -0,0 +1,26 @@
...
+        Type                         : JL659A
+        Model                        : 6300M 48-port HPE Smart Rate 1/2.5/5GbE Class 6 PoE and 4-port SFP56 Switch
+        Status                       : Conductor
+        ROM Version                  : FL.01.11.0001
+        Serial Number                : SG1234ABCD
+        Uptime                       : 4 weeks, 2 hours, 35 minutes
+        CPU Utilization              : 23%
+        Memory Utilization           : 35%
+        VSF Link 1                   :
+        VSF Link 2                   :
```
jmcgill298, line 26:
```suggestion
        VSF Link 2                   :

```

jmcgill298 (approved):

PR #1944, opened 2024-12-23
Huawei smartax display ont info summary

evilmonkey19:
Fixing bug when having multiple outputs from different ports


mjbear:
> Fixing bug when having multiple outputs from different ports

Howdy @evilmonkey19!
There are some merge conflicts that need fixed up. I'll see about these later today.

I have a few thoughts on some regexes, but it won't be until later that I can _monkey_ with those to see how what I think translates when parsing the test data.

evilmonkey19:
Howdy @mjbear ! No problem 😄. I know that aren't the best templates ever, but they are quite difficult to do, specially because of the multiple ports output.

ntc_templates/templates/huawei_smartax_display_ont_info_summary_ont.textfsm
```diff
@@ -1,17 +1,23 @@
...
 Value LAST_UPTIME (\S+\s*\S+)
 Value LAST_DOWNTIME (\S+\s*\S+)
 Value LAST_DOWNCAUSE (\S+)
 
 Start
-  ^\s+-
-  ^\s+ONT\s+Run\s+Last\s+Last\s+Last
-  ^\s+ID\s+State\s+UpTime\s+DownTime\s+DownCause -> ONTs
+  ^\s*Command\s*is\s*being\s*executed\.\s*Please\s*wait\s*$$
+  ^\s*-
```
mjbear, line 10: This could match a single hyphen with anything after it and accidentally match more lines than intended.

I've updated the rule to allow a hyphen with whitespace on either side, but _not_ other characters.
This will match the horizontal separator lines and is unlikely to match much else.

```suggestion
  ^\s*-+\s*$$
```

ntc_templates/templates/huawei_smartax_display_ont_info_summary_sn.textfsm
```diff
@@ -1,17 +1,26 @@
...
+Value TX_POWER_DBM (-?\d+\.\d+)
+Value DESCRIPTION (\S.+\S)
 
 Start
+  ^\s*Command\s*is\s*being\s*executed\.\s*Please\s*wait\s*$$
+  ^\s*In\s*port\s*${PORT},\s*the\s*total\s*of\s*ONTs\s*are:\s*\d+,\s*online:\s*\d+\s*$$
+  ^\s*ONT\s*Run\s*Last\s*Last\s*Last\s*$$
+  ^\s*ID\s*State\s*UpTime\s*DownTime\s*DownCause\s*$$
+  ^\s*\d+\s*\S+\s*(-|\S+\s\S+)\s*(-|\S+\s*\S+)\s*\S+\s*$$
   ^\s+-
```
mjbear, line 16: Similar to the other template above...
This will match the horizontal separator lines and is unlikely to match much else.

```suggestion
  ^\s+-+\s*$$
```

mjbear:
@evilmonkey19
I submitted a PR https://github.com/evilmonkey19/ntc-templates/pull/4 against your branch to clean up the merge conflicts.

Other items:
:dart: There's a template for huawei_smartax_display_ont_info_summary_sn.textfsm in this PR as well. Was that intended?

:earth_americas: It does not appear there is test data for the summary_sn template.

Thank you.
Merry Christmas! :christmas_tree:

evilmonkey19:
Howdy @mjbear!

This is not quite an usual template as you may see that there are multiple tables into one. The technique used here is using two templates and do the matching using the `key` keyword. Instead of seeing both as 2 separate templates you should see them as only one. Actually, the resulting YAML file have variables from both templates. For instance, just looking at one template and the resulting YAML you will more variables in the YAML file than in the template.

This is a special technique that @jvanderaa and @jmcgill298 told me long ago :)

mjbear:
> Howdy @mjbear!
> 
> This is not quite an usual template as you may see that there are multiple tables into one. The technique used here is using two templates and do the matching using the `key` keyword. Instead of seeing both as 2 separate templates you should see them as only one. Actually, the resulting YAML file have variables from both templates. For instance, just looking at one template and the resulting YAML you will more variables in the YAML file than in the template.

Ah, that makes sense.
 
> This is a special technique that @jvanderaa and @jmcgill298 told me long ago :)

Neat, I'll have to give this more of a look shortly.

ntc_templates/templates/huawei_smartax_display_ont_info_summary_ont.textfsm
mjbear: @evilmonkey19
Please review the above suggestion regarding the dashed separators.

ntc_templates/templates/huawei_smartax_display_ont_info_summary_sn.textfsm
mjbear: @evilmonkey19
Similarly another dashed separator item to review. Thank you!

jmcgill298 (approved):

Issue #1959, opened 2025-01-06
Create cisco_asa_show_running-config_access-group


wollmannbruno:
### Environment
* ntc-templates version:  6.0.0

### Proposed Functionality
Create a new template for the `show running-config access-group` command on the Cisco ASA

### Use Case
Ingest the access-list configured on an interface.

#### Proposed Template
```text
Value ACCESS_LIST (\S+)
Value INTERFACE (\S+)


Start
  ^access-group\s+${ACCESS_LIST}\s+in\s+interface\s+${INTERFACE}\s* -> Record
  ^. -> Error "Did not match any entries"

EOF
```

#### Raw Input
```text
# show running-config access-group
access-group INSIDE_acl in interface inside
access-group dmz1_acl in interface dmz1
access-group dmz2_acl in interface dmz2
access-group dmz3_acl in interface dmz3
access-group dmz4 in interface dmz4
access-group dmz5 in interface dmz5
access-group OUTSIDE_acl in interface outside
```

#### Output
```yaml
---
parsed_sample:
  - access_list: "INSIDE_acl"
    interface: "inside"
  - access_list: "dmz1_acl"
    interface: "dmz1"
  - access_list: "dmz2_acl"
    interface: "dmz2"
  - access_list: "dmz3_acl"
    interface: "dmz3"
  - access_list: "dmz4"
    interface: "dmz4"
  - access_list: "dmz5"
    interface: "dmz5"
  - access_list: "OUTSIDE_acl"
    interface: "outside"
```

mjbear:
> ### Proposed Functionality
> 
> Create a new template for the `show running-config access-group` command on the Cisco ASA

Hello @wollmannbruno
Are you interested in submitting a pull request (PR) to submit this template to the project?

wollmannbruno:
Hi @mjbear 
I just submitted PR [#1962](https://github.com/networktocode/ntc-templates/pull/1962).

Thanks.

mjbear:
@jmcgill298
This issue can be resolved since PR #1962 was merged in. Thank you!

jmcgill298:
Thanks @mjbear

PR #1962, opened 2025-01-07
Create cisco_asa_show_running-config_access-group

wollmannbruno:
This PR is for [issue#1959](https://github.com/networktocode/ntc-templates/issues/1959).


ntc_templates/templates/cisco_asa_show_running-config_access-group.textfsm
```diff
@@ -0,0 +1,9 @@
+Value ACCESS_LIST (\S+)
+Value INTERFACE (\S+)
+
+
```
mjbear, line 4: Minor item, we can drop one of the two blank lines between capture group definition and the Start state.

```suggestion
```

ntc_templates/templates/cisco_asa_show_running-config_access-group.textfsm
```diff
@@ -0,0 +1,8 @@
+Value ACCESS_LIST (\S+)
+Value INTERFACE (\S+)
+
+Start
+  ^access-group\s+${ACCESS_LIST}\s+in\s+interface\s+${INTERFACE}\s* -> Record
+  ^. -> Error "Did not match any entries"
+
+EOF
```
mjbear, line 8: * Since we're matching the entire line anyways, we can also add `$$` to anchor to the end of line too.
   * (Tested still works with the test data. Good job!)
* EOF is inherent so we don't need to explicitly call it out. (Saves us a few bits and bytes.)

```suggestion
  ^access-group\s+${ACCESS_LIST}\s+in\s+interface\s+${INTERFACE}\s*$$ -> Record
  ^. -> Error "Did not match any entries"
```

jmcgill298 (approved):

PR #1921, opened 2024-12-02
adding juniper_junos_show_system_processes_summary.textfsm support

jnicholson56:
Adding support for Juniper Junos **show system processes summary** command.

PR #1965, opened 2025-01-07
adding juniper_junos_show_system_processes_summary.textfsm support 

jmcgill298:


jmcgill298:
Replacing #1921 to add entry in the index file

PR #1971, opened 2025-01-10
Add new template for Fsas Si-R `show ether`

caribouHY:
I added a new template for Fsas Si-R  `show ether`.


ntc_templates/templates/fsas_sir_show_ether.textfsm
```diff
@@ -0,0 +1,30 @@
+Value Required ETHER_GROUP (\d+)
+Value Required ETHER_PORT (\d+)
+Value DESCRIPTION (\S+)
+Value STATUS (.+?)
+Value STATUS_AUTO (auto)
+Value STATUS_SPEED ([\d\.]+[MG])
```
mjbear, line 6: I wasn't able to find an instance where a literal period shows up in STATUS_SPEED.
If you're aware of an instance of this, please add a copy of that raw cli output to this PR. Thank you!

If you accept this change, please click the "commit suggestions" button.

```suggestion
Value STATUS_SPEED ([\d]+[MG])
```

tests/fsas_sir/show_ether/fsas_sir_show_ether.yml
```diff
@@ -0,0 +1,82 @@
...
+    ether_port: "1"
+    flow_control_receive: "on"
+    flow_control_send: "off"
+    media: "Metal"
+    since: "Feb 16 21:28:21 2024"
+    status: "100M Full MDI"
+    status_auto: ""
+    status_duplex: "Full"
+    status_mdi: "MDI"
+    status_speed: "100M"
```
mjbear, line 17: It's a bit of a bummer the status data is captured twice.
I do notice that's likely because of the `offline (backup)` and `disable` statuses.

caribouHY: Thank you for the review. Would you prefer not to capture `STATUS` when not `offline` or `disalble`?
```suggestion
    status: ""
    status_auto: ""
    status_duplex: "Full"
    status_mdi: "MDI"
    status_speed: "100M"
```

```ntc_templates/templates/fsas_sir_show_ether.textfsm
  ^status\s+:\s+(${STATUS_AUTO}\s+)?${STATUS_SPEED}\s+${STATUS_DUPLEX}\s+${STATUS_MDI}\s*$$
```

mjbear: > Thank you for the review. Would you prefer not to capture `STATUS` when not `offline` or `disalble`?
> 
> ```
>   ^status\s+:\s+(${STATUS_AUTO}\s+)?${STATUS_SPEED}\s+${STATUS_DUPLEX}\s+${STATUS_MDI}\s*$$
> ```

@jmcgill298 @jvanderaa @itdependsnetworks
What are your thoughts on the `STATUS` capturing data that _sometimes_ is captured in other capture groups?

The downside here is that the `STATUS` capture group would have to become much more rigid (to be the two statuses in the present test data) so it doesn't match the speed/duplex/MDI.

jmcgill298: I think if the goal is to split the status field into multiple capture groups, then the status field should capture `auto`, `disable`, and `offline` text, and then it would be an empty string if it is set to a particular speed.

ntc_templates/templates/fsas_sir_show_ether.textfsm
```diff
@@ -0,0 +1,30 @@
+Value Required ETHER_GROUP (\d+)
+Value Required ETHER_PORT (\d+)
+Value DESCRIPTION (\S+)
+Value STATUS (.+?)
+Value STATUS_AUTO (auto)
+Value STATUS_SPEED ([\d]+[MG])
```
mjbear, line 6: Square brackets aren't needed around the digit regex.

```suggestion
Value STATUS_SPEED (\d+[MG])
```

ntc_templates/templates/fsas_sir_show_ether.textfsm
```diff
@@ -0,0 +1,30 @@
+Value Required ETHER_GROUP (\d+)
+Value Required ETHER_PORT (\d+)
+Value DESCRIPTION (\S+)
```
mjbear, line 3: Is it possible for a description on this platform to accept a multi-word description?
If yes, we may want to update the regex.

```suggestion
Value DESCRIPTION (.+?)
```

ntc_templates/templates/fsas_sir_show_ether.textfsm
```diff
@@ -0,0 +1,30 @@
...
+Value SINCE (.+?)
+Value CONFIG_MODE (\S+)
+Value CONFIG_MDI (\S+)
+Value CONFIG_MEDIA (\S+)
+
+Start
+  ^\[ETHER -> Continue.Record
+  ^\[ETHER\s+GROUP\-${ETHER_GROUP}\s+PORT\-${ETHER_PORT}\]\s*$$
+  ^description\s+:(\s+${DESCRIPTION})?\s*$$
+  ^status\s+:\s+(${STATUS_AUTO}\s+)?${STATUS_SPEED}\s+${STATUS_DUPLEX}\s+${STATUS_MDI}\s*$$ -> Continue
```
mjbear, line 22: Change to use just the STATUS capture group.
Remove the Continue directive which in this case causes parsing issues when the regex for STATUS is changed.

```suggestion
  ^status\s+:\s+(${STATUS}\s+)?${STATUS_SPEED}\s+${STATUS_DUPLEX}\s+${STATUS_MDI}\s*$$
```

ntc_templates/templates/fsas_sir_show_ether.textfsm
```diff
@@ -0,0 +1,30 @@
...
+  ^status\s+:\s+${STATUS}\s*$$
+  ^media\s+:\s+${MEDIA}s*$$
+  ^flow\s+control\s+:\s+(\-|send\s+${FLOW_CONTROL_SEND},\s+receive\s+${FLOW_CONTROL_RECEIVE})\s*$$
```
mjbear, line 25: Functionally the dash doesn't need escaped.

```suggestion
  ^flow\s+control\s+:\s+(-|send\s+${FLOW_CONTROL_SEND},\s+receive\s+${FLOW_CONTROL_RECEIVE})\s*$$
```

tests/fsas_sir/show_ether/fsas_sir_show_ether.yml
mjbear: > I think if the goal is to split the status field into multiple capture groups, then the status field should capture `auto`, `disable`, and `offline` text, and then it would be an empty string if it is set to a particular speed.

I like this idea.
I cooked up some suggestions.

ntc_templates/templates/fsas_sir_show_ether.textfsm
mjbear: It's possible I should have just worked on this locally in order to easily regenerate the yaml structured output.
I may convert all of these suggestions into a PR if @caribouHY hasn't accepted them by that point.
Thank you.

ntc_templates/templates/fsas_sir_show_ether.textfsm
```diff
@@ -0,0 +1,30 @@
+Value Required ETHER_GROUP (\d+)
+Value Required ETHER_PORT (\d+)
+Value DESCRIPTION (\S+)
+Value STATUS (.+?)
+Value STATUS_AUTO (auto)
```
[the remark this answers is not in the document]
mjbear: The "commit suggestion" button can help easily merge in suggestions. Thank you.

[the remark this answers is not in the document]
caribouHY: `STATUS` have to support the following status:
- `offline`
- `offline (backup)`
- `offline (startup down)`
- `offline (recovery manual)`

I added a test case.

jmcgill298 (approved):

PR #1974, opened 2025-01-14
Juniper junos show system processes brief

jnicholson56:
Updating juniper_junos_show_system_processes_brief.textfsm to support both Classic and EVO varieties of Junos.


jnicholson56:
Set this to draft status until it is verified by someone else. This does work when using output from both Junos Classic and EVO. I had to make a few adjustments to the template to make it work for both properly. I tested using the Nornir site and the output in the test files I provided.

jnicholson56:
I did a double check for the `{master}` in the router output.  Netmiko strips that data out, but other sources such as Ansible do not. In Dual RE setups, Junos EVO is returning the `{master}` designation in the output. I can also verify this with a `| display` output on the CLI itself.

ntc_templates/templates/juniper_junos_show_system_processes_brief.textfsm
```diff
@@ -11,12 +11,27 @@ Value SWAP_USED (\S+)
...
   ^\%Cpu
   ^${MEMORY_FORMAT}iB\s+(Mem|MEMORY)\s+:\s+${MEMORY_ACTIVE}\s+total,\s+${MEMORY_FREE}\s+free,\s+${MEMORY_USED}\s+used,\s+${MEMORY_BUFFER}\s+buff/cache\s*
   ^${SWAP_FORMAT}iB\s+Swap:\s+${SWAP_TOTAL}\s+total,\s+${SWAP_FREE}\s+free,\s+${SWAP_USED}\s+used\.\s+${SWAP_AVAIL}\s+avail\s+(Mem|MEMORY)\s* -> Record
+  ^{(master|backup)}
   ^\s*$$
   ^. -> Error
+
+# Junos Classic template
+TYPE2
+  ^last pid:
```
mjbear, line 32: * Replace some literal spaces with whitespace regex
* Replace some whitespace `\s*` regexes with `\s+` where our test data shows a minimum of one space
* Percent symbol doesn't need escaped with a backslash
* Manipulate the Mem or MEMORY regex to consolidate

```suggestion
  ^-+\s*$$ -> TYPE1
  ^last\s*pid: -> TYPE2

# JUNOS EVO template
TYPE1
  ^-+\s*$$
  ^node:\s+${NODE}
  ^top\s+
  ^Tasks:\s+
  ^%Cpu\s+
  ^${MEMORY_FORMAT}iB\s+M(em|EMORY)\s+:\s+${MEMORY_ACTIVE}\s+total,\s+${MEMORY_FREE}\s+free,\s+${MEMORY_USED}\s+used,\s+${MEMORY_BUFFER}\s+buff/cache\s*
  ^${SWAP_FORMAT}iB\s+Swap:\s+${SWAP_TOTAL}\s+total,\s+${SWAP_FREE}\s+free,\s+${SWAP_USED}\s+used\.\s+${SWAP_AVAIL}\s+avail\s+M(em|EMORY)\s* -> Record
  ^{(master|backup)}\s*$$
  ^\s*$$
  ^. -> Error

# Junos Classic template
TYPE2
  ^last\s+pid:
```

mjbear:
> I did a double check for the `{master}` in the router output. Netmiko strips that data out, but other sources such as Ansible do not. In Dual RE setups, Junos EVO is returning the `{master}` designation in the output. I can also verify this with a `| display` output on the CLI itself.

Thank you for verifying.

What I found does indicate the `{master}` output is natively sent back by the JunOS device and Netmiko has code in place to trim that text out.
https://github.com/ktbyers/netmiko/issues/1328
https://github.com/ktbyers/netmiko/pull/2787/files

jnicholson56:
> > I did a double check for the `{master}` in the router output. Netmiko strips that data out, but other sources such as Ansible do not. In Dual RE setups, Junos EVO is returning the `{master}` designation in the output. I can also verify this with a `| display` output on the CLI itself.
> 
> Thank you for verifying.
> 
> What I found does indicate the `{master}` output is natively sent back by the JunOS device and Netmiko has code in place to trim that text out. [ktbyers/netmiko#1328](https://github.com/ktbyers/netmiko/issues/1328) https://github.com/ktbyers/netmiko/pull/2787/files

I concur with that. Ansible on the other hand which can use textfsm templates, does not strip that out.

jnicholson56:
The suggested fix of `^%Cpu\s+` did not work. The output does not have a space after `Cpu`.

Output:
`%Cpu(s):  0.4 us,  0.8 sy,  0.0 ni, 98.5 id,  0.0 wa,  0.4 hi,  0.0 si,  0.0 st`

I updated that line in the template to `^%Cpu\(s\):\s+` which should work.

@mjbear Out of curiosity, when we are adding lines to the template like this that cover the line being there but don't actually capture anything, is there a rule of thumb as to how much of the line to add?

mjbear:
> The suggested fix of `^%Cpu\s+` did not work. The output does not have a space after `Cpu`.
> 
> Output: `%Cpu(s): 0.4 us, 0.8 sy, 0.0 ni, 98.5 id, 0.0 wa, 0.4 hi, 0.0 si, 0.0 st`
> 
> I updated that line in the template to `^%Cpu\(s\):\s+` which should work.

Ooops on my part there. :disappointed:
Good you fixed it.

> @mjbear Out of curiosity, when we are adding lines to the template like this that cover the line being there but don't actually capture anything, is there a rule of thumb as to how much of the line to add?

That's where I was headed.
From everything I've heard or seen there isn't an absolute project rule on that.
Oftentimes when it's tabular data with column headers at the top, the exact headers will be added. However in this case it's a bit different since the labels for user/system/nice/idle/wait/hw irq/soft irq/steal CPU time are inline with the data.

jnicholson56:
> That's where I was headed. From everything I've heard or seen there isn't an absolute project rule on that. Oftentimes when it's tabular data with column headers at the top, the exact headers will be added. However in this case it's a bit different since the labels for user/system/nice/idle/wait/hw irq/soft irq/steal CPU time are inline with the data.

In this template I set the values that applied to me to variables and ignored the rest. After thinking about this a bit more, I am going to update this template and future ones to include all of the values listed in case those apply to someone else. That seems like the more complete way of doing this so others don't have to circle back later and "finish" the incomplete template I am contributing.

jnicholson56:
> You don't have to capture all the data unless there's really value in capturing that data ([reference](https://github.com/networktocode/ntc-templates/blob/1faea924bad3ecfbcffef7a81922e94117a5c5ef/docs/user/faq.md#why-dont-you-grab-all-of-the-data-in-the-template)).
> 
> (Use your best judgement, hehe.)

I understand and can see there may be value to others by adding the capture now. There just was not value in my particular use case so I never bothered capturing the data. Other than the top uptime lines, it was pretty quick and easy to add the other two lines to the template using the syntax we had already established in the work I had submitted.

mjbear (approved):

PR #1993, opened 2025-01-26
Adding new template for Huawei SmartAX: `display location <mac> ont`

evilmonkey19:
Adding new template

### DIFF
--- a/ntc_templates/templates/huawei_smartax_display_location_aaaa-aaaa-aaaa_ont.textfsm
+++ b/ntc_templates/templates/huawei_smartax_display_location_aaaa-aaaa-aaaa_ont.textfsm
@@ -1,5 +1,5 @@
-Value MAC (\S+)
-Value MAC_TYPE (\w+)
+Value MAC_ADDRESS (\S+)
+Value MAC_ADDRESS_TYPE (\w+)
 Value FSP (\d+\s*\/\s*\d+\s*\/\d+)
 Value ONT_ID (\d+)
 Value PORT_ID (\d+)
@@ -10,7 +10,7 @@ Start
   ^\s*It\s+will\s+take\s+several\s+minutes,\s+please\s+wait\s+or\s+press\s+CTRL_C\s+to\s+break\s*$$
   ^\s*-+\s*$$
   ^\s*(MAC\s+){2}TYPE\s+F/S/P\s+ONT\s+ID\s+Port\s+ID\s+Port\s+Type\s+VLAN\s*$$
-  ^\s*${MAC}\s+${MAC_TYPE}\s+${FSP}\s+${ONT_ID}\s+${PORT_ID}\s+${PORT_TYPE}\s+${VLAN_ID}\s*$$ -> Record
+  ^\s*${MAC_ADDRESS}\s+${MAC_ADDRESS_TYPE}\s+${FSP}\s+${ONT_ID}\s+${PORT_ID}\s+${PORT_TYPE}\s+${VLAN_ID}\s*$$ -> Record
   ^\s*{\s+<cr>\|+<K>\s+}:\s*$$
   ^\s+Command:\s*$$
   ^\s+display\s+location\s+\S+\s+ont\s*$$
--- a/ntc_templates/templates/huawei_smartax_display_location_aaaa-aaaa-aaaa_ont.textfsm
+++ b/ntc_templates/templates/huawei_smartax_display_location_aaaa-aaaa-aaaa_ont.textfsm
@@ -7,7 +7,7 @@ Value PORT_TYPE (\w+)
 Value VLAN (\d+)
 
 Start
-  ^\s*It\s*will\s*take\s*several\s*minutes,\s*please\s*wait\s*or\s*press\s*CTRL_C\s*to\s*break\s*$$
+  ^\s*It\s+will\s+take\s+several\s+minutes,\s+please\s+wait\s+or\s+press\s+CTRL_C\s+to\s+break\s*$$
   ^\s*-+\s*$$
   ^\s*MAC\s+MAC\s+TYPE\s+F/S/P\s+ONT\s+ID\s+Port\s+ID\s+Port\s+Type\s+VLAN\s*$$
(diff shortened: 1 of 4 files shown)


ntc_templates/templates/huawei_smartax_display_location_aaaa-aaaa-aaaa_ont.textfsm
```diff
@@ -0,0 +1,14 @@
+Value MAC (\S+)
+Value MAC_TYPE (\w+)
+Value FSP (\d+\s*\/\s*\d+\s*\/\d+)
+Value ONT_ID (\d+)
+Value PORT_ID (\d+)
+Value PORT_TYPE (\w+)
+Value VLAN (\d+)
+
+Start
+  ^\s*It\s*will\s*take\s*several\s*minutes,\s*please\s*wait\s*or\s*press\s*CTRL_C\s*to\s*break\s*$$
```
mjbear, line 10: There will likely always be at least one space in this message/sentence.

```suggestion
  ^\s*It\s+will\s+take\s+several\s+minutes,\s+please\s+wait\s+or\s+press\s+CTRL_C\s+to\s+break\s*$$
```

itdependsnetworks:
does huawei support a `display location ont` or similar? In general ntc templates presumes some post-processing, and would be less complex to support single way to get data.

evilmonkey19:
Let me check it out and i let you know :)

evilmonkey19:
At least in Huawei SmartAX MA5608T doesn't seem like there is:
```
OLT#display location ?           
---------------------------------------------
  Command of privilege Mode:
---------------------------------------------
mac-addr<P><XXXX-XXXX-XXXX>
                      MAC address
```

I will look others

evilmonkey19:
According to Huawei docs for example the output is as follows:
```
huawei#display location 00e0-fc00-1000 ont
{ <cr>||<K> }:

  Command:
          display location 00e0-fc00-1000 ont
  It will take several minutes, and console may be timeout, please use command
idle-timeout to set time limit
  Are you sure to query MAC address location ? (y/n)[n]:y
  ----------------------------------------------------------------------
  MAC             MAC TYPE  F/S/P    ONT ID  Port ID  Port Type  VLAN   
  ----------------------------------------------------------------------
  00e0-fc00-1000  DYNAMIC   0/ 1/0        0        1  ETH          10   
  ----------------------------------------------------------------------
  ```
  
  Actually, you can omit the (y/n) section for instance

**Comment by mjbear** [comment]
> According to Huawei docs for example the output is as follows:
> 
> ```
> huawei#display location 00e0-fc00-1000 ont
> { <cr>||<K> }:
> 
>   Command:
>           display location 00e0-fc00-1000 ont
>   It will take several minutes, and console may be timeout, please use command
> idle-timeout to set time limit
>   Are you sure to query MAC address location ? (y/n)[n]:y
>   ----------------------------------------------------------------------
>   MAC             MAC TYPE  F/S/P    ONT ID  Port ID  Port Type  VLAN   
>   ----------------------------------------------------------------------
>   00e0-fc00-1000  DYNAMIC   0/ 1/0        0        1  ETH          10   
>   ----------------------------------------------------------------------
> ```
> 
> Actually, you can omit the (y/n) section if you're using the mmi-mode

Probably should factor for the lowest common denominator ... and account for mmi not being enabled.

https://info.support.huawei.com/hedex/hdx.do?docid=EDOC1100278054&id=EN-US_CLIREF_0176920492

**Comment by evilmonkey19** [comment]
For the first time we will disagree! 😂

When you're using the `scroll` and the `undo smart`, which are used in Netmiko for accessing automatically the terminal correctly, we avoid outputting this part:
```
{ <cr>||<K> }:

  Command:
          display location 00e0-fc00-1000 ont
 ```

mjbear:
> For the first time we will disagree! 😂

:slightly_smiling_face: As long as it works I'm good with it.

> When you're using the `scroll` and the `undo smart`, which are used in Netmiko for accessing automatically the terminal correctly, we avoid outputting this part:
> 
> ```
> { <cr>||<K> }:
> 
>   Command:
>           display location 00e0-fc00-1000 ont
> ```

Now this is related to Juniper I believe, but somebody mentioned that while Netmiko prunes out some text that Ansible does not. So this means the template should include those items Netmiko prunes out so that the template can work with either automation tool.

I don't know if there's a similar situation for Huawei, but I guess that can be addressed later if something breaks.

For the minimal extra lines, it may be worth adding them in (along with raw data to support this) so that it doesn't need revisited again (or for a long long time).

itdependsnetworks:
Agreed, we should take from prompt to prompt all of the output, regardless less of what the libraries may or may not do.

mjbear:
> Agreed, we should take from prompt to prompt all of the output, regardless less of what the libraries may or may not do.

:+1:

If possible, we probably should add a template that supports all of the following:

```
display location <mac>
display location <mac> fabric
display location <mac> ont
```

:dart: Thoughts?

@evilmonkey19 Are you able to test these alternative commands and share if the output is similar amongst the three commands? :hugs: Please and thank you!

ref: https://networkingfromzero.wordpress.com/2017/03/31/gpon-basic-command-olt-huawei/
_Though this says **OLT** -- plus Wordpress might have munged some of the horizontal separator text (so I'm only using it as reference and not test data)._

```
How To find ONT by Mac

    OL01#display location aaaa-bbbb-cccc
    { <cr>|fabric<K>|ont<K> }:

    Command:
    display location 4cf9-5dd3-1914
    It will take several minutes, and console may be timeout, please use command  idle-timeout to set time limit
    Are you sure to query MAC address location ? (y/n)[n]:y
    ———————————————————————–
    SRV-P BUNDLE TYPE MAC            MAC TYPE F /S /P   VPI  VCI   VLAN ID
    INDEX INDEX
    ———————————————————————–
    5493     –  gpon aaaa-bbbb-cccc dynamic  0 /6 /6   50   1         203
    ———————————————————————–
```

mjbear:
Since this is a _brand new template_, I normalized a few capture groups:

* VLAN to VLAN_ID
* MAC to MAC_ADDRESS
* MAC_TYPE to MAC_ADDRESS_TYPE

mjbear (approved):

PR #2000, opened 2025-01-30
New template linux: `vzlist`

evilmonkey19:
new template (proxmox)


jvanderaa:
@evilmonkey19 what is vzlist? Just not familiar.

jvanderaa (approved):

mjbear:
> @evilmonkey19 what is vzlist? Just not familiar.

I recognize I'm not evilmonkey :grin:

@jvanderaa
This likely for OpenVZ (virtualization).
https://wiki.openvz.org/User_Guide/Operations_on_Containers#Listing_Containers

At one point OpenVZ only(?) used containerization (and therefore shared the same kernel as the GNU/Linux host-OS so the only supported client OS was GNU/Linux).
https://docs.openvz.org/openvz_users_guide.webhelp/_basics_of_os_virtualization.html

(Though from the hardware virtualization section it looks like they also leverage KVM now.)

evilmonkey19:
Yeees, unfortunately, we still do have some containers using this old Proxmox versions (openvz). We are migrating them, but I still think it might be worth for people in the same situation  :D

PR #2007, opened 2025-01-31
Hp comware display interface fix

CursedEnvy:
Apparently in the past we used cellular sics in our hp comware router, which leads to a difference in output for display interface when it comes to the template currently as is.
This small change to the regex should cover both routers without this interface, as is as well as our old setups where we have a cellular interface.
There is also a line about transfer time present, but this seems rather unimportant so I added it to the drop part

I added an example raw file, made the yaml and ran the tests successfully.
Hope this suffices


ntc_templates/templates/hp_comware_display_interface.textfsm
```diff
@@ -26,10 +26,10 @@ Start
   ^\s*Bandwidth:\s+${BANDWIDTH}
   ^\s*Maximum\s+[Tt]ransmi\S+\s+[Uu]nit:\s+${MTU}
   ^\s*The\sMaximum\s+Transmit\s+Unit\sis\s+${MTU}
-  ^\s*Maximum\s+frame\s+length:\s+${L2MTU}
+  ^\s*Maximum\s+frame\s+length:\s*${L2MTU}
   ^\s*Forbid\s+jumbo\s+frames\s+to\s+pass
   ^\s*The\s[Mm]aximum\s+[Ff]rame\s+[Ll]ength\sis\s+${L2MTU}
```
mjbear, line 31: Making the whitespace more flexible.

```suggestion
  ^\s*The\s+[Mm]aximum\s+[Ff]rame\s+[Ll]ength\s+is\s+${L2MTU}
```

ntc_templates/templates/hp_comware_display_interface.textfsm
```diff
@@ -26,10 +26,10 @@ Start
...
-  ^\s*Internet\s+[Aa]ddress:\s+${IP_ADDRESS}\s+\([Pp]rimary\)
+  ^\s*Internet\s+[Aa]ddress:\s+${IP_ADDRESS}\s+\(([Pp]rimary|[Cc]ellular-[Aa]llocated)\)
   ^\s*Internet\s+[Aa]ddress\sis\s+${IP_ADDRESS}\s+[Pp]rimary
   ^\s*Internet\s+[Aa]ddress:\s+${IP_ADDRESS}\s+\([Ss]ub\)
```
mjbear, line 34: Combine two regexes (similar to PR #2014) and make whitespace more flexible.
(Similar to the above mentioned PR, `Sub` doesn't show up in the test data which I wish it did for test coverage. Oh well.)

```suggestion
  ^\s*Internet\s+[Aa]ddress:\s+${IP_ADDRESS}\s+\(([Pp]rimary|[Ss]ub|[Cc]ellular-[Aa]llocated)\)
  ^\s*Internet\s+[Aa]ddress\s+is\s+${IP_ADDRESS}\s+[Pp]rimary
```

CursedEnvy: Thank you again for your suggestions.
Also I'll try to scour some routers in the field later, I do remember us using sub on some in the past. If I do find one, I'll add it to the test data! Our Comware router install base is 10K, so I am hopeful to find one.

CursedEnvy: It seems that with display interface, it does not show the sub interface ip addresses on the example router I used for display ip interface, where they do show

mjbear: > It seems that with display interface, it does not show the sub interface ip addresses on the example router I used for display ip interface, where they do show EDIT1:Checked another two I found and same behaviour, only have examples on a vlan interface though, they only show sub on display ip interface and not display interface EDIT2: Found a routed interface example and same behaviour

Understood.
Thank you for checking.

The `Sub` line has been in there for around two years. It doesn't really hurt anything and removing it isn't likely worth the risk of accidentally breaking parsing. (In the rare case the output we have might be missing something.)

CursedEnvy: You're welcome and I agree, best to have it there to cover the situation

jvanderaa (approved):

PR #2013, opened 2025-02-04
Fix protocol spelling in hp_comware disp ip interface

CursedEnvy:
Easy to miss, but protocal seems a misspelling of protocol
the "display_ip_interface" template for example uses PROTOCOL_STATUS

This seems a typo to me, other templates use protocol instead of protocal.


jvanderaa:
This will be a breaking change.

jvanderaa:
Tests are failing. Probably need to update the test YAML files.

CursedEnvy:
> Tests are failing. Probably need to update the test YAML files.

Aaah my bad, indeed did not think of those, I've updated them .

mjbear (approved):

PR #2014, opened 2025-02-05
Hp comware display ip interface update

CursedEnvy:
Updated the template to also allow CELLULAR-Allocated instead of Primary when handling an eth-channel interface on a router, this is not possible on the MSR958X and MSR2003X, so no test file for release 9119P17


mjbear:
@CursedEnvy
Thank you for the test data with `Sub`. Great work! :slightly_smiling_face:

CursedEnvy:
> @CursedEnvy Thank you for the test data with `Sub`. Great work! 🙂

Happy to return the favour a little bit!

jvanderaa (approved):

jvanderaa:
Thank you @CursedEnvy

PR #2016, opened 2025-02-07
Update Huawei VRP display vlan

evilmonkey19:
Adding support for AP. I accept suggestions for VLAN_ID and VLAN_TYPE


ntc_templates/templates/huawei_vrp_display_vlan.textfsm
```diff
@@ -40,3 +48,12 @@ STATUS
   ^${VLAN_ID}\s+${VLAN_STATUS}\s+\S+\s+\S+\s+\S+\s+${VLAN_NAME} -> Continue.Record
   ^\d+ -> Clearall
   ^. -> Error
+
+
+VLANS_AP
```
mjbear, line 53: The convention I've seen across templates has been single blank lines between sections.
So here I'm suggesting the minor edit to remove one of those two before the `VLANS_AP` State

By comparison there was already a `STATUS` State section preceded with two blank lines, which could also be reduced to a single blank line if don't mind. :grin: Thank you!

```suggestion

VLANS_AP
```

ntc_templates/templates/huawei_vrp_display_vlan.textfsm
```diff
@@ -36,7 +44,13 @@ VLANS
 
 
 STATUS
```
mjbear, line 46: We could drop this extra blank line before `STATUS` state too.
Thank you @evilmonkey19

```suggestion

STATUS
```

evilmonkey19:
This templates are for Huawei VRP platform, which is used in Huawei Switches and APs. The command `display vlan` is used in both platforms, but they have similar but not exactly the same values. Also I decided to use a more common approach on `display vlan` imitating `show etherchannel brief` from cisco IOS.

As well, I added the command `display vlan brief` which is used in Huawei VRP AP. I reused a template from `display vlan`.

mjbear:
> This templates are for Huawei VRP platform, which is used in Huawei Switches and APs. The command `display vlan` is used in both platforms, but they have similar but not exactly the same values. Also I decided to use a more common approach on `display vlan` imitating `show etherchannel brief` from cisco IOS.

I'm curious, so I looked but didn't find a template for etherchannel brief, but there is one for summary.

Did I miss something?

Thank you.

> As well, I added the command `display vlan brief` which is used in Huawei VRP AP. I reused a template from `display vlan`.

jvanderaa:
I'm working on getting some additional eyes on this one. This does look like a breaking change at the moment. Which then we need to do a major release here soon.

itdependsnetworks:
What information does bried have that hte regular does not? In general, I regret supporting things like "show cdp neighbors" and "show cdp neighbors details". It's extra work and you get everything from details, no exceptions. The compute cost of a few extra lines is so small, it's not worth the development effort and upkeep.

evilmonkey19:
The issue here is not that we are adding extra little details. Myself, I'm configuring Huawei AP and Huawei switches (VRP). Both cases they do have SSH and in both cases the command is the same. However the output isn't, that's why I suggest doing that.

Also, I wanted to make the command more standard similar to `show etherchannel brief` (or similar).

mjbear:
> The issue here is not that we are adding extra little details. Myself, I'm configuring Huawei AP and Huawei switches (VRP). Both cases they do have SSH and in both cases the command is the same. However the output isn't, that's why I suggest doing that.
> 
> Also, I wanted to make the command more standard similar to `show etherchannel brief` (or similar).

What Ken is suggesting is that we shouldn't unnecessarily create templates for the "brief" or "summary" commands (or their vendor/platform equivalents).

A template for the normal or detailed version of the command means ***one template*** that provides the same information (and more) plus it means one template to maintain and not two in the long term.

ntc_templates/templates/huawei_vrp_display_vlan_brief.textfsm
```diff
@@ -0,0 +1,62 @@
...
+  ^\s+(UT|TG):\s*[\w+\.\-\/\(\)]+\s+[\w+\.\-\/\(\)]+\s+[\w+\.\-\/\(\)]+\s+${INTERFACE}\(${INTERFACE_STATUS}\)\s* -> Continue
+  ^\s+${INTERFACE}\(${INTERFACE_STATUS}\)\s* -> Continue
+  ^\s+[\w+\.\-\/\(\)]+\s+${INTERFACE}\(${INTERFACE_STATUS}\)\s* -> Continue
+  ^\d+
+  ^\s+\S+
+  ^\s*-+\s*$$
+  ^\s*$$ -> Start
+  ^. -> Error
+
+
```
mjbear, line 41: Removing an extra blank line. Minor.

```suggestion
```

ntc_templates/templates/huawei_vrp_display_vlan_brief.textfsm
```diff
@@ -13,20 +13,23 @@ Start
...
   ^\s*MP:\s+Vlan-mapping;\s+ST:\s+Vlan-stacking;\s*$$
   ^\s*#:\s+ProtocolTransparent-vlan;\s+\*:\s+Management-vlan;\s*$$
-  ^\s*VID\s+Type\s+Ports\s*$$ -> VLANS_SW
+  ^\s*MAC-LRN:\s+MAC-address\s+learning;\s+STAT:\s+Statistic;\s*$$
+  ^\s*BC:\s+Broadcast;\s+MC:\s+Multicast;\s+UC:\s+Unknown-unicast;\s*$$
+  ^\s*FWD:\s+Forward;\s+DSD:\s+Discard;\s*$$
+  ^\s*VID\s+(?:Type\s+)?Ports\s*$$ -> VLANS_SW
   ^\s*VID\s+Name\s+Status\s+Ports\s*$$ -> VLANS_AP
-  ^\s*VID\s+Status\s+Property\s+MAC\-LRN\s+Statistics\s+Description\s*$$ -> STATUS_SW
+  ^\s*VID\s+(?:Type\s+)?Status\s+Property\s+MAC\-LRN\s+(STAT|Statistics)\s+(?:BC\s+)?(?:MC\s+)?(?:UC\s+)?Description\s*$$ -> STATUS_SW
```
mjbear: @evilmonkey19
What do you think about consolidating some of the repeated and optional chars for BC, MC, UC?

Edit: This could allow for BC BC to be supported, but that's not likely. :sweat_smile:

_If you think this is too complex we can avoid it._

```suggestion
  ^\s*VID\s+(?:Type\s+)?Status\s+Property\s+MAC\-LRN\s+(STAT|Statistics)\s+(?:[BMU]C\s+)*Description\s*$$ -> STATUS_SW
```

mjbear: @evilmonkey19
Thoughts on this BC/MC/UC regex topic?

mjbear:
I resolved an index merge conflict (that's the **Merge branch 'master'** commit)
Thank you.

ntc_templates/templates/huawei_vrp_display_vlan_brief.textfsm
mjbear: > @evilmonkey19 Thoughts on this BC/MC/UC regex topic?

We talked outside of this thread and you said you're open to the regex change if it works.

***It works. :tada:***
This first section above the horizontal line is for **ntc_templates/templates/huawei_vrp_display_vlan_brief.textfsm**

```diff
- ^\s*VID\s+(?:Type\s+)?Status\s+Property\s+MAC\-LRN\s+(STAT|Statistics)\s+(?:BC\s+)?(?:MC\s+)?(?:UC\s+)?Description\s*$$ -> STATUS_SW`
+ ^\s*VID\s+(?:Type\s+)?Status\s+Property\s+MAC\-LRN\s+(STAT|Statistics)\s+(?:(?:[BMU]C\s+)?)*Description\s*$$ -> STATUS_SW

# even more concise
+ ^\s*VID\s+(?:Type\s+)?Status\s+Property\s+MAC\-LRN\s+(STAT|Statistics)\s+(?:[BMU]C\s+)*Description\s*$$ -> STATUS_SW
```

Simplified comparison of the specific regex sections:

```diff
# existing at 31 chars
- `(?:BC\s+)?(?:MC\s+)?(?:UC\s+)?`

# 20 chars, could actually be reduced further, see below
+ (?:(?:[BMU]C\s+)?)*

# 15 chars, updated the regex suggestion above (in the first post / GH code suggestion)
+ (?:[BMU]C\s+)*
```

---

Next subject, only stated in case the question is posed about the other _BC/MC/UC_ or the _Xcast_ words. There are other spots that could be done similarly in those two templates _however_ in those situations they are not optional. With them being mandatory their regex is already pretty short so I'm going to leave them be for the time being. The regex would probably be about as long as the mandatory one.

For example in **ntc_templates/templates/huawei_vrp_display_vlan.textfsm**:

```diff
- ^\s*VID\s+Type\s+Status\s+Property\s+MAC\-LRN\s+STAT\s+BC\s+MC\s+UC\s+Description\s*$$ -> STATUS_SW_2`
+ ^\s*VID\s+Type\s+Status\s+Property\s+MAC\-LRN\s+STAT\s+(?:[BMU]C\s+)+Description\s*$$ -> STATUS_SW_2
```

The specific regexes isolated:

```diff
- BC\s+MC\s+UC\s+
+ (?:[BMU]C\s+)+
```

Reduction by one character, or reduction by three characters if we get rid of `?:` at the beginning of the non-capturing group. Not markedly worth the change in my opinion.

ntc_templates/templates/huawei_vrp_display_vlan_brief.textfsm
```diff
@@ -0,0 +1,64 @@
+Value Key,Required,Filldown VLAN_ID (\d+)
+Value Filldown VLAN_NAME (\w.+\w|\w+)
+Value Filldown MODE ((TG|UT))
+Value Filldown VLAN_STATUS ((enable|disable))
+Value Filldown VLAN_TYPE ((common|uncommon))
+Value List INTERFACE ([\w+\.\-\/]+)
+Value List INTERFACE_STATUS ((U|D))
```
mjbear: Dropping some parentheses that aren't needed. (I will commit this suggestion.)

```suggestion
Value Filldown MODE (TG|UT)
Value Filldown VLAN_STATUS (enable|disable)
Value Filldown VLAN_TYPE (common|uncommon)
Value List INTERFACE ([\w+\.\-\/]+)
Value List INTERFACE_STATUS (U|D)
```

PR #2019, opened 2025-02-07
Normalize pvid with vlan_id

mjbear:
:warning: Breaking Change :warning:

Given PRs that have come in, I noticed that some of the hp_comware templates used `PVID` for the VLAN ID instead of `VLAN_ID`. I decided to adjust that with this PR.

This is along the same vein as PR #1984 and consequently will cause a merge conflict on `data_model.md`. I'll fix that once we merge more PRs in.

#

:spiral_notepad: Note: Several of these same templates use `NAME` for the port description. Though I'm not sure how many of those we should consider normalizing since NAME is pretty common and pattern replaces en masse could be more of a pain than the benefit.

The normalization is probably like anything else - work on those with the highest reward first where it makes sense. :wink:


mjbear:
Rebased off of current upstream state to resolve merge conflicts. :grinning:

jmcgill298 (approved):

PR #2020, opened 2025-02-11
Add hp_comware_display_bgp_peer_ipv4

CursedEnvy:
Adding a template to read out the bgp ipv4 neighbours on a comware router 
(it is also possible to use vprn-instance {instance_name} after the commando, output is the same (except for the neighbours of course))

Test file is the data as shown on a router (when I use my script to read out the config, the new line and spaces before BGP local router are gone, hence the use of * to cover this) 
I choose to Filldown the top variables, because I noticed this was the case on the IOS template as well

Had to get a lot of help from a colleague as my first steps creating something from nothing, I am not good at regex, yet :)


ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
```diff
@@ -0,0 +1,26 @@
...
+Start
+  ^\s*BGP\s+local\s+router\s+ID:\s+${LOCAL_ROUTER_ID}
+  ^\s+Local\s+AS\s+number:\s+${LOCAL_AS_NUMBER}
+  ^\s+Total\s+number\s+of\s+peers:\s+${TOTAL_PEERS}\s+Peers\s+in\s+established\s+state:\s+${PEERS_ESTABLISHED} -> PeerInfo
+  ^. -> Error
+
+PeerInfo
+  ^\s*\*\s+-\s+Dynamically\s+created\s+peer
+  ^\s*Peer\s+AS\s+MsgRcvd\s+MsgSent\s+OutQ\s+PrefRcv\s+Up\/Down\s+State
+  ^\s{2}${PEER}\s+${AS}\s+${MESSAGES_RECEIVED}\s+${MESSAGES_SENT}\s+${OUTPUT_QUEUE}\s+${PREFIXES_RECEIVED}\s+${UP_DOWN}\s+${STATE}\s*$$ -> Record
```
mjbear, line 23: Could we standardize on `REMOTE_ASN` to make this more descriptive?
_(This would mean one less thing to standardize later.)_

For example [nxos_show_ip_bgp_neighbors](https://github.com/networktocode/ntc-templates/blob/master/ntc_templates/templates/cisco_ios_show_ip_bgp_neighbors.textfsm#L3) uses that capture group name.

```suggestion
Value Required REMOTE_ASN (\d+)
Value MESSAGES_RECEIVED (\d+)
Value MESSAGES_SENT (\d+)
Value OUTPUT_QUEUE (\d+)
Value PREFIXES_RECEIVED (\d+)
Value UP_DOWN (\d+h\d+m)
Value STATE (\S+)

Start
  ^\s*BGP\s+local\s+router\s+ID:\s+${LOCAL_ROUTER_ID}
  ^\s+Local\s+AS\s+number:\s+${LOCAL_AS_NUMBER}
  ^\s+Total\s+number\s+of\s+peers:\s+${TOTAL_PEERS}\s+Peers\s+in\s+established\s+state:\s+${PEERS_ESTABLISHED} -> PeerInfo
  ^. -> Error

PeerInfo
  ^\s*\*\s+-\s+Dynamically\s+created\s+peer
  ^\s*Peer\s+AS\s+MsgRcvd\s+MsgSent\s+OutQ\s+PrefRcv\s+Up\/Down\s+State
  ^\s{2}${PEER}\s+${REMOTE_ASN}\s+${MESSAGES_RECEIVED}\s+${MESSAGES_SENT}\s+${OUTPUT_QUEUE}\s+${PREFIXES_RECEIVED}\s+${UP_DOWN}\s+${STATE}\s*$$ -> Record
```

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
```diff
@@ -0,0 +1,26 @@
+Value Filldown LOCAL_ROUTER_ID (\d{1,3}(?:\.\d{1,3}){3}) 
+Value Filldown LOCAL_AS_NUMBER (\d+)
+Value Filldown TOTAL_PEERS (\d+)
```
mjbear, line 3: Does it make sense to capture this info?
(We'd already know how many peers based on the length of the list that is returned.)

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
```diff
@@ -0,0 +1,26 @@
...
+Value Filldown PEERS_ESTABLISHED (\d+)
```
mjbear, line 4: Similar to Total Peers count, is the Established Peer count important since the State is already being captured?

mjbear:
@CursedEnvy
Once we get through the above discussion items, I have some other thoughts.
Thank you.

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
CursedEnvy: Yes I agree, that's indeed more descriptive, oversight on my behalf as there are indeed two or even more ASN in play potentially.

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
CursedEnvy: Coincidentally had an discussion on that with my colleague yesterday, in the end I looked at some ios show ip bgp templates and saw them do this, but I still agree it felt weird and indeed now that you mention it, redundant. He was also in favour of leaving it out.

Only thing I now start to wonder is how will this handle multiple AS neighbours, maybe I'll have to search a test case today.

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
CursedEnvy: I agree, can be left out and handled for example with a property if you turn the dict into a basemodel later.

CursedEnvy:
> @CursedEnvy Once we get through the above discussion items, I have some other thoughts. Thank you.

Sure, thank you for your input always, I am happy to hear your thoughts!

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
CursedEnvy: Update: actually don't think we have such a case of multiple AS neighbours anywhere or at least I'll be searching a needle in a haystack 
Anyway, I removed the total and established peers!

ntc_templates/templates/hp_comware_display_bgp_peer_ipv4.textfsm
```diff
@@ -0,0 +1,24 @@
...
+Start
+  ^\s*BGP\s+local\s+router\s+ID:\s+${LOCAL_ROUTER_ID}
+  ^\s+Local\s+AS\s+number:\s+${LOCAL_AS_NUMBER}
+  ^\s+Total\s+number\s+of\s+peers:\s+(\d+)\s+Peers\s+in\s+established\s+state:\s+(\d+) -> PeerInfo
+  ^. -> Error
+
+PeerInfo
+  ^\s*\*\s+-\s+Dynamically\s+created\s+peer
+  ^\s*Peer\s+AS\s+MsgRcvd\s+MsgSent\s+OutQ\s+PrefRcv\s+Up\/Down\s+State
+  ^\s{2}${PEER}\s+${REMOTE_ASN}\s+${MESSAGES_RECEIVED}\s+${MESSAGES_SENT}\s+${OUTPUT_QUEUE}\s+${PREFIXES_RECEIVED}\s+${UP_DOWN}\s+${STATE}\s*$$ -> Record
```
mjbear, line 21: A state transition isn't necessary since the raw output is pretty simple.
Reduces a couple of lines in total.

```suggestion
  ^\s+Total\s+number\s+of\s+peers:\s+(\d+)\s+Peers\s+in\s+established\s+state:\s+(\d+)
  ^\s*\*\s+-\s+Dynamically\s+created\s+peer
  ^\s*Peer\s+AS\s+MsgRcvd\s+MsgSent\s+OutQ\s+PrefRcv\s+Up\/Down\s+State
  ^\s{2}${PEER}\s+${REMOTE_ASN}\s+${MESSAGES_RECEIVED}\s+${MESSAGES_SENT}\s+${OUTPUT_QUEUE}\s+${PREFIXES_RECEIVED}\s+${UP_DOWN}\s+${STATE}\s*$$ -> Record
```

CursedEnvy: Okay good to know!

jvanderaa (approved):

PR #2021, opened 2025-02-14
fix: alcatel_sros_show_lag_port

ppetilli:
fix: noticed that the output is slightly different in firmware release22. Previous template was based on an output matching firmware r16. Corrected the template to cover both outputs.


ntc_templates/templates/alcatel_sros_show_lag_port.textfsm
mjbear, line 1: Probably shouldn't add `\s*` to the capture group definition (Value line), but instead either make the capture group at the rule level.

I'll see about working up a solution so it's tested. :grin:

tests/alcatel_sros/show_lag_port/show_lag_port_r22.raw
mjbear, line 1: To adequately test for the changes the raw data should cover the modifications being made. Thank you.

1. Where does this raw output have Active Status being absent?
1. Where is the sub-group absent in the raw data?
1. Where is forced absent in the raw data?
1. Where is the priority absent in the raw data?

mjbear: Active status, sub-group, forced, and priority can be reverted without throwing a parsing error presently.

ppetilli: > * Where does this raw output have Active Status being absent?

good catch. I initially captured the data starting with lag-* but that didn't make sense as we get the id number on the next line. the absent Active status, sub-group, forced  and priority  are remnants from that config and should of course not be present of the code anymore.

ntc_templates/templates/alcatel_sros_show_lag_port.textfsm
ppetilli: I have update the code taking into consideration your feedback.

jvanderaa (approved):

jvanderaa:
Thank you @ppetilli

PR #2023, opened 2025-02-15
Replace ubuntu-20.04 with 22.04 in ci.yml

mjbear:
:warning: Likely considered a breaking change by ntc-templates.

Ubuntu 20.04 runners are being deprecated and will be removed in April 2025.
Switching to 22.04

ref: https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ env:
 
 jobs:
   black:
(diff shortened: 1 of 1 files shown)


jvanderaa (approved):

jvanderaa:
This should be good to go. Won't impact any changes on the users here.

PR #2035, opened 2025-02-21
Add additional nonmatching line to EOS show ip mroute vrf all detail template

brandomando:
I found an additional non-matching line in the wild that is breaking at the wildcard `-> Error` line:
```
* - Interface has EVPN information available in the 'detail' command output
```

This change introduces a fix for this pattern


jvanderaa (approved):

jvanderaa:
@brandomando we need a release with this one out?

PR #2037, opened 2025-02-23
New template for Huawei ONT: `display access mode`

evilmonkey19:
This is a platform for the Huawei ONT like Echolife HG8310M


mjbear:
@evilmonkey19
Are there any other values for the access modes?

evilmonkey19:
Not that I have seen so far, but I've only tried with one ONU. Actually it would be easy to have ONUs on GPON mode, 10G EPON, XGSPON, etc...

This type of ONUs are XPON so it can be either EPON or GPON.

mjbear:
> Not that I have seen so far, but I've only tried with one ONU. Actually it would be easy to have ONUs on GPON mode, 10G EPON, XGSPON, etc...
> 
> This type of ONUs are XPON so it can be either EPON or GPON.

Based on that, they'd still be single words without spaces so the capture group regex would be fine.

Good!

mjbear:
I'm guessing we/I merged a PR that was based off this PR since now this one shows as no files changed.

PR #2046, opened 2025-02-23
Adding template for Huawei ONT: `display sn`

evilmonkey19:


ntc_templates/templates/huawei_ont_display_sn.textfsm
```diff
@@ -0,0 +1,7 @@
+Value SERIAL_NUMBER (\S+)
+
+Start
+  ^\s*sn\s*=\s*${SERIAL_NUMBER}\s*$$
```
mjbear, line 4: There are spaces around the equals sign.
@evilmonkey19 Have you ever seen them not be present for output separated by equals signs?

(If you have, we can ignore this suggestion. Thank you!)

```suggestion
  ^\s*sn\s+=\s+${SERIAL_NUMBER}\s*$$
```

jvanderaa (approved):

PR #2048, opened 2025-02-23
Adding new template for Huawei ONT: `display productmac`

evilmonkey19:


ntc_templates/templates/huawei_ont_display_productmac.textfsm
mjbear, line 1: @evilmonkey19
I'm going to figure that not all devices have always have LAN, WAN, etc MACs which is why some patterns show up in multiple State sections?

evilmonkey19: Yeees! ONTs are used as a WAN gateway for multiple operators (look at NEBA network in Spain). Also many some ONTs do have WLAN, others don't, some have multiple LAN ports, etc. The one I used for testing is the EchoLife HG8310M, which only has one LAN port and no WLAN.

jvanderaa (approved):

PR #2050, opened 2025-02-23
Adding new template for Huawei ONT: `display portstatistics portnum 0`

evilmonkey19:


ntc_templates/templates/huawei_ont_display_portstatistics_portnum_0.textfsm
mjbear, line 1: At minimum in the raw data there is one space on either side of the colon `:` for all of those rows. Probably should have those be `\s+` - if you're onboard with that I can submit a suggestion. :wink:

mjbear: I'll work on this.
Resolving the convo. :grinning:

mjbear (approved):

PR #2052, opened 2025-02-24
fix: parsing for cisco ios interface

gertzakis:
When there is empty after `media type is`  in Cisco IOS "show interfaces" command the parsing takes the whole line into speed variable. 
Specifically, `speed: "10Gb/s, link type is auto, media type is"` and not just `10Gb/s`. This PR tries to address this issue


mjbear:
@gertzakis
I submitted a PR https://github.com/gertzakis/ntc-templates/pull/1 against your fork/branch with some suggestions for you to review. Thank you!

jvanderaa (approved):

PR #2053, opened 2025-02-24
Adding new template for Huawei ONT: `display optmode`

evilmonkey19:


ntc_templates/templates/index
mjbear, line 1: @evilmonkey19

1. Do the index entries look right after a few changes to fix the CI tests?
1. As in when you run `display optmode`, you have to type at least `di optm` right?
1. Similarly for `display optic`, you have to type `di opti`, yes?

Thank you!

evilmonkey19: Good point! Yes, my bad :)

mjbear: All good. :slightly_smiling_face:
Thank you for confirming!

mjbear (approved):

PR #2056, opened 2025-02-24
Adding new template for Huawei ONT: `display machineItem`

evilmonkey19:


ntc_templates/templates/index
mjbear, line 1: @evilmonkey19 

Similar to another PR, the absolute minimum typing for this command at the CLI, would be `di mach`, right?

_Please verify on a live device if you would._ Thank you!

evilmonkey19: To my understanding and the ones i've tried yes.

mjbear (approved):

PR #2059, opened 2025-02-24
Adding new template for Huawei ONT: `display lanmac`

evilmonkey19:


mjbear:
@evilmonkey19 
Are there instances where there could be more than one LAN MAC address?

evilmonkey19:
Yes! Typically ONTs have more than 1 LAN port:
rear_left-3473843339

mjbear:
> Yes! Typically ONTs have more than 1 LAN port:

Why I asked is that at one point I wondered if it was worth setting `LAN_MAC` capture group to be a List. Though I don't know that my thought there really gains us much. :sweat_smile:

ntc_templates/templates/huawei_ont_display_lanmac.textfsm
mjbear, line 1: @evilmonkey19
Using `wanmac` as a precedent, are there any concerns with having `LAN_MAC` instead be `MAC_ADDRESS` and be a List data type?

ref: https://github.com/networktocode/ntc-templates/blob/master/tests/huawei_ont/display_wanmac/huawei_ont_display_wanmac.yml

Thoughts everyone?

evilmonkey19: I like the idea! :) I did this first I didn't realize the list thing 😅

mjbear (approved):

PR #2064, opened 2025-02-26
Improve Alcatel AOS show lldp remote-system

evilmonkey19:


ntc_templates/templates/alcatel_aos_show_lldp_remote-system.textfsm
```diff
@@ -7,7 +7,7 @@ Value CHASSIS_SUBTYPE (.+?)
 Value PORT_SUBTYPE (.+?)
 Value NEIGHBOR_INTERFACE (.+?)
 Value NEIGHBOR_NAME (.+?)
-Value NEIGHBOR_DESCRIPTION (.+?)
+Value List NEIGHBOR_DESCRIPTION (\S.+\S)
```
mjbear, line 10: This is _might be_ the only template that uses a `List` for the Neighbor Description.
At that point the data structure for this template would be unlike many of the others.

* [cisco_nxos lldp neigh det](https://github.com/networktocode/ntc-templates/blob/master/tests/cisco_nxos/show_lldp_neighbors_detail/cisco_nxos_show_lldp_neighbors_detail.yml)
* [cisco_ios lldp neigh det](https://github.com/networktocode/ntc-templates/blob/master/tests/cisco_ios/show_lldp_neighbors_detail/cisco_ios_show_lldp_neighbors_detail1.yml)

:warning: Do we really want to make this change using `List` data type?

evilmonkey19: The issue is that the output is multiple lines. What are other options?

mjbear: > The issue is that the output is multiple lines. What are other options?

Right you are. Many of the neighbor descriptions aren't multi-line so until I saw the diffs the reasoning didn't jump out at me.

To your point, without list only the last line would be captured, but there are several before that.

```diff
     mgmt_address: "1.1.1.1"
-    neighbor_description:
-      - "S5700-52P-PWR-LI-AC"
-      - "Huawei Versatile Routing Platform Software"
-      - "VRP (R) software, Version 1.110 (S5700 A000A000A00AAA000)"
-      - "Copyright (C) 2000-2001 HUAWEI TECH Co., Ltd.,"
+    neighbor_description: "Copyright (C) 2000-2001 HUAWEI TECH Co., Ltd.,"
     neighbor_interface: "UPLINK"
```

Thank you for the patience and nudge for me to go test. :blush:

mjbear (approved):
Good work.
Looks good. Thank you!

PR #2067, opened 2025-02-26
Update cisco_nxos_show_nve_vni.textfsm

julien-lebreton:
Add S-ND line and Hybrid mode line
Those 2 lines are mandatory to work properly with version 10.3.4a of NX-OS


mjbear:
@julien-lebreton
We also need [test data](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#tests) included in this PR.

The test data is [necessary](https://github.com/networktocode/ntc-templates/blob/master/docs/user/faq.md#i-simply-want-to-add-my-template-to-the-project-i-do-not-want-to-add-all-of-these-tests-and-controls-can-i-just-do-so) so that we can make sure the template is backwards compatible in the future when more changes are needed.

mjbear:
As time permits I plan to build a topology to get raw output ... if Julien hasn't responded by then. :slightly_smiling_face:

julien-lebreton:
update with raw data in tests/cisco_nxos/show_nve_vni directory

mjbear:
> update with raw data in tests/cisco_nxos/show_nve_vni directory

Hello @julien-lebreton

Thank you for submitting this PR and adding test data!
_You saved me a step in getting test data. :laughing:_

You might see I made changes.

Since the old test data had various VRF names and what not there is some merit in retaining it. This enables us to have both the old and new data to run through the GH Actions CI validating no regressions occur.

mjbear (approved):

PR #2068, opened 2025-03-03
Update cisco_ios_show_power_inline.textfsm

paineta:
There is a blank line between module output, so I'm not sure why that would be an Error and not simply ingested.  Tested on 16.12.8 and 17.9.5.


ntc_templates/templates/cisco_ios_show_power_inline.textfsm
```diff
@@ -21,4 +21,4 @@ Interfaces
   ^\s+\(Watts\)
   ^-+
   ^\s*$$
-  ^. -> Error
+  ^.
```
mjbear, line 24: There must be an [error directive](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#error-directive).

I'd be glad to assist with updating the template if you provide example test data.
Thank you!

```suggestion
  ^. -> Error
```

mjbear:
Ah, I see your plain text file attachment.
I've worked off of that and I have a fix I'll suggest on here.

paineta:
I see.  My apologies for bungling this pull request.  The explicit Error catch does make sense.

mjbear:
I've got a few things to learn myself in how I can add to your PR.

I opened up https://github.com/paineta/ntc-templates/pull/2 thinking I could pull them in here. :zany_face:

mjbear:
@paineta
:dart: We should really get some 16.12.x output where there are multiple switches in the stack. This way we can to some degree guarantee backwards compatibility.

:point_right: Is this something you are willing to help with?

From what I can tell on some hardware I have access to, 17.6.3, 17.9.[45], and 17.12.4 have the extra white space between modules. I don't have any switches still running 16.x code. :man_shrugging:

paineta:
In the interest of not screwing this up more than I have, attached are 3 outputs from IOS-XE stacks for 16.12.x, 17.9.x, and 17.12.x

[cisco_ios_show_power_inline_16-12multi.txt](https://github.com/user-attachments/files/19063306/cisco_ios_show_power_inline_16-12multi.txt)
[cisco_ios_show_power_inline_17-9multi.txt](https://github.com/user-attachments/files/19063307/cisco_ios_show_power_inline_17-9multi.txt)
[cisco_ios_show_power_inline_17-12multi.txt](https://github.com/user-attachments/files/19063308/cisco_ios_show_power_inline_17-12multi.txt)

paineta:
Thank you for the encouragement!  Yes, I will attempt fleshing this dev/test process out on my fork.  Thanks again.

mjbear:
_This probably classifies as a tip_ ... certainly not being critical here.

You'll notice I trimmed had down the raw output and cut out lots of repetition (for the PR to your fork). This reduced the structured output to around approximately 100 lines as opposed to over 2000 with the original raw output.

Since there was repetition there wasn't value in the duplicates. This still gave us the separate module sections and kept the test data that is ran against for each push and each PR a bit _"leaner"_ so to speak.

paineta:
Ok, progress I think.  I've committed updates to my patch-1.

mjbear:
> Ok, progress I think. I've committed updates to my patch-1.

Thank you @paineta.
I wasn't clear that if the raw output is structured the same then we don't need new (nearly duplicate) output. Apologies.

:white_check_mark: The ios12 output is different than the [existing output](https://github.com/networktocode/ntc-templates/blob/master/tests/cisco_ios/show_power_inline/cisco_ios_show_power_inline.raw), so that makes sense to have.

:white_check_mark: We have the `_whitespace` files so that seems to cover versions from 16 to 17 code. ***Please certainly say something if I've missed anything when briefly comparing them. Thank you.***

:question: Other than one less line of whitespace, 16.12 is not different than 17.9 output so we don't need to duplicate that. _So we can remove the \_iosxe16-12stack files from this PR._

:question: Since the first round of output with `_whitespace` identically matches the formatting for 17.12.4, we don't need to duplicate that. _So we can remove the \_iosxe17-12stack files from this PR._

Hopefully this makes sense.

Minor note: If you want your username to show up on commits, that can be configured via the local git cli for [username](https://docs.github.com/en/get-started/git-basics/setting-your-username-in-git) and [email](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address).

paineta:
That makes sense.  From an automation perspective, I was just look at it as a way of tracking tested versions, but perhaps there's a better (platform/command) change log for that?

mjbear:
> That makes sense. From an automation perspective, I was just looking at it as a way of tracking tested versions, but perhaps there's a better (platform/command) change log for that?

:+1:

Generally speaking the supported OS versions aren't tracked in the template files.

To me alone, it adds minimal value in that the template is still broke if it's broke and version notes may not get updated (ex: it doesn't break between version X and Y so nobody is the wiser that it works with Y and not Z).

PR history and `git blame` can be helpful for history.

mjbear (approved):

PR #2069, opened 2025-03-04
added cisco ios power consumption templates

siguroot:
Adding the following templates:

cisco_ios_show_stack-power.textfsm <-- iosxe command to show power information per stack member
cisco_ios_show_power_used.textfsm <-- ios command to show used power

I have tested and used these on a power consumption report I created for my environment.  Please let me if there are any changes you think would be prudent.

Thanks

Jason

### DIFF
--- a/ntc_templates/templates/cisco_ios_show_power_used.textfsm
+++ b/ntc_templates/templates/cisco_ios_show_power_used.textfsm
@@ -1,6 +1,8 @@
-Value USED_WATTS (\d+\.\d+)
+Value WATTS (\d+\.\d+)
+Value AMPS (\d+\.\d+)
+Value VOLTS (\d+)
 
 Start
-  ^system\s+power\s+used\s+=\s+${USED_WATTS} -> Record
+  ^system\s+power\s+used\s+=\s+${WATTS}\s+Watts\s+\(${AMPS}\s+Amps\s+@\s+${VOLTS}V\)\s*$$ -> Record
   ^\s*$$
   ^. -> Error
--- a/ntc_templates/templates/cisco_ios_show_power_used.textfsm
+++ b/ntc_templates/templates/cisco_ios_show_power_used.textfsm
@@ -1,0 +1,6 @@
+Value USED_WATTS (\d+\.\d+)
+
+Start
+  ^system\s+power\s+used\s+=\s+${USED_WATTS} -> Record
+  ^\s*$$
+  ^. -> Error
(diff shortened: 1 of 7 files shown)


mjbear:
Hello @siguroot
Generally speaking it is one template per pull request (PR), although there can be exceptions.

We also need [test data](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#tests) included in this PR.

As long as you put the raw data in the correct spot, you can use [helper scripts](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#development-helper-scripts) to generate the structured data (yaml).

The test data is [necessary](https://github.com/networktocode/ntc-templates/blob/master/docs/user/faq.md#i-simply-want-to-add-my-template-to-the-project-i-do-not-want-to-add-all-of-these-tests-and-controls-can-i-just-do-so) so that we can make sure the template is backwards compatible in the future when more changes are needed.

siguroot:
Thanks for the guidance, that testing tool actually would have saved me lots of time!  I'll know to use it for next time.

I ended up making 'cisco_ios_show_power_used.textfsm' more complete, as I realized I used it for my purposes but left out some other variables that might be of interest to others.

Anything else you'd like me to do?  Perfectly willing to decouple the templates and put in two individual PR's if you want.

mjbear:
> Thanks for the guidance, that testing tool actually would have saved me lots of time! I'll know to use it for next time.

:+1: :grin:

> I ended up making 'cisco_ios_show_power_used.textfsm' more complete, as I realized I used it for my purposes but left out some other variables that might be of interest to others.

We don't need to capture all the possible data, but what you've done there looks fine.

> Anything else you'd like me to do?

I'm thinking about taking a gander at some hardware I have access to for potentially some more test data.

> Perfectly willing to decouple the templates and put in two individual PR's if you want.

I'm going to leave this decision up to the repo owners, but the plus side is it is only two templates and not twelve. :grin:

siguroot:
If you have a stacked 3850, that would be very useful to test with 'cisco_ios_show_stack-power.textfsm'

mjbear:
> If you have a stacked 3850, that would be very useful to test with 'cisco_ios_show_stack-power.textfsm'

No stacked 3850s (or 3650) here.

I could get output from a power stacked ring, but it just says `Ring` in the Stack Topology column which isn't that much different or exciting.

The couple places I looked I haven't found `show power used` so that must be some pretty old hardware (or IOS or both).

mjbear (approved):

PR #2070, opened 2025-03-07
Improving template for Huawei SmartAX: display_ont_wlan-status_0_1

evilmonkey19:
##### ISSUE TYPE

 - New Template Pull Request (Only one template at a time is allowed)
 - Bugfix Pull Request
 - Additional Testing
 - Docs Pull Request

##### SUMMARY

```

```


mjbear:
I think I'd like to swap out some of the `\s*` regex for `\s+` where we can tell in the raw output there is one or more spaces.

_(Hence the `WIP` label.)_

mjbear:
> I think I'd like to swap out some of the `\s*` regex for `\s+` where we can tell in the raw output there is one or more spaces.
> 
> _(Hence the `WIP` label.)_

* Updated some of the `\s*` to `\s+` where it fit the raw output
* Replaced tab indents with two spaces (for several rule lines)
* Changed capitalization of the first letter of words for state transitions

Good work!

mjbear (approved):

PR #1964, opened 2025-01-07
Update hp_comware_display_interface_brief.textfsm

juliencorsini:
Fixed regexp of hp_comware_display_interface_brief.textfsm for duplex value. 
Fixes https://github.com/networktocode/ntc-templates/issues/1949

PR #1977, opened 2025-01-14
Modify cisco_asa_show_route template to include VPN routes

wollmannbruno:
The modifications work on both single and multiline VPN routes.

PR #2076, opened 2025-03-12
Regenerate all hp comware disp int br test data

mjbear:
Fix for mistake in #1964 where it would appear all the test data was not regenerated.
(Only the new test data was generated and present.)


mjbear:
Actually the likely cause is PR #1977 has been around for a while and likely needs rebased.

PR #2081, opened 2025-03-18
Update Linux ip route show

evilmonkey19:
Improving the template for Linux command `ip route show`:
- Reducing the template lines
- Making template more robuts.


ntc_templates/templates/linux_ip_route_show.textfsm
mjbear, line 1: @evilmonkey19
There are several lines that somehow are duplicated in this template.
Would you please look into this?

evilmonkey19: checked 🐵✅

mjbear:
@evilmonkey19
I don't mean to be pain, but could you please add a description up toward the top where it says "No description provided"?

Bullet lists are perfectly fine for these PRs. :grinning:

So far I've figured out some of the modifications allow the `dhcp` routes to be captured instead of being excluded.

Thank you! :metal:

mjbear:
@evilmonkey19
A few updates:

* Remove duplicate pattern lines (unreachable, broadcast, local)
* Remove asterisk from broadcast (which was actually making it zero or more `t`s)
* Fix reachability and metric issue (if there's both, reachability is last)
* Escape the periods for a literal period in the IP address capture groups

@dainok
Might you be willing to review this please? :crossed_fingers:
I'd appreciate the double check with other test systems and by an expert. :grinning:

evilmonkey19:
In this link i found the case for linkdown
https://www.cyberciti.biz/faq/howto-linux-configuring-default-route-with-ipcommand/

evilmonkey19:
I haven't found yet any for unreachable/broadcast/local

It was already on the template but not on the tests :(

mjbear:
> In this link i found the case for linkdown
> https://www.cyberciti.biz/faq/howto-linux-configuring-default-route-with-ipcommand/

I didn't find the text "linkdown" on that page, but I sure did have some example output on one of my hosts! :laughing:

> I haven't found yet any for unreachable/broadcast/local
> 
> It was already on the template but not on the tests :(

There's all three examples in the [original test data](https://github.com/networktocode/ntc-templates/blob/master/tests/linux/ip_route_show/linux_ip_route_show.raw) (from Andrea) so we're set in terms of unit testing.
However, yes, I've noticed I don't have several of those instances/occurrences on my systems too. :sweat_smile:

mjbear:
> @mjbear / @evilmonkey19 to my sounds fine. I tested with my lab systems and works like a charm. (sorry for the very late response)

Thank you for reviewing this @dainok!!!
:star_struck: :grin:

mjbear (approved):

PR #2087, opened 2025-03-24
Juniper junos show system configuration database usage

jnicholson56:
Adding new file juniper_junos_show_system_configuration_database_usage.textfsm.

Send PR as draft for feedback before finalizing it.

### DIFF
--- a/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
+++ b/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
@@ -7,11 +7,11 @@ Value AVAIL_DB_SPACE (\S+)
 # Otherwise, the database usage will be displayed
 
 Start 
-  ^\s*Cannot\s*proceed,\s*database\s*is\s*currently\s*being\s*modified
   ^\s*Maximum\s*size\s*of\s*the\s*database:\s*${MAX_DB_SIZE}\s*
   ^\s*Current\s*database\s*size\s*on\s*disk:\s*${CURR_DB_SIZE}\s*
   ^\s*Actual\s*database\s*usage:\s*${ACTUAL_DB_SIZE}\s*
   ^\s*Available\s*database\s*space:\s*${AVAIL_DB_SPACE}\s* -> Record
   ^{(master|backup)}\s*$$
+  ^\s*Cannot\s*proceed,\s*database\s*is\s*currently\s*being\s*modified -> Error
   ^\s*$$
   ^. -> Error
--- a/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
+++ b/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
@@ -1,0 +1,17 @@
+Value MAX_DB_SIZE (\S+)
+Value CURR_DB_SIZE (\S+)
+Value ACTUAL_DB_SIZE (\S+)
+Value AVAIL_DB_SPACE (\S+)
+
+# If the database is being modified, the first line will be displayed
+# Otherwise, the database usage will be displayed
+
+Start 
+  ^\s*Cannot\s*proceed,\s*database\s*is\s*currently\s*being\s*modified
+  ^\s*Maximum\s*size\s*of\s*the\s*database:\s*${MAX_DB_SIZE}\s*
+  ^\s*Current\s*database\s*size\s*on\s*disk:\s*${CURR_DB_SIZE}\s*
+  ^\s*Actual\s*database\s*usage:\s*${ACTUAL_DB_SIZE}\s*
+  ^\s*Available\s*database\s*space:\s*${AVAIL_DB_SPACE}\s* -> Record
+  ^{(master|backup)}\s*$$
+  ^\s*$$
+  ^. -> Error
--- a/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
+++ b/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
@@ -3,7 +3,7 @@ Value CURR_DB_SIZE (\S+)
 Value ACTUAL_DB_SIZE (\S+)
 Value AVAIL_DB_SPACE (\S+)
 
-# If the database is being modified, the first line will be displayed
+# If the database is being modified, a "cannot proceed" message will be displayed
 # Otherwise, the database usage will be displayed
 
 Start 
--- a/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
+++ b/ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
@@ -12,6 +12,6 @@ Start
   ^\s*Actual\s+database\s+usage:\s+${ACTUAL_DB_SIZE}\s*
   ^\s*Available\s+database\s+space:\s+${AVAIL_DB_SPACE}\s* -> Record
   ^{(master|backup)}\s*$$
-  ^\s*Cannot\s+proceed,\s+database\s+is\s+currently\s+being\s+modified -> Error
+  ^\s*Cannot\s+proceed,\s+database\s+is\s+currently\s+being\s+modified
   ^\s*$$
   ^. -> Error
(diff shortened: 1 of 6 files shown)


mjbear:
@jnicholson56 
Should the megabyte (MB) unit abbreviation be captured?

Are those units always in megabytes?
(Food for though possibly.)

In the meanwhile, I do have some whitespace regex suggestions I'll start putting out there.

ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
```diff
@@ -0,0 +1,17 @@
...
+# If the database is being modified, the first line will be displayed
+# Otherwise, the database usage will be displayed
+
+Start 
+  ^\s*Maximum\s*size\s*of\s*the\s*database:\s*${MAX_DB_SIZE}\s*
+  ^\s*Current\s*database\s*size\s*on\s*disk:\s*${CURR_DB_SIZE}\s*
+  ^\s*Actual\s*database\s*usage:\s*${ACTUAL_DB_SIZE}\s*
+  ^\s*Available\s*database\s*space:\s*${AVAIL_DB_SPACE}\s* -> Record
+  ^{(master|backup)}\s*$$
+  ^\s*Cannot\s*proceed,\s*database\s*is\s*currently\s*being\s*modified -> Error
```
mjbear, line 15: Since there is at least one space in the raw cli output, I'm suggesting the whitespace regex be modified for one or more (plus sign).

To accept my suggestion, please click the "commit suggestion" button. Thank you!

```suggestion
  ^\s*Maximum\s+size\s+of\s+the\s+database:\s+${MAX_DB_SIZE}\s*
  ^\s*Current\s+database\s+size\s+on\s+disk:\s+${CURR_DB_SIZE}\s*
  ^\s*Actual\s+database\s+usage:\s+${ACTUAL_DB_SIZE}\s*
  ^\s*Available\s+database\s+space:\s+${AVAIL_DB_SPACE}\s* -> Record
  ^{(master|backup)}\s*$$
  ^\s*Cannot\s+proceed,\s+database\s+is\s+currently\s+being\s+modified -> Error
```

jnicholson56:
> @jnicholson56 ~Should the megabyte (MB) unit abbreviation be captured?~ Reworded: Should we also capture the megabyte (MB) unit abbreviation?
> 
> Are those units always in megabytes? Could those ever be in gigabytes (GB)? (Food for though possibly.)
> 
> In the meanwhile, I do have some whitespace regex suggestions I'll start putting out there.

In all the examples I have on my network (100's), they are all labeled as MB. The output is the same on both Classic and EVO as well.

Thanks for the suggestions. I had a few mistakes that I have committed as well. Hopefully tests pass this time :D

jnicholson56:
This is ready to go unless there are more suggestions.

mjbear:
> In all the examples I have on my network (100's), they are all labeled as MB. The output is the same on both Classic and EVO as well.

And now that I look at the raw cli output again, I do notice some values are more than a thousand and are still in MB units.

All good. :grinning: :sweat_smile:

jnicholson56:
Juniper is nothing if not inconsistent across the CLI :)

ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
```diff
@@ -0,0 +1,17 @@
+Value MAX_DB_SIZE (\S+)
+Value CURR_DB_SIZE (\S+)
+Value ACTUAL_DB_SIZE (\S+)
+Value AVAIL_DB_SPACE (\S+)
+
+# If the database is being modified, the first line will be displayed
+# Otherwise, the database usage will be displayed
```
mjbear, line 7: I believe this comment refers to what is currently line 15.
I'll change this comment slightly.

```suggestion
# If the database is being modified, a "cannot proceed" message will be displayed
# Otherwise, the database usage will be displayed
```

ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
```diff
@@ -0,0 +1,17 @@
...
+# If the database is being modified, a "cannot proceed" message will be displayed
+# Otherwise, the database usage will be displayed
+
+Start 
+  ^\s*Maximum\s+size\s+of\s+the\s+database:\s+${MAX_DB_SIZE}\s*
+  ^\s*Current\s+database\s+size\s+on\s+disk:\s+${CURR_DB_SIZE}\s*
+  ^\s*Actual\s+database\s+usage:\s+${ACTUAL_DB_SIZE}\s*
+  ^\s*Available\s+database\s+space:\s+${AVAIL_DB_SPACE}\s* -> Record
+  ^{(master|backup)}\s*$$
+  ^\s*Cannot\s+proceed,\s+database\s+is\s+currently\s+being\s+modified -> Error
```
mjbear, line 15: @jnicholson56
Would you by chance add raw output to this PR for an instance where "cannot proceed" is present?

Please and thank you!

jnicholson56:
Yes I can do that. It might take me a day or two to stage the conditions
required.

ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
mjbear: @jnicholson56 
In hindsight this may not work out well for the unit tests since this triggers the Error state.

As a suggestion, please test on your local machine with `invoke` first to make sure it's kosher before committing/pushing the changes up to GitHub. _(This way you should be able to save yourself the hassle of having to revert the changes if this turns out to be a bad idea.)_

* https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#testing
* https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_environment.md#invoke

mjbear: > Yes I can do that. It might take me a day or two to stage the conditions
required.

@jnicholson56
I guess we'll let this one ride as it is.

**Question:** :question:
I suppose the idea was to rather have the output not parse because it throws an error as opposed to parse and contain empty strings for each key?

jnicholson56:
Originally this template was failing when the configuration was being
updated. The configuration database gets locked and can't be polled. I
added the error as a way to let me know it failed for that reason and that
I could ignore the issue. There may be a better way to handle it but this
worked for me in my ignorance.

mjbear:
> Originally this template was failing when the configuration was being updated. The configuration database gets locked and can't be polled. I added the error as a way to let me know it failed for that reason and that I could ignore the issue.

Gotcha

> There may be a better way to handle it but this worked for me in my ignorance.

I'll make suggestion(s) and you can decide.

If the config db was locked the result would be an empty list which is easy to test for in Python.

```yaml
---
parsed_sample: []
```

(An empty list evaluates to the boolean `False` in Python.)
:partying_face: I'd go with this solution if it were me. :sweat_smile:

jnicholson56:
My original solution for the problem a few years back now. I am using
netconf to poll and the model is a bit more forgiving. If you want I can
remove this error from the template. Since this has been dangling for some
time I suspect you guys want to get it completed and pushed through.

mjbear:
> My original solution for the problem a few years back now. I am using netconf to poll and the model is a bit more forgiving. If you want I can remove this error from the template.

How about keeping the message in the template, but remove the `Error` action?
(I will submit a suggestion on GH here.)

(Ultimately since nothing matches we end up with an empty Python list which is easy to test for in the code that will use the list of parsed data.)

> Since this has been dangling for some time I suspect you guys want to get it completed and pushed through.

Things got busy -- breaking changes were held to the side for a major release and normal changes piled up a bit too. :grinning: I'm looking to merge a few to reduce our number of remaining PRs. :slightly_smiling_face:

ntc_templates/templates/juniper_junos_show_system_configuration_database_usage.textfsm
mjbear: Here's my suggestion so parsing doesn't throw an error -- though it still results in an empty list which can be easily tested for. _Best of both worlds?_

```suggestion
  ^\s*Cannot\s+proceed,\s+database\s+is\s+currently\s+being\s+modified
```

jnicholson56:
@mjbear That works for me.

mjbear:
> @mjbear That works for me.

My commit message is a bit off - it was from my testing with that line in a test output file.

I have since pushed the test output file (with the config db locked message) to this branch so we have test coverage.

Apologies for any confusion ... I should have amended my commit message. :neutral_face:

mjbear (approved):

jvanderaa:
Thank you @jnicholson56

PR #2088, opened 2025-03-25
Juniper junos show rsvp interface

jnicholson56:
adding juniper_junos_show_rsvp_interface.textfsm file


jnicholson56:
I think I worked out all of the kinks for this one.

jvanderaa (approved):

jvanderaa:
@jnicholson56 I'm just looking to compare this one with other RSVP templates. Are the keys that we have here aligned when compared to these:

https://github.com/networktocode/ntc-templates/blob/master/ntc_templates/templates/alcatel_sros_show_router_rsvp_interface.textfsm
https://github.com/networktocode/ntc-templates/blob/master/ntc_templates/templates/cisco_xr_show_rsvp_neighbors.textfsm

I think we are mostly there, and we may have some keys on the Alcatel one that may align here. I wouldn't mind at some point if the names make sense that we do a second PR upcoming for aligning all of them to the same if the keys don't make sense on the other ones. 

What this helps with is in a multi-vendor environment. By having the keys already aligned for similar commands on each vendor in a generic sense, then it is easier to compare.

mjbear (approved):

PR #2092, opened 2025-03-31
Fix XR7 parse error where ARP EVPN line exists

garymccann:
Fixes an IOS-XR version 7 issue, where a line is now included for ARP EVPN

### DIFF
--- a/ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
+++ b/ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
@@ -35,10 +35,10 @@ Start
   ^\s+Outgoing\s+access\s+list\s+is\s+${OUT_ACL}
   ^\s+Inbound\s+(?:common)?\s+access\s+list\s+is\s+${IN_ACL}
   ^\s+Proxy.*$$
-  ^\s+ARP EVPN.*$$
-  ^\s+ICMP redirects.*$$
-  ^\s+ICMP unreachables.*$$
-  ^\s+ICMP mask.*$$
+  ^\s+ARP\s+EVPN.*$$
+  ^\s+ICMP\s+redirects.*$$
+  ^\s+ICMP\s+unreachables.*$$
+  ^\s+ICMP\s+mask.*$$
   ^\s+Table.*$$
   ^\s+mLACP.*$$
   ^IP\s+unicast\s+RPF\s+check\s+is\s+${RPF_CHECK}
--- a/ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
+++ b/ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
@@ -35,6 +35,7 @@ Start
   ^\s+Outgoing\s+access\s+list\s+is\s+${OUT_ACL}
   ^\s+Inbound\s+(?:common)?\s+access\s+list\s+is\s+${IN_ACL}
   ^\s+Proxy.*$$
+  ^\s+ARP EVPN.*$$
   ^\s+ICMP redirects.*$$
   ^\s+ICMP unreachables.*$$
   ^\s+ICMP mask.*$$
--- a/tests/cisco_xr/show_ipv4_interface/cisco_xr_show_ipv4_interface6.raw
+++ b/tests/cisco_xr/show_ipv4_interface/cisco_xr_show_ipv4_interface6.raw
@@ -1,0 +1,33 @@
+Mon Mar 31 18:49:52.390 UTC
+Bundle-Ether123 is Down, ipv4 protocol is Down 
+  Vrf is default (vrfid 0x60000000)
+  Internet protocol processing disabled
+MgmtEth0/RP0/CPU0/0 is Up, ipv4 protocol is Up 
+  Vrf is MGMT (vrfid 0x60000001)
+  Internet address is 192.168.254.130/24
+  MTU is 1514 (1500 is available to IP)
+  Helper address is not set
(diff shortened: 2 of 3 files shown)


mjbear:
Hi @garymccann 
I made a couple of adjustments.

* Regenerated structured data so the keys are in alphabetical order
* Replaced literal spaces with regexes
* Replaced a few singular whitespace regexes with a more flexible whitespace regex (one or more)

Thank you!

garymccann:
hey @mjbear - can you please break these adjustments into their own PR - i do not like touching files unrelated to the feature, increasing the blast radius and while i don't see anything wrong with changing the order, it may confuse folk who are using the raw data, since it does not follow any alphabetical ordering. I will revert these PR changes.

mjbear:
> hey @mjbear - can you please break these adjustments into their own PR - i do not like touching files unrelated to the feature, increasing the blast radius and while i don't see anything wrong with changing the order, it may confuse folk who are using the raw data, since it does not follow any alphabetical ordering. I will revert these PR changes.

@garymccann
Sure thing.
I'll start canceling them if you haven't already.

There might be one or two items left for us to include on this PR, but we'll see as I take a gander now.

ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
mjbear, line 1: Straight forward changes here.
Thank you for adding new test data to accompany the existing test data!

mjbear (approved):

ntc_templates/templates/cisco_xr_show_ipv4_interface.textfsm
```diff
@@ -35,6 +35,7 @@ Start
   ^\s+Outgoing\s+access\s+list\s+is\s+${OUT_ACL}
   ^\s+Inbound\s+(?:common)?\s+access\s+list\s+is\s+${IN_ACL}
   ^\s+Proxy.*$$
+  ^\s+ARP EVPN.*$$
   ^\s+ICMP redirects.*$$
   ^\s+ICMP unreachables.*$$
   ^\s+ICMP mask.*$$
```
mjbear, line 41: Could we please replace literal whitespaces with regex?

(We can leave the other regexes alone on this PR and I can follow up with another PR for other things as mentioned in earlier discussion.)

```suggestion
  ^\s+ARP\s+EVPN.*$$
  ^\s+ICMP\s+redirects.*$$
  ^\s+ICMP\s+unreachables.*$$
  ^\s+ICMP\s+mask.*$$
```

garymccann: These two lines are not equivalent. \s+ matches one or more whitespace characters, where the literal whitespace is just one. 
I checked performance of literal, \s and \s+ - all 3 take 14 processing steps (regex101.com) so there is no processing gain to be had from any change in implementation. 

My personal preference is to keep the literal whitespace, since it reads most similar to the CLI output, however, given there is no performance penalties, i am agnostic. 

Please let me know your thoughts, main thing here is to get this merged into a pypy release for my use in production uses.

mjbear: I'm looking at this through a future-proofing lens -- balancing the flexibility of the template and its regexes to hopefully reduce the maintenance burden in the long term.

I certainly hope that Cisco doesn't add extra accidental whitespace, but we've seen other vendors make somewhat haphazard changes (ex: spacing, capitalization, spelling).

> [!NOTE]
> Since I already have it on my mind to tackle putting the structured data alphabetic ordering in a different PR, this literal whitespace could be something that happens there.

_My gut feeling is for me to hold off with the whitespace regex, etc until a later PR (as in not for this PR)._
I welcome feedback from the other maintainers if they have thoughts on it.

mjbear: @itdependsnetworks @jmcgill298 @jvanderaa
:wave:
Thoughts on moving forward with merging this as it is?

mjbear: > These lines are not equivalent. \s+ matches one or more whitespace characters, where the literal whitespace is just one.

True

> I checked performance of literal, \s and \s+ - all 3 take 14 processing steps (regex101.com) so there is no processing gain to be had from any change in implementation.

Yes, though in terms of future proofing the whitespace matching there is a possible longer-term maintenance avoidance "gain".

> My personal preference is to keep the literal whitespace, since it reads most similar to the CLI output, however, given there is no performance penalties, i am agnostic.

Seeing as you're agnostic to the whitespace, my personal stance is to use the **regexes**.
 
> Please let me know your thoughts, main thing here is to get this merged into a pypy release for my use in production uses.

Yep, we have a few PRs to review and merge in.

I don't have any details on the next pypi release of ntc-templates.

Once this one is merged in, I'll spin up a PR to regenerate the test data keys in alphabetic order which helps us "spot changes" in the longer term. :grin:

jmcgill298: We do prefer `\s+` so that an accidental extra whitespace doesn't break the template

mjbear: > We do prefer `\s+` so that an accidental extra whitespace doesn't break the template

I'll pull in the whitespace suggestion, but leave "alphabetizing the test data keys" to a second PR.

mjbear (approved):

PR #2093, opened 2025-04-01
Add Cisco IOS show vlan group

forrejam:
Add support for "show vlan group" on cisco IOS

### DIFF
--- a/ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
+++ b/ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
@@ -1,0 +1,6 @@
+Value Required VLAN_GROUP_NAME (\S+)
+Value VLAN_ID (\d+)
+
+Start
+  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_ID}$$ -> Record
+  ^. -> Error
--- a/ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
+++ b/ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
@@ -1,6 +1,6 @@
 Value Required VLAN_GROUP_NAME (\S+)
-Value VLAN_IDS (\d+.*)
+Value VLAN_ID (\d+.*)
 
 Start
-  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_IDS}$$ -> Record
+  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_ID}$$ -> Record
   ^. -> Error
--- a/ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
+++ b/ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
@@ -1,6 +1,6 @@
 Value Required VLAN_GROUP_NAME (\S+)
-Value VLAN_ID (\d+)
+Value VLAN_IDS (\d+.*)
 
 Start
-  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_ID}$$ -> Record
+  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_IDS}$$ -> Record
   ^. -> Error
--- a/ntc_templates/templates/index
+++ b/ntc_templates/templates/index
@@ -339,7 +339,7 @@ cisco_ios_show_policy-map.textfsm, .*, cisco_ios, sh[[ow]] pol[[icy-map]]
 cisco_ios_show_power_used.textfsm, .*, cisco_ios, sh[[ow]] pow[[er]] u[[sed]]
 cisco_ios_show_redundancy.textfsm, .*, cisco_ios, sh[[ow]] redu[[ndancy]]
 cisco_ios_show_snmp_group.textfsm, .*, cisco_ios, sh[[ow]] snm[[p]] g[[roup]]
-cisco_ios_show_vlan_group.textfsm, .*, cisco_ios, sh[[ow]] vlan group
+cisco_ios_show_vlan_group.textfsm, .*, cisco_ios, sh[[ow]] vlan g[[roup]]
 cisco_ios_show_vrf_detail.textfsm, .*, cisco_ios, sh[[ow]] vrf detail
 cisco_ios_show_vrrp_brief.textfsm, .*, cisco_ios, sh[[ow]] vrr[[p]] b[[rief]]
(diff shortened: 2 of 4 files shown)


forrejam:
Please let me know if there is a better way to handle vlan output like this in textfsm

`vlan group MULTI_TST_1 : 761-763, 997, 1160, 1180`


```yaml
- vlan_group_name: "MULTI_VLAN_TEST_2"
    vlan_ids: "761-763, 197, 200, 208"
```

mjbear:
> Please let me know if there is a better way to handle vlan output like this in textfsm
> 
> `vlan group MULTI_VLAN_TEST_2 : 761-763, 997, 1160, 1180`
> 
> ```yaml
> - vlan_group_name: "MULTI_VLAN_TEST_2"
>   vlan_ids: "761-763, 197, 200, 208"
> ```

Since those VLAN IDs are all on one line, the string capture as you have it is probably the simplest route.

If they were on different lines we could have had the chance to use `List`, but that's not the case here.

ntc_templates/templates/index
```diff
@@ -339,6 +339,7 @@ cisco_ios_show_policy-map.textfsm, .*, cisco_ios, sh[[ow]] pol[[icy-map]]
 cisco_ios_show_power_used.textfsm, .*, cisco_ios, sh[[ow]] pow[[er]] u[[sed]]
 cisco_ios_show_redundancy.textfsm, .*, cisco_ios, sh[[ow]] redu[[ndancy]]
 cisco_ios_show_snmp_group.textfsm, .*, cisco_ios, sh[[ow]] snm[[p]] g[[roup]]
+cisco_ios_show_vlan_group.textfsm, .*, cisco_ios, sh[[ow]] vlan group
```
mjbear, line 342: While the entire `vlan` word is required (since the plural form is also possible), `group` can be shortened to just the letter g.

```suggestion
cisco_ios_show_vlan_group.textfsm, .*, cisco_ios, sh[[ow]] vlan g[[roup]]
```

ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
```diff
@@ -0,0 +1,6 @@
+Value Required VLAN_GROUP_NAME (\S+)
+Value VLAN_IDS (\d+.*)
+
+Start
+  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_IDS}$$ -> Record
```
mjbear, line 5: I thought about this for a bit ... we've normalized/standardized on `VLAN_ID` in other templates. For the raw output here there can be one or more VLAN IDs, but that holds true for other templates where `VLAN_ID` is used ([cisco_ios example](https://github.com/networktocode/ntc-templates/blob/master/tests/cisco_ios/show_interfaces_status/cisco_ios_show_interfaces_status_pvlan.raw#L5)).

```suggestion
Value Required VLAN_GROUP_NAME (\S+)
Value VLAN_ID (\d+.*)

Start
  ^vlan\s+group\s+${VLAN_GROUP_NAME}\s+:\s+${VLAN_ID}$$ -> Record
```

forrejam: Yep that's fine. The only thing I would say about the use of VLAN_ID in other templates is usually you can rely on it being a single vlan (\d+) so I thought using a different identifier would prompt the user that it will need more post processing (which is what I'm doing in production) in the event of a vlan range .

mjbear: > so I thought using a different identifier would prompt the user that it will need more post processing (which is what I'm doing in production) in the event of a vlan range.

You have a good point there to differentiate and call out the need for post processing.
We can certainly get some input from others.

ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
```diff
@@ -0,0 +1,6 @@
+Value Required VLAN_GROUP_NAME (\S+)
+Value VLAN_ID (\d+.*)
```
jmcgill298, line 2: It would be preferable to make this a List, and call it VLAN_IDS

jmcgill298 (changes requested):

ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
mjbear: > It would be preferable to make this a List, and call it VLAN_IDS

Jacob wondered if this output might wrap.

From my test, the vlan group list ends up on a single long line and does not have newlines injected.

```
switch#show vlan group
vlan group TEST : 2, 4, 6, 8, 10-15, 20, 22, 24, 26, 28, 30, 99-199, 201-299, 301-399, 401-499, 501-599, 601-699, 701-799, 801-899, 901-999, 1001-1010, 1012-1013, 1015-1016, 1018-1019, 1021-1022, 1024-1027, 1029-1032, 1034-1035
```

A `List` data type does not seem to be needed.
Certainly this leaves post-processing work to split the string into individual VLAN IDs, but that's how it does with TextFSM.

ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
mjbear: The more I think about it, I ask "how often is a vlan group a single VLAN ID?"
Probably less frequently a single than multiple VLAN IDs.

That IOS template I linked to where there were multiple IDs is probably more an "exception" for that particular command output.

In this use case, it seems more likely that a [vlan group](https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst3850/software/release/37e/consolidated_guide/b_37e_consolidated_3850_cg/b_37e_consolidated_3850_cg_chapter_01101001.pdf#unique_2715) ***will*** contain multiple VLAN IDs in practice. While the parsed capture group will be a string and not multiple separate values, naming the capture group with a plural indicates the parsed data _likely_ contains more than one VLAN ID.

I'm in agreement with you @forrejam, my thought is it may be best to make it plural `VLAN_IDS` for clarity/post-processing reasons.

Thank you for being patient and allowing me time to revisit this. :sweat_smile:

jvanderaa:
Everything all set for this one @jmcgill298 / @mjbear ?

jmcgill298 (approved):

ntc_templates/templates/cisco_ios_show_vlan_group.textfsm
mjbear: > No problem @mjbear, either way will work for me, but I think `VLAN_IDS` is more fitting in this instance. Perhaps `VLAN_IDS` can become convention for other similar contexts.

I agree.

I got busy and didn't bump some updates into this PR. It's not bad this got merged this way but now that it's merged I'll probably leave it alone unless somebody feels strongly. :relaxed:

PR #2094, opened 2025-04-04
Fix space issue for command date, seen on XR6.6.3

garymccann:
During work on Nautobot Device Onboarding, this template was failing due to the double space allocated to the date. 
11 April was OK, 4 April was not, due to the space layout. 

The regex now takes account of multiple spaces and the unit tests increased to reflect this. 

Developer Note: This is one of the few commands to match on the command date. It may be possible to remove this match and converge on a common solution within the XR commands. Will open an Issue to discuss

### DIFF
--- a/ntc_templates/templates/cisco_xr_show_running-config_hostname.textfsm
+++ b/ntc_templates/templates/cisco_xr_show_running-config_hostname.textfsm
@@ -2,7 +2,7 @@ Value HOSTNAME (\S+)
 
 Start
   # Capture date output by XR by 'show run hostname' first
-  ^\S+ \S+ \S+ .*$$
+  ^\S+\s+\S+\s+\d+\s+\d+:\d+:\d+\.\d+\s+\S+$$
   # Capture hostname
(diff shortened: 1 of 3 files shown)


ntc_templates/templates/cisco_xr_show_running-config_hostname.textfsm
```diff
@@ -2,7 +2,7 @@ Value HOSTNAME (\S+)
 
 Start
   # Capture date output by XR by 'show run hostname' first
-  ^\S+ \S+ \S+ .*$$
+  ^\S+\s+\S+\s+\d+\s+\d+:\d+:\d+\.\d+\s+\S+$$
   # Capture hostname
   ^hostname ${HOSTNAME}$$
```
mjbear, line 7: Let's take care of another literal space and replace it with a regex too.
Minor change that I'll send in.

```suggestion
  ^hostname\s+${HOSTNAME}$$
```

mjbear (approved):
Thank you!

itdependsnetworks (approved):

PR #2095, opened 2025-04-05
Juniper junos show rsvp interface

jnicholson56:
I have updated the variables for juniper_junos_show_rsvp_interface to match variables for alcatel_sros/show_router_rsvp_interface where possible. 

The cisco_xr equivalent template shows a smaller subset of info compared to these two. The proper cisco XR command does not have a template in the repository.


mjbear:
Although this template was first merged into the repo, technically it is a breaking change.

That's sort of beside the fact in a way though since with breaking changes queuing up there will be a major version release around the corner.

ntc_templates/templates/juniper_junos_show_rsvp_interface.textfsm
```diff
@@ -1,11 +1,11 @@
...
 Value Required INTERFACE (\S+)
-Value STATE (\S+)
+Value OPER_STATE (\S+)
 Value ACTIVE_RESV (\S+)
 Value SUBSCRIPTION (\S+)
-Value STATIC_BW (\S+)
+Value TOTAL_BW (\S+)
 Value AVAIL_BW (\S+)
-Value RESERVED_BW (\S+)
+Value RESV_BW (\S+)
```
mjbear, line 8: When it comes to reserved bandwidth, RESERVED_BW is better in my opinion.

_This may mean we could consider a breaking change against the Alcatel RSVP template to somewhat normalize._

This is a thought and let's see what others think. Thank you!

mjbear: @jvanderaa @jmcgill298 @itdependsnetworks
Thoughts on using `RESERVED_BW` in this Juniper template and [I can] update the Alcatel template that has `RESV_BW`?

Thank you! :sweat_smile:

jnicholson56:
My juniper templates are new and the chance of someone using them is next
to nothing. How would a change in an existing template affect users that
may be relying on it?

mjbear:
> My juniper templates are new and the chance of someone using them is next to nothing. How would a change in an existing template affect users that may be relying on it?

While I'm not saying the capture group must be `RESERVED_BW`, I do think the **reserved** word spelled out is a better choice for clarity.

We all have no clear way of knowing if someone in the community happened to use the RSVP template.

In checking, this template was first merged in after release 7.8.0 and it's only been out there for about three weeks. **I'll let someone else make the call on this**, but I can say that whether it is considered a "breaking change" doesn't exclude it from being merged or anything major.

jnicholson56:
Ahh ok maybe it doesn't matter

ntc_templates/templates/juniper_junos_show_rsvp_interface.textfsm
mjbear: To the wider group, what are the thoughts on:
`RESERVED_BW`

And similarly:
`AVAILABLE_BW`

If this is agreeable and we change this template ... then I'd work up a separate PR for the Alcatel RSVP template to align it. :relaxed:

mjbear: Thoughts?

`RESERVED_BW`
`AVAILABLE_BW`
`ACTIVE_RESERVATION`

Or do we leave things well enough alone?
(As in leave this PR as it is with the abbreviated `RESV_BW`.)

**Edit:** The plus side is leaving things as this PR was submitted means the Alcatel template doesn't need any modifications. :sweat_smile:

jvanderaa: I like keeping things and then let's plan to get this in once it is ready with other breaking changes.

mjbear (approved):

PR #2098, opened 2025-04-08
Update cisco_xr_ping.textfsm for trailing commas

jorlandobr:
Updated cisco_xr_ping.textfsm.

Found that some versions of Cisco XR ommit the comma after the destination address

`Sending 5, 100-byte ICMP Echos to 10.245.179.18 timeout is 2 seconds`

instead of

`Sending 5, 100-byte ICMP Echos to 10.245.179.18, timeout is 2 seconds`

Corrected template and added test case.


mjbear:
@jorlandobr
I worked up a few enhancements and while I'm at it I may as well push to your branch.

One of them includes using the helper scripts to generate the structured test output with keys in alphabetical order. _(While we're working on this PR we may as well do that.)_

I tested them so have validated the template still functions after the modifications, but if for some reason you disagree with something I push the commit can be reverted.

Please review - thank you!

jorlandobr:
@mjbear , no objections on the improvements. 🙂

mjbear (approved):

PR #2122, opened 2025-05-04
Add Linux iwconfig

evilmonkey19:
New template.


ntc_templates/templates/index
```diff
@@ -806,6 +806,7 @@ linux_pct_config_1.textfsm, .*, linux, pct config \d+
 linux_ip_vrf_show.textfsm, .*, linux, ip v[[rf]] [[show]]
 linux_qm_config_1.textfsm, .*, linux, qm c[[onfig]] \d+
 linux_pveversion.textfsm, .*, linux, pveversion
+linux_iwconfig.textfsm, .*, linux, iwc[[onfig]]
```
mjbear, line 809: @evilmonkey19
Can `iwconfig` really be shorted to `iwc` and still work?

If I may ask, what distro and version is being tested against?

I expect a shell alias would be required for that short version to work.
Thoughts?

evilmonkey19: You can install it using `sudo apt install wireless-tools`. I have tested in both cases against Linux Ubuntu 24. Also, this cannot be shortened as it throws the following
imagen

mjbear: > You can install it using `sudo apt install wireless-tools`. I have tested in both cases against Linux Ubuntu 24.

:+1:

> Also, this cannot be shortened as it throws the following imagen

I do believe the index needs a tweak since the command cannot be shortened.
I'll submit a suggestion.

mjbear, line 809: Change command regex so no characters are optional since the fully command has to be specified at the GNU/Linux command line.

Thank you!

```suggestion
linux_iwconfig.textfsm, .*, linux, iwconfig
```

ntc_templates/templates/linux_iwconfig.textfsm
```diff
@@ -0,0 +1,40 @@
...
+Value INVALID_MISC (\w+)
+Value MISSED_BEACON (\w+)
+
+
+Start
+  ^\s*w -> Continue.Record
+  ^\s*${INTERFACE}\s+no\s*wireless\s*extensions\.$$ -> Clear
+  ^\s*${INTERFACE}\s+(IEEE\s+802\.11\w*|unassociated)(?:\s+ESSID:off\/any)?(?:\s+Nickname:\"${NICKNAME}\")?\s*$$
+  ^\s*${INTERFACE}\s+(IEEE\s+802\.11\w*|unassociated)(?:\s+ESSID:\"${ESSID}\"\s+)?(?:\s*Nickname:\"${NICKNAME}\")?\s*$$
+  ^\s*${INTERFACE}\s+IEEE\s+802\.11\s+Mode:\s*${MODE}\s*Frequency:\s*${FREQUENCY}\s+Tx-Power=${TX_POWER}\s*$$
```
mjbear, line 31: @evilmonkey19 Would you be able to provide test data that covers this `802.11 Mode:` regex?

tests/linux/iwconfig/linux_iwconfig.raw
```diff
@@ -0,0 +1,64 @@
...
+          Bit Rate=200 Mb/s   Tx-Power=31 dBm   
+          Retry short limit:7   RTS thr:off   Fragment thr:off
+          Encryption key:off
+          Power Management:on
+          Link Quality=62/70  Signal level=-48 dBm  
+          Rx invalid nwid:0  Rx invalid crypt:0  Rx invalid frag:0
+          Tx excessive retries:7  Invalid misc:0   Missed beacon:0
+
+wlan3     unassociated  Nickname:"<WIFI@REALTEK>"
+          Mode:Managed  Frequency=5.18 GHz  Access Point: Not-Associated
```
mjbear, line 31: `Frequency=5.18 GHz`

Was this equal sign `=` instead of colon `:` an accidental modification?

tests/linux/iwconfig/linux_iwconfig.raw
```diff
@@ -0,0 +1,64 @@
+lo        no wireless extensions.
+
+eth0      no wireless extensions.
+
+wlan0     IEEE 802.11  ESSID:"My_home"  
+          Mode:Managed  Frequency:5.18 GHz  Access Point: AA:AA:AA:AA:AA:AA
```
mjbear, line 6: The only "label" that has a space after the colon (or equal sign) separator is `Access Point`.

1. Do we have test data that supports the existence of spaces after the colon/equals?

1. Should we remove some of those fluff whitespace regexes?

ntc_templates/templates/linux_iwconfig.textfsm
mjbear: > @evilmonkey19 Would you be able to provide test data that covers this `802.11 Mode:` regex?

Hello @evilmonkey19, by chance did you see this question?

Thank you!

tests/linux/iwconfig/linux_iwconfig.raw
evilmonkey19: No this change is not accidental it is on purpose! I have seen this in output using raspberry pi and some external wifi antennas :)

tests/linux/iwconfig/linux_iwconfig.raw
evilmonkey19: Perhaps we can reach a middle ground here. Just to be on the safe zone we can use a \s* instead of \s+ so if it exists the whitespace (that can happen) it works and also if it doesn't. Deal :)?

mjbear: > Perhaps we can reach a middle ground here. Just to be on the safe zone we can use a \s* instead of \s+ so if it exists the whitespace (that can happen) it works and also if it doesn't. Deal :)?

:grinning:
We have the tradeoff regexes in place currently.

I'm not opposed to using that whitespace regex in the long term.

ntc_templates/templates/linux_iwconfig.textfsm
mjbear: @evilmonkey19
(I've been busy too.)
When you have a chance would you get test data for covering `802.11 Mode:`?

Thank you! :relaxed:

evilmonkey19: Probably we can take it out. Sorry for the hurdle. i remember I saw it somewhere but I cannot find it again. I can open a new PR whenever I see it again :)

PR #2129, opened 2025-05-13
Cisco IOS-XE additional transceiver data (CURRENT)

paineta:
Existing template did not properly parse the additional "current" value in the "show interface transceiver" output on Cisco IOS-XE.  I added support to collect the Current value or leave it blank, as to not break legacy output, and I also fleshed out the template to include an Error directive.


ntc_templates/templates/cisco_ios_show_interface_transceiver.textfsm
```diff
@@ -1,11 +1,14 @@
...
 Value RX_PWR (\S+)
 
 Start
-  ^${INTERFACE}\s+${TEMPERATURE}\s+${VOLTAGE}\s+${TX_PWR}\s+${RX_PWR}\s+ -> Record
+  ^${INTERFACE}\s+${TEMPERATURE}\s+${VOLTAGE}\s+${CURRENT}?\s+${TX_PWR}\s+${RX_PWR}\s? -> Record
   # Capture time-stamp if vty line has command time-stamping turned on
   ^Load\s+for\s+
   ^Time\s+source\s+is
+  ^(If|\+|NA|mA|\s|Port|\-)
+  ^. -> Error
```
mjbear, line 14: Absolutely like the addition of the Error directive.

The `\s` has the possibility to match too many lines in my opinion.



```suggestion
  ^(\s+Optical){2}
  ^\s+Temperature\s+Voltage(\s+Current)?(\s+[TR]x\s+Power){2}
  ^Port\s+\(Celsius\)\s+\(Volts\)(\s+\(mA\))?\s+(\(dBm\)\s*){2}
  ^(If\s+device|\+\+|NA|mA|---+)
  ^. -> Error
```

ntc_templates/templates/cisco_ios_show_interface_transceiver.textfsm
```diff
@@ -1,11 +1,14 @@
 Value INTERFACE (\w+\d+\/\S+)
 Value TEMPERATURE (\d+.\d+)
 Value VOLTAGE (\d+.\d+)
+Value CURRENT (\d+.\d+)
 Value TX_PWR (\S+)
 Value RX_PWR (\S+)
 
 Start
-  ^${INTERFACE}\s+${TEMPERATURE}\s+${VOLTAGE}\s+${TX_PWR}\s+${RX_PWR}\s+ -> Record
+  ^${INTERFACE}\s+${TEMPERATURE}\s+${VOLTAGE}\s+${CURRENT}?\s+${TX_PWR}\s+${RX_PWR}\s? -> Record
```
mjbear, line 9: Just in case I'd like to see `\s*` used that way we can have zero or more whitespace. (That is zero, one, two, etc.)

I'm going to push this one in as it is something regularly seen in the templates. Thank you.

```suggestion
  ^${INTERFACE}\s+${TEMPERATURE}\s+${VOLTAGE}\s+${CURRENT}?\s+${TX_PWR}\s+${RX_PWR}\s* -> Record
```

ntc_templates/templates/cisco_ios_show_interface_transceiver.textfsm
paineta: That seems reasonable.  I guess related to this, I've found a couple more use cases where the output doesn't produce a table, but does produce messages.  I can add those expressions for those cases as well, but does there need to "tests" files for those scenarios as well?

e.g.
^No transceiver present

mjbear: > That seems reasonable. I guess related to this, I've found a couple more use cases where the output doesn't produce a table, but does produce messages. I can add those expressions for those cases as well, but does there need to have "tests" files for those scenarios as well?

Excellent question.

I'm not sure what has been done in the past for this.

> e.g. ^No transceiver present

I'd say for now call out the patterns you're seeing in the comments here and we can get more eyes on it and ask for feedback whether we'd want test files or not.

Thank you!

ntc_templates/templates/cisco_ios_show_interface_transceiver.textfsm
```diff
@@ -1,11 +1,19 @@
...
+  ^${INTERFACE}\s+${TEMPERATURE}\s+${VOLTAGE}\s+${CURRENT}?\s+${TX_PWR}\s+${RX_PWR}\s* -> Record
   # Capture time-stamp if vty line has command time-stamping turned on
   ^Load\s+for\s+
   ^Time\s+source\s+is
+  ^(\s+Optical){2}
+  ^\s+Temperature\s+Voltage(\s+Current)?(\s+[TR]x\s+Power){2}
+  ^Port\s+\(Celsius\)\s+\(Volts\)(\s+\(mA\))?\s+(\(dBm\)\s*){2}
+  ^(If\s+device|\+\+|NA|mA|---+)
+  ^No transceiver present
+  ^Diagnostic Monitoring is not implemented\.
```
mjbear, line 18: I see the error messages are in here -- all good.

Swapping regexes in for literal spaces.

```suggestion
  ^No\s+transceiver\s+present
  ^Diagnostic\s+Monitoring\s+is\s+not\s+implemented\.
```

mjbear: That commit message reads pretty badly in hindsight.

_For clarification I swapped literal whitespace for regexes, not the other way around._

mjbear:
> Since the message output from the networking device would be string literals, I feel like matching (a space) with regex adds unnecessary ambiguity.

I can see where you're coming from.

While not an absolute, there's a general consensus going back years to flexibly match white space so the possibility of breakage on vendor white space changes is minimized. This minimizes the possibility for issues and maintenance on the templates (or at least that's the hope! :relaxed:).

For those couple error messages, tucking `\s+` for a literal space doesn't make them terribly long in line length and the convention with whitespace regexes is widespread across the templates I've reviewed over several years. **This is an open discussion and I wouldn't want it either way. :slightly_smiling_face:**

I'd like to leave this up for discussion so some templating veterans can comment here.

Thank you for the contributions and discussion @paineta!

paineta:
No problem.  Thanks.

jvanderaa (approved):

PR #2139, opened 2025-06-09
Adding new template for `Huawei VRP`: `display eth-trunk`

evilmonkey19:
It is quite a difficult template... Any suggestions?


mjbear:
> It is quite a difficult template...

Ah, it's probably the existence of the Partner section, right?

> Any suggestions?

(I'm not being cheeky here)
Could you point me at what you'd like looked at?
Thank you!

ntc_templates/templates/huawei_vrp_display_eth-trunk.textfsm
```diff
@@ -0,0 +1,51 @@
+Value Filldown ETH_TRUNK (\S+)
+Value Filldown WORKING_MODE (\S+)
+Value Filldown HASH_ARITHMETIC (\S.+\S)
+Value Filldown LEAST_ACTIVE_LINKNUMBER (\d+)
+Value Filldown MAX_ACTIVE_LINKNUMBER (\d+)
+Value Filldown MAX_BANDWIDTH_AFFECTED_LINKNUMBER (\d+)
+Value Filldown OPERATE_STATUS (\S+)
+Value Filldown NUMBER_OF_UP_PORT_IN_TRUNK (\d+)
+Value Filldown LOCAL (\S+|\S.+\S)
```
mjbear: This `LOCAL` capture group doesn't show up in the rules.
Remove it?

```suggestion
```

mjbear:
> Basically i'm refering is the resulting parsed data structure is correct or not. To me it looks like but i haven't come across similar templates like this one.

The separate partner section lands us with data that has GigabitEthernet0/0/11 and 12 showing up more than once. Since the sections are separate in the output, TextFSM doesn't have the capability to combine them (so sometimes output ends up with some duplicate keys/data :grinning::man_shrugging:).

When thinking about consumers of the structured data, the data that doesn't have a `partner_system_id` are the local ones so that can be used to differentiate the sections (local vs partner).

ntc_templates/templates/huawei_vrp_display_eth-trunk.textfsm
mjbear: > This `LOCAL` capture group doesn't show up in the rules. Remove it?

Following up via Slack to ask for feedback

tests/huawei_vrp/display_eth-trunk/huawei_vrp_display_eth-trunk.yml
mjbear, line 1: I'd say this result looks about right.

We'd expect some repetition of data per member interface.

mjbear (approved):

PR #2140, opened 2025-06-09
Improve Huawei VRP display interface brief

evilmonkey19:
~Avoid capturing none expected values~

**Edit:**
Remove `^(\w+):\s+\S+` since it is loose and can match several lines (unexpectedly or unwanted).
Instead explicitly match line syntax.

### DIFF
--- a/ntc_templates/templates/huawei_vrp_display_interface_brief.textfsm
+++ b/ntc_templates/templates/huawei_vrp_display_interface_brief.textfsm
@@ -9,10 +9,20 @@ Value OUTERRORS (\d+)
 
 Start
   ^\s*${INTERFACE}\s+${PHY}\s+${PROTOCOL}\s+${INUTI}\s+${OUTUTI}\s+${INERRORS}\s+${OUTERRORS} -> Record
-  ^PHY:\s+Physical
-  ^(?:\*|\^|\#)down:
-  ^\(\w+\):\s+\S+
-  ^InUti/OutUti:
-  ^Interface\s+PHY\s+Protocol\s+InUti\s+OutUti\s+inErrors\s+outErrors\s*$$
-  ^\s*$$
+  ^\s*PHY:\s+Physical\s*$$
+  ^\s*\*down:\s+administratively\s+down\s*$$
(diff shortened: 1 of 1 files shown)


ntc_templates/templates/huawei_vrp_display_interface_brief.textfsm
mjbear, line 1: Ah, all of the added/modified lines are simply to match the line but not capture any values.

> Avoid capturing none expected values

@evilmonkey19 Based on the above quote, what was being captured that shouldn't have been?

evilmonkey19: For example the ^\(\w+\):\s+\S+ may induce to errors. I do it to force that any unexpected value throughs an error just in case.

mjbear: Message sent via Slack to ask for feedback

evilmonkey19: Just in case, I like to force that the values are the ones we already have data over it and nothing else. Imagine that Huawei in a new update introduces a new value like S-Trunk. I might work but the output might be wrong. Nevertheless, either ways is fine, just a personal opinion here.

mjbear: > Just in case, I like to force that the values are the ones we already have data over it and nothing else. Imagine that Huawei in a new update introduces a new value like S-Trunk. I might work but the output might be wrong. Nevertheless, either ways is fine, just a personal opinion here.

I understand you -- thank you for the explanation.

This clarifies the description _"Avoid capturing none expected values"_ at the top of the PR.
All good now.

mjbear (approved):

ntc_templates/templates/huawei_vrp_display_interface_brief.textfsm
evilmonkey19: Thanks! Sometimes I have clear on my mind but it is hard to express hahahaha

PR #2144, opened 2025-06-15
Fix broken uppercase capture group test

mjbear:
Fix broken uppercase capture group test.

In short, the regex function and flag usage was wrong. **The multiline flag needed specified at regex compile time.** This caused the capture group regex to not return a list of capture groups and instead an empty list. Running `isupper()` against that passed because nothing was lowercase. :face_exhaling:

When I developed the tests I did not use the compiled version and later changed it to that. Tests continued passing so unfortunately I didn't recognize the test was broken. Ugh. :confounded:

Then #2139 comes along and I decide to begin writing a unit test to look for defined, but unused capture groups in rules. *And voila things aren't acting as expected, ooof.*

> [!IMPORTANT]
> Python regex flags such as multiline need to be specified in the compiled regex and not afterwards.


mjbear:
When the `re` flag was specified in the wrong location it resulted in an empty capture group list which when looped over results in no items to assert against (so automatic "pass").

The negative test may be marginally useful for regex changes -- though it would not have caught the empty capture group list.

mjbear:
> When the `re` flag was specified in the wrong location it resulted in an empty capture group list which when looped over results in no items to assert against (so automatic "pass").
> 
> The negative test may be marginally useful for regex changes -- though it would not have caught the empty capture group list.

I'm going to keep the "bogus" test *.textfsm template file in this PR since I'll also leverage it in a test that looks for "unused but defined" capture groups.

ntc_templates/templates/bogus_for_testing.textfsm
```diff
@@ -0,0 +1,26 @@
...
+  ^${language},${Creator},${year_created},${level},${object_oriented},?$$ -> Record
+  ^${language},${Creator},${year_created},,${object_oriented},?$$ -> Record
+  ^#$$
+  ^. -> Error
+
+
+# Mock output below was pulled together by referencing Wikipedia,
+#  opinons can vary and this is just test data
+
+#Python,Guido van Rossum,1991,high-level,yes
```
mjbear: * Fix spelling typo
* Add header to mock data (although the template doesn't run against it)

```suggestion
# Mock output below was pulled together by referencing Wikipedia,
#  opinions can vary and this is just test data

##Language,Creator,Year Created,Level,Object Oriented
#Python,Guido van Rossum,1991,high-level,yes
```

PR #2149, opened 2025-06-17
Add Mikrotik routeros ip hotspot ip-binding print terse without-paging

evilmonkey19:


ntc_templates/templates/index
```diff
@@ -822,6 +822,7 @@ linux_arp_-a.textfsm, .*, linux, arp -a
 linux_vzlist.textfsm, .*, linux, vzlist
 linux_top.textfsm, .*, linux, top
 
+mikrotik_routeros_ip_hotspot_ip-binding_print_terse_without-paging.textfsm, .*, mikrotik_routeros, [[/]]ip h[[otspot]] ip-b[[inding]] p[[rint]] t[[erse]] wi[[thout-paging]]
```
mjbear, line 825: The forward slash is likely in square brackets `[[/]]` to make that character optional.

Most of the time on Mikrotik RouterOS the commands are prefixed with a forward slash unless a command moves to a different level/context. **However the unit tests don't prefix the command name with a forward slash** ***(so that is likely the main reason for the optional forward slash)***.

I suspect the index could be changed to make the forward slash optional (without square brackets).
_If we (including maintainers) agree on that I'd open a new PR and not place that change in this PR._

**Edit:** Changed font styles to emphasize parts of a sentence.

ntc_templates/templates/mikrotik_routeros_ip_hotspot_ip-binding_print_terse_without-paging.textfsm
```diff
@@ -0,0 +1,12 @@
+Value INDEX (\d+)
+Value BY_PASSED (\S)
```
mjbear: The letter may be more than just bypassed.
Are you able to get an example of the other options?

Should the letter be `TYPE` and the current field named type be called `ACTION`?

https://wiki.mikrotik.com/Manual:IP/Hotspot#IP_Bindings

Please discuss here as I've not (yet?) spun up an instance of RouterOS to test.

mjbear: @evilmonkey19
Thoughts on the IP bindings output? (questions above)

mjbear: Message sent via Slack to ask for feedback

ntc_templates/templates/index
evilmonkey19: Looks fine to me. I was just following how it was done in previous templates :)

ntc_templates/templates/mikrotik_routeros_ip_hotspot_ip-binding_print_terse_without-paging.textfsm
evilmonkey19: Totally agree with that. I haven't never used the option block, but seeing the docs makes more sense. I will change it now :+1:

ntc_templates/templates/index
mjbear: > Looks fine to me. I was just following how it was done in previous templates :)

Thank you for the good discussion.
(Many of my comments I was verbose so I could remind myself weeks or month(s) later what I had figured out, haha.)

I'm going to let this one go for now, but maybe it's something I'll vet out before the next release. :grinning:

PR #2151, opened 2025-06-26
fix: Corrected Huawei VRP display startup template

arthurguyon:


ntc_templates/templates/huawei_vrp_display_startup.textfsm
```diff
@@ -24,4 +24,6 @@ Start
   ^\s+Next\sstartup\s+patch\s+package:\s+${NEXT_STARTUP_PATCH_PACKAGE}\s*$$
   ^\s+Startup\s+voice-files:\s+${STARTUP_VOICE_FILES}\s*$$
   ^\s+Next\sstartup\s+voice-files:\s+${NEXT_STARTUP_VOICE_FILES}\s*$$
-  ^. -> Error
+  ^Info:\s+.*$$
+  ^\s*$$
+  ^return\s*$$
```
mjbear: @arthurguyon
The `Error` directive needs to remain per [project guidelines](https://github.com/networktocode/ntc-templates/blob/master/docs/user/faq.md#why-is-there-a-requirement-to-use-error-in-every-template), but would be moved to the bottom.

```suggestion
  ^return\s*$$
  ^. -> Error
```

ntc_templates/templates/huawei_vrp_display_startup.textfsm
```diff
@@ -24,4 +24,7 @@ Start
   ^\s+Next\sstartup\s+patch\s+package:\s+${NEXT_STARTUP_PATCH_PACKAGE}\s*$$
   ^\s+Startup\s+voice-files:\s+${STARTUP_VOICE_FILES}\s*$$
   ^\s+Next\sstartup\s+voice-files:\s+${NEXT_STARTUP_VOICE_FILES}\s*$$
+  ^Info:\s+.*$$
+  ^\s*$$
+  ^return\s*$$
```
mjbear: @arthurguyon
Where are the regexes on lines 28 and 29 present in the raw test output?

I don't see a match in the [existing test](https://github.com/networktocode/ntc-templates/tree/master/tests/huawei_vrp/display_startup) data either.

arthurguyon: Sorry about that, return was useless, I added the blank line

PR #2166, opened 2025-07-29
Modify asa sh acl PROTOCOL to accept numeric vals

### LINKED ISSUE #2158: PROTOCOL regex needs to be updated in ASA ACL template
When an ACL is configured to match an IP protocol which are not defined in Cisco ASA, it can be a number as well. I see that it is defined accordingly in the ENTRY_PROTOCOL field. I want that value to be available in PROTOCOL field as well.

Current values -->
Value PROTOCOL ([a-z]+)
Value ENTRY_PROTOCOL ([0-9a-z\-]+)

New values -->
Value PROTOCOL ([0-9a-z\-]+)
Value ENTRY_PROTOCOL ([0-9a-z\-]+)

https://github.com/networktocode/ntc-templates/blob/0ba1f71432b92ab33a090efc6831514462b614f6/ntc_templates/templates/cisco_asa_show_access-list.textfsm#L8

mjbear:
resolves #2158 

Modified the ASA show access-list template so `PROTOCOL` will allow numeric protocol types.

This change did not cause generated output changes for the ***existing test data*** so it seems benign.

It changes the behavior for lines with **numeric protocol types** so the value is in `PROTOCOL` and no longer `ENTRY_PROTOCOL`.


jmcgill298 (approved):

mjbear:
Per a side conversation with Jacob, this is likely more of a bug fix item than anything else.
(Not so much breaking change for example.)

PR #2170, opened 2025-08-15
Release notes + version bump.

jvanderaa:
Version 8.0.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,28 @@
 ARG PYTHON_VER
-
 FROM python:${PYTHON_VER}-slim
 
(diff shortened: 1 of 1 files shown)


jvanderaa:
The Dockerfile needed to be updated thanks to a recent change in the Python Slim image.

Dockerfile
```diff
@@ -1,27 +1,28 @@
 ARG PYTHON_VER
-
 FROM python:${PYTHON_VER}-slim
```
glennmatthews, line 2: Alternatively could pin to `python:${PYTHON_VER}-slim-bookworm` I believe.

glennmatthews (approved):

PR #2175, opened 2025-08-21
Add values to cisco_xr bgp vrf all ipv4 uni sum

jorlandobr:
Old template only captured RD, VRF, STATE and LOCAL_AS_NUMBER.

Added values: BGP_NEIGHBOR, NEIGHBOR_AS, MESSAGES_RECEIVED, MESSAGES_SENT, TABLE_VERSION, INPUT_QUEUE, 
OUTPUT_QUEUE, UP_DOWN, STATE_OR_PREFIXES_RECEIVED, SPK.

Updated cisco_xr_show_bgp_vrf_all_ipv4_unicast_summary.yml with new values.

### DIFF
--- a/ntc_templates/templates/cisco_xr_show_bgp_vrf_all_ipv4_unicast_summary.textfsm
+++ b/ntc_templates/templates/cisco_xr_show_bgp_vrf_all_ipv4_unicast_summary.textfsm
@@ -9,7 +9,7 @@ Value List MESSAGES_SENT (\d+)
 Value List TABLE_VERSION (\d+)
 Value List INPUT_QUEUE (\d+)
 Value List OUTPUT_QUEUE (\d+)
-Value List UP_DOWN (\S+?)
+Value List UP_DOWN (\S+)
 Value List STATE_OR_PREFIXES_RECEIVED (\S+)
 Value List SPK (\d+)
 


ntc_templates/templates/cisco_xr_show_bgp_vrf_all_ipv4_unicast_summary.textfsm
```diff
@@ -2,8 +2,21 @@ Value Required VRF (\S+)
...
 Value LOCAL_AS_NUMBER (\S+)
+Value List BGP_NEIGHBOR (\d+\.\d+\.\d+\.\d+)
+Value List NEIGHBOR_AS (\d+)
+Value List MESSAGES_RECEIVED (\d+)
+Value List MESSAGES_SENT (\d+)
+Value List TABLE_VERSION (\d+)
+Value List INPUT_QUEUE (\d+)
+Value List OUTPUT_QUEUE (\d+)
+Value List UP_DOWN (\S+?)
+Value List STATE_OR_PREFIXES_RECEIVED (\S+?\s+\S+?|\S+?)
```
mjbear: @jorlandobr
In the existing test data I can only find integers for State or Prefixes Received.

:dart: Would you please add test data that includes values that have spaces in them?

Thank you for adding test data to improve our test coverage (and of course contributions to the template)!

mjbear: @jorlandobr
Following up with you.

Within your recent commit (https://github.com/networktocode/ntc-templates/pull/2175/commits/1bdcd994e4ececb3847a0755504dcbcca3dd793c), the `St/PfxRcd` column I didn't catch any of the three instances with values that are two words.

I don't regularly work on XR so I'm not certain if the BGP states of OpenSent and OpenConfirm are displayed as one string of characters or if they're separated by a space. _I'm figuring that's about the only time there might be two words in that column._

***Do you have (or can you induce) an example of Open Sent/Confirm BGP state(s)?***

Thank you!

jorlandobr: I pasted the raw data and left office,. i will work on it. I could not find an instance of an "open" event. The space was just in case, if there is a spaced state..

In a few hours I should have everything working 🙂

mjbear: > I pasted the raw data and left office,. i will work on it. I could not find an instance of an "open" event. The space was just in case, if there is a spaced state..

Gotcha.
Based on experience those open sent/confirm states can be tougher to catch in action. :sweat_smile:

> In a few hours I should have everything working 🙂

I appreciate that - take your time, make sure to get something to eat. :sandwich: :apple: :milk_glass: :grin:

jorlandobr: I finished the editing the raw file, some fields were misalligned due  cleanning sensitive data. I added a spaced "Idle Test" and it is captured. So in the event of an unknown or future spaced State/Prefix Received the template will work.

But, I got curious and started searching for OpenSent and OpenConfirm and they are single string. What do you think? Adjust it for single string or let it "future proof"?

Thanks for generating the yaml! 🙂

mjbear: > I finished the editing the raw file, some fields were misalligned due cleanning sensitive data.

When I have a situation like this I try to add spaces to align the text the best I can.

> I added a spaced "Idle Test" and it is captured. So in the event of an unknown or future spaced State/Prefix Received the template will work.

I would say that if "Idle Test" does not show up in any live output, we should replace it with a BGP state that does. (Fun BGP stuff :sweat_smile:, I'd expect it to be one of Active, Idle, OpenSent, OpenConfirm, Established.)

> But, I got curious and started searching for OpenSent and OpenConfirm and they are single string. What do you think? Adjust it for single string or let it "future proof"?

Given what you found regarding OpenSent and OpenConfirm, my preference would be to keep the regex simple and account for just the single "[camel case](https://en.wikipedia.org/wiki/Camel_case)" word.

> Thanks for generating the yaml! 🙂

:100: :sweat_smile:
You're welcome!

jorlandobr: Template, raw and yaml adjusted. 🙂

ntc_templates/templates/cisco_xr_show_bgp_vrf_all_ipv4_unicast_summary.textfsm
```diff
@@ -2,8 +2,21 @@ Value Required VRF (\S+)
...
 Value STATE (\w+)
 Value LOCAL_AS_NUMBER (\S+)
+Value List BGP_NEIGHBOR (\d+\.\d+\.\d+\.\d+)
+Value List NEIGHBOR_AS (\d+)
+Value List MESSAGES_RECEIVED (\d+)
+Value List MESSAGES_SENT (\d+)
+Value List TABLE_VERSION (\d+)
+Value List INPUT_QUEUE (\d+)
+Value List OUTPUT_QUEUE (\d+)
+Value List UP_DOWN (\S+?)
```
mjbear: @jorlandobr
Is there any instance where Up/Down is blank?

***If not*** the regex can probably drop the question mark.

```suggestion
Value List UP_DOWN (\S+)
```

jorlandobr: Tried different routers, none had that field empty. Cleaned.

mjbear (approved):

PR #2181, opened 2025-08-27
Add Mikrotik RouterOS 'interface bridge host print terse without-paging'

oe1rfc:
Add a new template for Mikrotik RouterOS `/interface bridge host print terse without-paging` for parsing the bridge table


tests/mikrotik_routeros/interface_bridge_host_print_terse_without-paging/interface_bridge_host_print_terse_without-paging.raw
mjbear, line 1: @oe1rfc
Please add examples to the raw test data file for the flags (X, I, D, L, E).

The current test data has DLE shown.
Could we get examples for X (disabled) and I (invalid)?

Thank you!

ntc_templates/templates/mikrotik_routeros_interface_bridge_host_print_terse_without-paging.textfsm
```diff
@@ -0,0 +1,15 @@
+Value ID (\d+)
+Value DISABLED (X)
+Value INVALID (I)
+Value DYNAMIC (D)
+Value LOCAL (L)
+Value EXTERNAL (E)
```
mjbear, line 6: _(Might be some silly questions.)_

Is there a benefit to having each of these flags in its own key?

Is it feasible to have them all represented in a single key?

_(Though it seems like the single key per flag is the general convention for the Mikrotik RouterOS TextFSM templates.)_

The thought is to discuss before changing anything - if anything actually needs changed as well. :sweat_smile:

tests/mikrotik_routeros/interface_bridge_host_print_terse_without-paging/interface_bridge_host_print_terse_without-paging.raw
oe1rfc: I currently don't have write access to a RouterOS device, I have to check. And I'm not sure how to get an invalid entry :sweat_smile:

ntc_templates/templates/mikrotik_routeros_interface_bridge_host_print_terse_without-paging.textfsm
oe1rfc: I thought about this as well.
I don't think there can be mutually exclusive fields (like `X/R disabled/running` in `interface print brief`) that could be summarized in a single field, since the local MAC on the bridge itself is even marked as `LE`, both _local and external at the same time_, what ever that means in RouterOS.
Seen from the user perspective I think that having them as separate fields is a bit more convenient, since you don't have to know what the values mean, the key defines functionality.

tests/mikrotik_routeros/interface_bridge_host_print_terse_without-paging/interface_bridge_host_print_terse_without-paging.raw
oe1rfc: So, I added more examples for X,I, added the comment and vlan-id/vid fields, and set `on-interface` as optional, since it is actually not mandatory.

PR #2205, opened 2025-10-09
Fix cisco_ios_traceroute for no VRF info

akira6592:
Older versions of Cisco IOS do not display the `VRF info` line.

Fixed to allow templates to be parsed regardless of whether the `VRF info`  line is present or not.

### DIFF
--- a/ntc_templates/templates/cisco_ios_traceroute.textfsm
+++ b/ntc_templates/templates/cisco_ios_traceroute.textfsm
@@ -40,6 +40,5 @@ Entries
   ^.*?(?:(?:\d+\s+msec|![A-Z]|\*|\?)\s+){8}${RTT_RESPONSE} -> Continue
   ^.*?(?:(?:\d+\s+msec|![A-Z]|\*|\?)\s+){9}${RTT_RESPONSE} -> Continue
   ^.* -> Record
-  ^. -> Error
 
 EOF


ntc_templates/templates/cisco_ios_traceroute.textfsm
```diff
@@ -20,8 +20,8 @@ Value DETAILS (.*)
 
 Start
   ^Type\s+escape
-  ^Tracing\s+the\s+route
-  ^VRF\s+info: -> Entries
+  ^Tracing\s+the\s+route -> Entries
+  ^VRF\s+info:
```
mjbear: @akira6592
Based on your change, VRF info no longer needs to be in the Start state.

I would expect the VRF info line to be within the `Entries` state since it comes after `Tracing the route` line in the raw output.

:bulb: Ahhh, the Entries state doesn't have an Error directive ... hmmm, that may take some testing.

If you're up for it, try moving VRF info to the Entries state and adding an Error directive. Then see if you can get the cli.py helper script to work on the template.

Thank you!

akira6592: @mjbear Thank you for reviewing! I fixed it.

I simply removed the VRF info from the Start state. The template does not parse VRF.

ntc_templates/templates/cisco_ios_traceroute.textfsm
```diff
@@ -41,5 +40,6 @@ Entries
   ^.*?(?:(?:\d+\s+msec|![A-Z]|\*|\?)\s+){8}${RTT_RESPONSE} -> Continue
   ^.*?(?:(?:\d+\s+msec|![A-Z]|\*|\?)\s+){9}${RTT_RESPONSE} -> Continue
   ^.* -> Record
+  ^. -> Error
```
mjbear: As noted in my comment the Error directive doesn't do anything based on the Record line's pattern. (So my suggestion on the VRF in start was helpful, however no Error directive was needed in `Entries`. My mistake.)

```suggestion
```

ntc_templates/templates/cisco_ios_traceroute.textfsm
mjbear: @akira6592
Because of the pattern at the Record action in the Entries state, the Error directive doesn't do anything. (My bad for suggesting it without realizing.)

I just put a commit in to remove the Error directive from `Entries`.

There are some templates where the logic ends up requiring the Error directive to be absent in order to work ***or in this case the Error dir doesn't do anything!*** :zany_face:

mjbear (approved):

ntc_templates/templates/cisco_ios_traceroute.textfsm
akira6592: Thank you!

PR #2210, opened 2025-10-20
Add hp_comware disp link-aggregation member-port

marcus-cain:
As part of professional service work with NTC. We are updating the nautobot device_onboarding_app for the hp_comware command_mapper. As part of that work, we created a new textFSM template for "display link-aggregation member-port" and have ran this template against comware 5 and comware 7 devices. This PR is to contribute back to the community to be able to parse this command.

Index file (following longest-match rule) has been updated, along with test files from both comware 5 and comware 7 devices


ntc_templates/templates/hp_comware_display_link-aggregation_member-port.textfsm
```diff
@@ -0,0 +1,49 @@
...
+  ^\s+Port\sPriority:\s+${LOCAL_PORT_PRIORITY}\s*$$ 
+  ^\s+Oper-Key:\s+${LOCAL_OPER_KEY}\s*$$ 
+  ^\s+Flag:\s+${LOCAL_FLAG}\s*$$
+  ^Remote:\s*$$ -> REMOTE_PORT_OUTPUT
+  ^. -> Error
+
+REMOTE_PORT_OUTPUT
+  ^\s+System\sID:\s+${REMOTE_SYSTEM_ID}\s*$$ 
+  ^\s+Port\sNumber:\s+${REMOTE_PORT_NUMBER}\s*$$ 
+  ^\s+Port\sPriority:\s+${REMOTE_PORT_PRIORITY}\s*$$
```
mjbear: @marcus-cain
Thanks for the eye test! :laughing: :grin:
There's a couple of single space whitespace regexes you might want to swap to `\s+` to "future proof".

Food for thought since I can't push updates to your branch and you'll have to go through more review.

```suggestion
  ^Received\s+LACP\s+Packets:\s+${RECEIVED_PACKETS}\s+packet\(s\)\s*$$
  ^Illegal:\s+${ILLEGAL_PACKETS}\s+packet\(s\)\s*$$
  ^Sent\s+LACP\s+Packets:\s+${SENT_PACKETS}\s+packet\(s\)\s*$$ -> Record
  ^\s*$$
  ^. -> Error

LOCAL_PORT_OUTPUT
  ^\s+Port\s+Number:\s+${LOCAL_PORT_NUMBER}\s*$$
  ^\s+Port\s+Priority:\s+${LOCAL_PORT_PRIORITY}\s*$$ 
  ^\s+Oper-Key:\s+${LOCAL_OPER_KEY}\s*$$ 
  ^\s+Flag:\s+${LOCAL_FLAG}\s*$$
  ^Remote:\s*$$ -> REMOTE_PORT_OUTPUT
  ^. -> Error

REMOTE_PORT_OUTPUT
  ^\s+System\s+ID:\s+${REMOTE_SYSTEM_ID}\s*$$ 
  ^\s+Port\s+Number:\s+${REMOTE_PORT_NUMBER}\s*$$ 
  ^\s+Port\s+Priority:\s+${REMOTE_PORT_PRIORITY}\s*$$ 
```

marcus-cain: @mjbear - I can certainly do that... In general, is it always best to just use \s+ instead of \s for a single for whitespace, so account for any variations where the vendor may put multiple whitespace characters instead of a single ?

mjbear: > @mjbear - I can certainly do that... In general, is it always best to just use \s+ instead of \s for a single for whitespace, so account for any variations where the vendor may put multiple whitespace characters instead of a single ?

Yes, that's correct.

Before I started to help out more often, the maintainers expressed that it is best to use `\s+` even when there's a single literal or regex space/tab, etc so the template is flexible.

It should in theory reduce maintenance efforts that would otherwise result in issues/PRs for just some spacing that changed. :relaxed:

tests/hp_comware/display_link-aggregation_member-port/display_link-aggregation_member-port_cw7_2.yml
mjbear, line 1: @marcus-cain
There isn't a Flags section in cw7_2.raw so the structured output ends up as an empty list.

Would you please look into this?
Thank you!

marcus-cain: @mjbear 
This is a copy/paste error on my behalf. I was trying not to include all the ports and only wanted to have unique output that warranted this new test file, but in doing so,  i missed copying the flags section.

To prove to you that my comware devices have "**Flags:**" in first line of the output. I ran a nornir script that runs the command "display link-aggregation member-port". This is stored in a result object. There is a result object for each host

As you can see, i have 871 comware devices:
```
>>> len(result)
871
```

I then looped through the results and print the first line of output from each device:
```
for item in result:
    output_of_first_line = result[item][1].result.splitlines()[0]
    if "No interface has link" not in output_of_first_line:
            print(item, result[item][1].result.splitlines()[0])
```
All items had **"Flags:"** in the output

Additionally, i did the reverse test. If Flags isn't in the first line (excluding devices where there are no interfaces participating in an etherchannel), then print

```
for item in result:
    output_of_first_line = result[item][1].result.splitlines()[0]
    if "Flags:" not in output_of_first_line and "No interface has link" not in output_of_first_line:
            print(item, result[item][1].result.splitlines()[0])
```
This resulted in no output.

So i will update the raw text of this testcase by putting the flags section and will regenerate the parsed test yaml files.

marcus-cain: @mjbear - all set. I have updated the raw file, regenerated the test. Requesting your review,please

mjbear: > @mjbear - all set. I have updated the raw file, regenerated the test. Requesting your review,please

@marcus-cain
I totally understand the copy/paste - it's small stuff, it happens.
We might owe @kayhern some coffees or something :sweat_smile: ... Thank you for the assistance (and patience) Kristen!

PR #2216, opened 2025-10-28
Add cisco_ios_show_sdwan_bfd_summary

sjhloco:
Add template for Cisco SDWAN show_sdwan_bfd_summary on IOS-XE


ntc_templates/templates/cisco_ios_show_sdwan_bfd_summary.textfsm
mjbear, line 1: @sjhloco
I suppose these summarized BFD stats aren't available on any of the non-summary commands, are they?

sjhloco: @mjbear not that I know off. Is not a biggy to have this I can just as easily screen-scrape what I need so just squash it if think isn't suitable.

mjbear: > @mjbear not that I know off. Is not a biggy to have this I can just as easily screen-scrape what I need so just squash it if think isn't suitable.

This was just a general question. We don't have a redundant template here, but avoiding overlapping/redundant templates is the reason for the question (ex: think of brief/summary versus detail output).

We have another pretty simple template for some Cisco WLC stuff that I questioned whether I should create ... but I was given reassurance from the group that it would be handy.

So that's to conclude by saying this template will be useful. :grinning:

PR #2223, opened 2025-10-29
Fix for FPR running ASA to cover admin down

sjhloco:
This off FPR hardware running ASA software:
-Hardware:   FPR-1010, 7151 MB RAM, CPU Atom C3000 series 2200 MHz, 1 CPU (4 cores)
-Cisco Adaptive Security Appliance Software Version 9.18(4)

Small alteration to fix difference in how down interface is shown on this platform

### DIFF
--- a/ntc_templates/templates/cisco_asa_show_interface_ip_brief.textfsm
+++ b/ntc_templates/templates/cisco_asa_show_interface_ip_brief.textfsm
@@ -1,6 +1,6 @@
 Value INTERFACE (\S+)
 Value IP_ADDRESS (\S+)
-Value STATUS (up|down|administratively down|admin down)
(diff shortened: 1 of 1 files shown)


ntc_templates/templates/cisco_asa_show_interface_ip_brief.textfsm
```diff
@@ -1,6 +1,6 @@
 Value INTERFACE (\S+)
 Value IP_ADDRESS (\S+)
-Value STATUS (up|down|administratively down)
+Value STATUS (up|down|administratively down|admin down)
```
mjbear: Alternatively make part of the string optional to reduce duplication.
(Will push the change to this branch/PR.)

```suggestion
Value STATUS (up|down|admin(istratively)? down)
```

PR #2225, opened 2025-10-29
Fix for different active/standby location name

sjhloco:
When looking at SSO redundancy on a _Catalyst 9600 Series 6 Slot Chassis_ it uses _'Switch'_ rather than _'slot'_ for _'Active/Standby Location'_. Isn't an issue to fix it, but it does make the variable naming slightly wrong as it is called ACTIVE_SLOT and STANDBY_SLOT.

Dunno if you are happy with that or want abstract and call it ACTIVE_LOCATION and STANDBY_LOCATION, but TBH that doesnt sound very self-explanatory to me and is also obviously would be a breaking change.


mjbear:
> When looking at SSO redundancy on a _Catalyst 9600 Series 6 Slot Chassis_ it uses _'Switch'_ rather than _'slot'_ for _'Active/Standby Location'_. Isn't an issue to fix it, but it does make the variable naming slightly wrong as it is called ACTIVE_SLOT and STANDBY_SLOT.
> 
> Dunno if you are happy with that or want abstract and call it ACTIVE_LOCATION and STANDBY_LOCATION, but TBH that doesnt sound very self-explanatory to me and is also obviously would be a breaking change.

If the word slot or switch were part of the captured data then location for part of the capture group name might be a bit more descriptive. Unfortunately that too would qualify as a breaking change (since presumably consumers expect a numeric string).

The bummer about the breaking changes is those PRs sit out here for a while until the next major release. (Understandably, just how it goes.)

mjbear (approved):

PR #2241, opened 2025-11-17
Add cisco_xr_show_running-config_interface

jorlandobr:
Template and test files for Cisco XR command `show running-config interface`.


ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,50 @@
+Value Required INTERFACE (\S+)
+Value DESCRIPTION (.*)
+Value BANDWIDTH (\d+)
+Value DOT1Q (\d+)
+Value SECOND_DOT1Q (\d+)
+Value VRF (\S+)
+Value List IP_ADDRESS (\d+.\d+.\d+.\d+)
+Value List PREFIX_LENGTH (\d+.\d+.\d+.\d+)
```
matt852:
```suggestion
Value List IP_NETMASK (\d+.\d+.\d+.\d+)
```

The `Value` named `PREFIX_LENGTH` is used to capture an IPv4 subnet mask (e.g., `255.255.255.252`), not a CIDR prefix length (e.g., `30`). The established convention in this repository for a subnet mask is `IP_NETMASK`.

You will also need to update the key in the corresponding YAML test file (`tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml`) from `prefix_length` to `ip_netmask`.

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,50 @@
...
+Value FLOW_MPLS_SAMPLER (\S+)
+
+Start
+  ^interface\s+${INTERFACE}
+  ^\s+description\s+${DESCRIPTION}
+  ^\s+bandwidth\s+${BANDWIDTH}
+  ^\s+service-policy\s+input\s+${INPUT_POLICY}
+  ^\s+service-policy\s+output\s+${OUTPUT_POLICY}
+  ^\s+vrf\s+${VRF}
+  ^\s+ipv4\s+address\s+${IP_ADDRESS}\s+${PREFIX_LENGTH}(?:\s+secondary)?
```
matt852:
```suggestion
  ^\s+ipv4\s+address\s+${IP_ADDRESS}\s+${IP_NETMASK}(?:\s+secondary)?
```

Same explanation as above in the .textfsm file.

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,50 @@
...
+  ^\s+ipv4\s+mtu\s+${MTU_IPV4}
+  ^\s+ipv6\s+enable
+  ^\s+load-interval\s+${LOAD_INTERVAL}
+  ^\s+flow\s+ipv4\s+monitor\s+${FLOW_MONITOR}\s+sampler\s+${FLOW_SAMPLER}\s+ingress
+  ^\s+flow\s+ipv6\s+monitor\s+${FLOW_MONITOR_V6}\s+sampler\s+${FLOW_SAMPLER_V6}\s+ingress
+  ^\s+flow\s+mpls\s+monitor\s+${FLOW_MPLS}\s+sampler\s+${FLOW_MPLS_SAMPLER}\s+ingress
+  ^\s+encapsulation\s+dot1q\s+${DOT1Q}(?:\s+second-dot1q\s+${SECOND_DOT1Q})?
+  ^\s+(lacp|bundle|dampening|mpls|shutdown|!)
+  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP}\s+(ingress|egress)
+  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6}\s+(ingress|egress)
```
matt852:
```suggestion
  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_IN}\s+ingress
  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_OUT}\s+egress
  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_IN}\s+ingress
  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_OUT}\s+egress
```
To match with changes requested in above .textfsm file.

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,50 @@
...
+Value List PREFIXV6_LENGTH (\d+)
+Value LOAD_INTERVAL (\d+)
+Value MTU (\d+)
+Value MTU_IPV4 (\d+)
+Value UNICAST_SOURCE (\S+)
+Value UNICAST_SOURCE_V6 (\S+)
+Value INPUT_POLICY (\S+)
+Value OUTPUT_POLICY (\S+)
+Value ACCESS_GROUP (\S+)
+Value ACCESS_GROUP_V6 (\S+)
```
matt852:
```suggestion
Value ACCESS_GROUP_IN (\S+)
Value ACCESS_GROUP_OUT (\S+)
Value ACCESS_GROUP_V6_IN (\S+)
Value ACCESS_GROUP_V6_OUT (\S+)
```

The current rules for `ipv4 access-group` and `ipv6 access-group` capture the name of the access list but discard the direction (`ingress` or `egress`). This can lead to incorrect data if an interface has ACLs applied in both directions.

This will also require updating the YAML test file to use `access_group_in`, `access_group_out`, etc.

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,50 @@
...
+  ^\s+load-interval\s+${LOAD_INTERVAL}
+  ^\s+flow\s+ipv4\s+monitor\s+${FLOW_MONITOR}\s+sampler\s+${FLOW_SAMPLER}\s+ingress
+  ^\s+flow\s+ipv6\s+monitor\s+${FLOW_MONITOR_V6}\s+sampler\s+${FLOW_SAMPLER_V6}\s+ingress
+  ^\s+flow\s+mpls\s+monitor\s+${FLOW_MPLS}\s+sampler\s+${FLOW_MPLS_SAMPLER}\s+ingress
+  ^\s+encapsulation\s+dot1q\s+${DOT1Q}(?:\s+second-dot1q\s+${SECOND_DOT1Q})?
+  ^\s+(lacp|bundle|dampening|mpls|shutdown|!)
+  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP}\s+(ingress|egress)
+  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6}\s+(ingress|egress)
+  ^\S+ -> Record
+  ^. -> Error
```
matt852:
```suggestion
```

The `^. -> Error` rule will cause the entire template to fail if it encounters any interface configuration line that is not explicitly defined in the template. It might be safer to remove this line.

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.raw
matt852, line 1: The template includes `Value` definitions and parsing rules for `mtu`, `ipv4 mtu`, and various `flow` monitor configurations, but the provided test data (`.raw` file) does not include these lines. All parsing logic in a template should be covered by tests.

Please add examples of the following configuration lines to your `.raw` test file and add the corresponding expected output to the `.yml` file:
- `mtu`
- `ipv4 mtu`
- `flow ipv4 monitor ...`
- `flow ipv6 monitor ...`
- `flow mpls monitor ...`

Here is an example you could add to `cisco_xr_show_running-config_interface.raw`:
```
interface TenGigE3/0/0/0
 description TEST MTU and FLOW
 mtu 9216
 ipv4 mtu 9000
 flow ipv4 monitor IPV4-MONITOR sampler IPV4-SAMPLER ingress
 flow ipv6 monitor IPV6-MONITOR sampler IPV6-SAMPLER ingress
 flow mpls monitor MPLS-MONITOR sampler MPLS-SAMPLER ingress
!
```

matt852 (changes requested):

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
jorlandobr: @matt852 , my raw data has over 1000 interfaces, I prefer to leave it and correct in case of fail.

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.raw
jorlandobr: I'll add these cases!

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,54 @@
+Value Required INTERFACE (\S+)
+Value DESCRIPTION (.*)
+Value BANDWIDTH (\d+)
+Value DOT1Q (\d+)
+Value SECOND_DOT1Q (\d+)
+Value VRF (\S+)
+Value List IP_ADDRESS (\d+.\d+.\d+.\d+)
+Value List IP_NETMASK (\d+.\d+.\d+.\d+)
+Value List IPV6_ADDRESS (\S+?)
```
mjbear: @jorlandobr
Why is the question mark `?` in the IPV6_ADDRESS capture group regex?
(This runs successfully against the test data without the question mark.)

_You can accept my suggestion with the 'commit suggestion' button. Thank you!_

```suggestion
Value List IPV6_ADDRESS (\S+)
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.raw
mjbear: I regenerated the yml for this and tested locally that it passes.
I pushed the yml update to your GH branch and would expect GH CI should pass as well. :slightly_smiling_face:

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,54 @@
+Value Required INTERFACE (\S+)
+Value DESCRIPTION (.*)
+Value BANDWIDTH (\d+)
+Value DOT1Q (\d+)
+Value SECOND_DOT1Q (\d+)
+Value VRF (\S+)
+Value List IP_ADDRESS (\d+.\d+.\d+.\d+)
+Value List IP_NETMASK (\d+.\d+.\d+.\d+)
```
mjbear: @jorlandobr @matt852
To align with some of the normalization (see [data_model.md](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/data_model.md)) we've done, we would want IP_NETMASK instead to be NETMASK. _(I'm open to making this change and pushing the updates to this PR if there are no disagreements on that direction.)_


_(For items not already defined in the data model, they'd presumably be up for discussion.)_

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,54 @@
+Value Required INTERFACE (\S+)
+Value DESCRIPTION (.*)
+Value BANDWIDTH (\d+)
+Value DOT1Q (\d+)
+Value SECOND_DOT1Q (\d+)
```
mjbear: Along the same sort of normalization lines...
There are existing templates that have INNER_VLANS ([cisco_ios](https://github.com/networktocode/ntc-templates/blob/3f1a8bc009f4bed52eb8aec009c897844aab59f0/ntc_templates/templates/cisco_ios_show_vlans.textfsm#L5), [raw](https://github.com/networktocode/ntc-templates/blob/3f1a8bc009f4bed52eb8aec009c897844aab59f0/tests/cisco_ios/show_vlans/cisco_ios_show_vlans_09.raw#L173)) and INTERNAL_VLAN / EXTERNAL_VLAN ([zte_zxros](https://github.com/networktocode/ntc-templates/blob/3f1a8bc009f4bed52eb8aec009c897844aab59f0/ntc_templates/templates/zte_zxros_show_arp.textfsm#L5), [raw](https://github.com/networktocode/ntc-templates/blob/3f1a8bc009f4bed52eb8aec009c897844aab59f0/tests/zte_zxros/show_arp/zte_zxros_show_arp.raw#L3)).

Based on the [XR docs on Q-in-Q](https://www.cisco.com/en/US/docs/ios/lanswitch/configuration/guide/lsw_ieee_802.1q.html#wp1049783), the (first) dot1q is the outer VLAN tag and the second dot1q is the inner VLAN tag.

Do we want INTERNAL_VLAN / EXTERNAL_VLAN or define a new "standard" of INNER_VLAN / OUTER_VLAN?
_(I'm leaning toward a new standard of inner/outer and updating zxros to align [yeah, breaking change].)_

Should we agree, I can work up those changes on this PR.

jorlandobr: Using INTERNAL_VLAN / EXTERNAL_VLAN  differs from the command used for interface configuration (dot1q/second dot1q). Personally, I don't like :-(

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
jorlandobr: I am ok with that modification. Since it's a normalization issue, I leave that to you decide the best course.

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
mjbear: Agree on the internal/external, that's not my first choice.

I like the INNER_VLAN and OUTER_VLAN as it is more explanatory to what each dot1q tag is what. (Hopefully more vendor-neutral as well.)

***Edit:*** Updated with [Cisco Q-in-Q (pdf)](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus5600/sw/interfaces/7x/b_5600_Interfaces_Config_Guide_Release_7x/configuring_q_in_q___vlan_tunnels.pdf) and [JunOS Q-in-Q](https://www.juniper.net/documentation/us/en/software/junos/multicast-l2/topics/topic-map/q-in-q.html) docs to support my suggestion for capture group naming.

jmcgill298: I do think inner/outer is more indicative of what it is

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
matt852: This was my mistake, I missed it from the docs the first time.  I'll make some suggested changes in the review here to fix it.

matt852:
```suggestion
Value List NETMASK (\d+.\d+.\d+.\d+)

```

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,54 @@
...
+Value FLOW_MPLS_SAMPLER (\S+)
+
+Start
+  ^interface\s+${INTERFACE}
+  ^\s+description\s+${DESCRIPTION}
+  ^\s+bandwidth\s+${BANDWIDTH}
+  ^\s+service-policy\s+input\s+${INPUT_POLICY}
+  ^\s+service-policy\s+output\s+${OUTPUT_POLICY}
+  ^\s+vrf\s+${VRF}
+  ^\s+ipv4\s+address\s+${IP_ADDRESS}\s+${IP_NETMASK}(?:\s+secondary)?
```
matt852:
```suggestion
  ^\s+ipv4\s+address\s+${IP_ADDRESS}\s+${NETMASK}(?:\s+secondary)?

```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_mpls_sampler: ""
+    flow_sampler: ""
+    flow_sampler_v6: ""
+    input_policy: "IN_BAND_S_50000_V1"
+    interface: "TenGigE2/1/0/0.1234256"
+    ipv6_address:
+      - "fe80:a8:2:b0::309"
+    ip_address:
+      - "10.242.26.65"
+    ip_netmask:
```
matt852:
```suggestion
    netmask:
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_mpls_sampler: ""
+    flow_sampler: ""
+    flow_sampler_v6: ""
+    input_policy: "IN_BAND_S_100000_V1"
+    interface: "TenGigE2/4/0/2.22227"
+    ipv6_address:
+      - "fe80:a8:2:b8::a671"
+    ip_address:
+      - "10.23.125.209"
+    ip_netmask:
```
matt852:
```suggestion
    netmask:
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_mpls_sampler: ""
+    flow_sampler: ""
+    flow_sampler_v6: ""
+    input_policy: "IN_BAND_S_100000_V1"
+    interface: "TenGigE2/4/0/2.22229"
+    ipv6_address:
+      - "fe80:a8:2:b8::10ed"
+    ip_address:
+      - "10.228.247.93"
+    ip_netmask:
```
matt852:
```suggestion
    netmask:
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_mpls: ""
+    flow_mpls_sampler: ""
+    flow_sampler: ""
+    flow_sampler_v6: ""
+    input_policy: ""
+    interface: "TenGigE2/4/0/2.10012"
+    ipv6_address: []
+    ip_address:
+      - "10.86.193.193"
+    ip_netmask:
```
matt852:
```suggestion
    netmask:
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_mpls_sampler: ""
+    flow_sampler: ""
+    flow_sampler_v6: ""
+    input_policy: "IN_100000K_1_0_1_0_0_1_1"
+    interface: "TenGigE2/4/0/2.1002"
+    ipv6_address:
+      - "fe80:a8:2:2b8::751"
+    ip_address:
+      - "10.245.37.157"
+    ip_netmask:
```
matt852:
```suggestion
    netmask:
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_mpls_sampler: ""
+    flow_sampler: ""
+    flow_sampler_v6: ""
+    input_policy: "IN_BLUE_2M_0_0_0_98_0_1_XR_V1"
+    interface: "TenGigE2/4/0/2.10023"
+    ipv6_address:
+      - "fe80:a8:2:2b8::ac5"
+    ip_address:
+      - "192.168.226.53"
+    ip_netmask:
```
matt852:
```suggestion
    netmask:
```

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml
```diff
@@ -0,0 +1,213 @@
...
+    flow_monitor_v6: "IPV6-MONITOR"
+    flow_mpls: "MPLS-MONITOR"
+    flow_mpls_sampler: "MPLS-SAMPLER"
+    flow_sampler: "IPV4-SAMPLER"
+    flow_sampler_v6: "IPV6-SAMPLER"
+    input_policy: ""
+    interface: "TenGigE3/0/0/0"
+    ipv6_address: []
+    ip_address: []
+    ip_netmask: []
```
matt852:
```suggestion
    netmask:
```

mjbear:
* :white_check_mark: Normalized IP_NETMASK to NETMASK
* :white_check_mark: Normalized DOT1Q to OUTER_VLAN (added to *data_model.md*)
* :white_check_mark: Normalized SECOND_DOT1Q to INNER_VLAN (added to *data_model.md*)
* :white_check_mark: Matched the date stamp more (to avoid inadvertent matches)
* :white_check_mark: Updated the `Record` pattern to be the exclamation mark `!`
* :question: Since IPv4 addresses, IPv6 addresses, and IPv6 prefix lengths are all ***plural***, we probably should use the plural ones from the [data_model.md](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/data_model.md)
   * _Though_ we don't have a standard plural capture group for NETMASK
   * I suppose in retrospect those templates with the plural form ***could have been*** the singular form with the expectation that consumers/users would `type()` check the returned key+value pairs in the dictionary.
   * Thoughts? @jmcgill298 @jvanderaa @matt852

jorlandobr:
@mjbear , I noticed that template still marked as open/WIP. Some more input needed?

mjbear:
> @mjbear , I noticed that template still marked as open/WIP. Some more input needed?

Yes, it's still open.

There were varying thoughts on the singular vs plural capture group names which stalled merging this one. Specifically netmasks (despite what I wrote below).

***Update:*** I notice my mention contradicts a pattern that has been started in the [data model](https://github.com/networktocode/ntc-templates/blob/6961417e2c52b6306edf8460299410b56cf831cb/docs/dev/data_model.md#common-capture-groups) and in templates with IP_ADDRESSES and IPV6_ADDRESSES ... so forget what I wrote, there are precedents. :face_with_peeking_eye:

:point_right: *I will update the capture group names.*

~As a general statement, there are at least two schools of thought on the capture group names.~

1. ~Normalize so the capture group names are similar between vendors -- even despite the type potentially being a list.~
   * ~We can (and should already be) type checking so for the specific returned data that is a List can be detected and handled accordingly.~
   * ~:dart: My vote is to retain capture group normalization (rather than pluralize name based on data type).~
2. ~Have plural capture group names when the data type is a List.~
   * ~This breaks normalization and can be more difficult to work with.~
   * ~It does somewhat document the data type (being a List) -- however this can (and should) be done with a type check (we type check the data we process post-parse, right? :wink:).~

---

:new: In looking at this template again, I see that none of the IPv4 lists have more than one value. I have added an IPv4 secondary address by referencing some [XR documentation](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/ip-addresses/710x/b-ip-addresses-cg-ncs5500-710x/implementing-network-stack-ipv4-and-ipv6.html#concept_1F19F306F03E4F0D91B2AD74E0FF10CC) that has an example.

And all of the IPv6 lists only have one value as well.
As it so happens the example (also at the above Cisco XR docs link) has an EUI-64 v6 address configured ... so I updated the template to work with those.

@jorlandobr Would you please test this ***eui-64 configuration*** on a lab/development XR device?
Thank you!

matt852 (changes requested):
Thanks @jorlandobr, one suggestion before merge:

**Add test coverage for the egress ACLs and the swallow-keyword rule.** Every fixture interface uses only `ingress` ACLs, so `access_group_out` / `access_group_v6_out` are empty in every record, and the `^\s+(lacp|bundle|dampening|mpls|shutdown|!)` rule never fires against the test data. One extra interface covers both. Append to `cisco_xr_show_running-config_interface.raw`:

  ```diff
  +interface TenGigE3/0/0/1
  + description TEST EGRESS ACL
  + shutdown
  + ipv4 access-group iACL-IN ingress
  + ipv4 access-group iACL-OUT egress
  + ipv6 access-group FilterPackIPv6-IN ingress
  + ipv6 access-group FilterPackIPv6-OUT egress
  +!
  ```
Then regenerate the .yml with gen-yaml-folder / clean-yaml-folder and commit both.

jorlandobr:
@mjbear, @matt852 ,

since are many modifications I tested the new template against a newer box (over 900 interfaces, including bundle ones, that I hadn't before). I had to add new modifications. Please, take a look before I commit them.

```
Value Required INTERFACE (\S+)
Value DESCRIPTION (.*)
Value BANDWIDTH (\d+)
Value DOT1Q (\d+)
Value SECOND_DOT1Q (\d+)
Value VRF (\S+)
Value MAC_ADDRESS (\S+?)
Value List IP_ADDRESS (\d+.\d+.\d+.\d+)
Value List IP_NETMASK (\d+.\d+.\d+.\d+)
Value List IPV6_ADDRESS (\S+?)
Value List PREFIXV6_LENGTH (\d+)
Value LOAD_INTERVAL (\d+)
Value MTU (\d+)
Value MTU_IPV4 (\d+)
Value UNICAST_SOURCE (\S+)
Value UNICAST_SOURCE_V6 (\S+)
Value INPUT_POLICY (\S+)
Value OUTPUT_POLICY (\S+)
Value ACCESS_GROUP (\S+)
Value ACCESS_GROUP_V6 (\S+)
Value FLOW_MONITOR (\S+)
Value FLOW_MONITOR_V6 (\S+)
Value FLOW_MPLS (\S+)
Value FLOW_SAMPLER (\S+)
Value FLOW_SAMPLER_V6 (\S+)
Value FLOW_MPLS_SAMPLER (\S+)

Start
  ^interface\s+${INTERFACE}
  ^\s+description\s+${DESCRIPTION}
  ^\s+bandwidth\s+${BANDWIDTH}
  ^\s+service-policy\s+input\s+${INPUT_POLICY}
  ^\s+service-policy\s+output\s+${OUTPUT_POLICY}
  ^\s+vrf\s+${VRF}
  ^\s+ipv4\s+address\s+${IP_ADDRESS}\s+${IP_NETMASK}(?:\s+secondary)?
  ^\s+ipv4\s+verify\s+unicast\s+source\s+reachable-via\s+${UNICAST_SOURCE}(?:\s+allow\S)?
  ^\s+ipv4\s+unnumbered\s+\S
  ^\s+ipv6\s+(nd|verify)\s+\S
  ^\s+logging\s+\S
  ^\s+(mac-accounting|satellite-fabric-link|bfd|remote-ports|rewrite|carrier-delay|negotiation|ethernet|speed)\s+\S
  ^\s+ipv4\s+flowspec\s+disable
  ^\s+ipv6\s+flowspec\s+disable
  ^\s+ipv6\s+verify\s+unicast\s+source\s+reachable-via\s+${UNICAST_SOURCE_V6}\s+allow-self-ping
  ^\s+ipv6\s+address\s+${IPV6_ADDRESS}((?:\s+link-local)?|(?:\/+${PREFIXV6_LENGTH}\s*$$)?)
  ^\s+mac-address\s+${MAC_ADDRESS}
  ^\s+mtu\s+${MTU}
  ^\s+ipv4\s+mtu\s+${MTU_IPV4}
  ^\s+ipv6\s+enable
  ^\s+ipv6\s+mtu\s+\S
  ^\s+load-interval\s+${LOAD_INTERVAL}
  ^\s+flow\s+ipv4\s+monitor\s+${FLOW_MONITOR}\s+sampler\s+${FLOW_SAMPLER}\s+ingress
  ^\s+flow\s+ipv6\s+monitor\s+${FLOW_MONITOR_V6}\s+sampler\s+${FLOW_SAMPLER_V6}\s+ingress
  ^\s+flow\s+mpls\s+monitor\s+${FLOW_MPLS}\s+sampler\s+${FLOW_MPLS_SAMPLER}\s+ingress
  ^\s+encapsulation\s+(?:untagge|dot1q\s+${DOT1Q})?(?:\s+second-dot1q\s+${SECOND_DOT1Q})?
  ^\s+(lacp|bundle|dampening|mpls|shutdown|nv|lldp|l2transport|cdp|!)
  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP}\s+(ingress|egress)
  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6}\s+(ingress|egress)
  ^\S+ -> Record
  ^. -> Error
```

mjbear:
> since are many modifications I tested the new template against a newer box (over 900 interfaces, including bundle ones, that I hadn't before). I had to add new modifications. Please, take a look before I commit them.

This took a bit to adjust as the output above doesn't base off the [latest commit on this branch](https://github.com/jorlandobr/ntc-templates/blob/8feb15d51b0896f859ccd915535c2c36daeeaaa3/ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm) with the name changes.

We need the related raw output to support and test modifications.

> [!IMPORTANT]
> My suggestion would be to `git pull` the newest copy from your GitHub fork so you can work and test that. It would be easier to diff with git if commits are made to this branch -- certainly there could be a bit of discussion and adjustments as we go, but that's just the git workflow.

---

This is my attempt at a diff after updating some of the capture group names to match, but it's not precise and can't be easily referenced by line number on GH.

1. Why is this regex `Value List IPV6_ADDRESSES (\S+?)` made optional?
   (supporting output would help and it's easier to commit the changes/files)

@jorlandobr Please share the raw cli test data to see and explain the changes. Please work off the latest commit 8feb15d and layer on your changes. Thank you!

```diff
6a7
> Value MAC_ADDRESS (\S+?)
9c10
< Value List IPV6_ADDRESSES (\S+)
---
> Value List IPV6_ADDRESSES (\S+?)
18,21c19,20
< Value ACCESS_GROUP_IN (\S+)
< Value ACCESS_GROUP_OUT (\S+)
< Value ACCESS_GROUP_V6_IN (\S+)
< Value ACCESS_GROUP_V6_OUT (\S+)
---
> Value ACCESS_GROUP (\S+)
> Value ACCESS_GROUP_V6 (\S+)
37c36,42
<   ^\s+ipv4\s+verify\s+unicast\s+source\s+reachable-via\s+${UNICAST_SOURCE}\s+allow-self-ping
---
>   ^\s+ipv4\s+verify\s+unicast\s+source\s+reachable-via\s+${UNICAST_SOURCE}(?:\s+allow\S)?
>   ^\s+ipv4\s+unnumbered\s+\S
>   ^\s+ipv6\s+(nd|verify)\s+\S
>   ^\s+logging\s+\S
>   ^\s+(mac-accounting|satellite-fabric-link|bfd|remote-ports|rewrite|carrier-delay|negotiation|ethernet|speed)\s+\S
>   ^\s+ipv4\s+flowspec\s+disable
>   ^\s+ipv6\s+flowspec\s+disable
39c44,45
<   ^\s+ipv6\s+address\s+${IPV6_ADDRESSES}\/+${IPV6_PREFIX_LENGTHS}\s*(eui-64)?$$
---
>   ^\s+ipv6\s+address\s+${IPV6_ADDRESSES}((?:\s+link-local)?|(?:\/+${IPV6_PREFIX_LENGTHS}\s*$$)?)
>   ^\s+mac-address\s+${MAC_ADDRESS}
42a49
>   ^\s+ipv6\s+mtu\s+\S
47,55c54,58
<   ^\s+encapsulation\s+dot1q\s+${OUTER_VLAN}(?:\s+second-dot1q\s+${INNER_VLAN})?
<   ^\s+(lacp|bundle|dampening|mpls|shutdown|!)
<   ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_IN}\s+ingress
<   ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_OUT}\s+egress
<   ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_IN}\s+ingress
<   ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_OUT}\s+egress
<   ^!$$ -> Record
<   # date stamp
<   ^\w+\s+\w+\s+\d+
---
>   ^\s+encapsulation\s+(?:untagge|dot1q\s+${OUTER_VLAN})?(?:\s+second-dot1q\s+${INNER_VLAN})?
>   ^\s+(lacp|bundle|dampening|mpls|shutdown|nv|lldp|l2transport|cdp|!)
>   ^\s+ipv4\s+access-group\s+${ACCESS_GROUP}\s+(ingress|egress)
>   ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6}\s+(ingress|egress)
>   ^\S+ -> Record
```

jorlandobr:
@mjbear and @matt852 ,

added extra interfaces to cover the lines with rules not covered by the raw files.

tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.raw
```diff
@@ -0,0 +1,137 @@
...
+ ipv6 flowspec disable
+ service-policy output SubRate_200
+ mac-accounting ingress
+ encapsulation dot1q 31 exact
+  rewrite ingress tag pop 1 symmetric
+ ipv4 mtu 9180
+ ipv4 address 10.230.130.5 255.255.255.254
+ ipv6 address 2804:a8:2:b8::13a9/126
+ ipv6 address fe80::1000:16:3/126 link-local
+ ipv6 address fe80::1000:16:2 link-local
```
mjbear, line 111: The missing prefix length on raw file line 111, and 113 likely resulted in the `None` as seen in the [parsed yaml file](https://github.com/jorlandobr/ntc-templates/blob/20f7d97d972160188ba29ffe3b22a059f9685a42/tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml#L259) on lines 259 and 261.

The ***/19*** prefix length on line 114 may seem off, but is otherwise valid.

mjbear:
> @mjbear and @matt852 ,
> 
> added extra interfaces to cover the lines with rules not covered by the raw files.

Thank you @jorlandobr

I updated a couple of capture group names to match the data_model.md

I also added a note about some lines in the raw file that show up with a prefix length of `None` in the [parsed data](https://github.com/jorlandobr/ntc-templates/blob/20f7d97d972160188ba29ffe3b22a059f9685a42/tests/cisco_xr/show_running-config_interface/cisco_xr_show_running-config_interface.yml#L259) (yaml).

jorlandobr:
Dumping a xr box I had IPv6 lines ending with link-local and eui-64 without prefix lengths, so I added these lines. It's not an exercise, it's a real value. The /19 line was an exercise 🙂

mjbear:
> Dumping a xr box I had IPv6 lines ending with link-local and eui-64 without prefix lengths, so I added these lines. It's not an exercise, it's a real value. The /19 line was an exercise 🙂

Glad I asked - thank you @jorlandobr!

jmcgill298 (approved):

matt852:
Apologizes for the delay @jorlandobr, but I wanted to add a couple more requested changes before we merge this in that my Claude reviewer found.

- **Collapsing the ACLs into a single `ACCESS_GROUP` drops the ingress value when an interface has both directions.** On `TenGigE3/0/0/1` the raw has both `iACL-IN ingress` and `iACL-OUT egress`, but the parsed record keeps only `access_group: "iACL-OUT"` — the ingress ACL is lost. Restoring the direction-specific Values preserves both (I validated this — the record then captures all four ACLs and the scoped tests still pass).

  In the Value declarations:

  ```diff
  -Value ACCESS_GROUP (\S+)
  -Value ACCESS_GROUP_V6 (\S+)
  +Value ACCESS_GROUP_IN (\S+)
  +Value ACCESS_GROUP_OUT (\S+)
  +Value ACCESS_GROUP_V6_IN (\S+)
  +Value ACCESS_GROUP_V6_OUT (\S+)
  ```

  And the matching rules in `Start`:

  ```diff
  -  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP}\s+(ingress|egress)
  -  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6}\s+(ingress|egress)
  +  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_IN}\s+ingress
  +  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_OUT}\s+egress
  +  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_IN}\s+ingress
  +  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_OUT}\s+egress
  ```

  Then regenerate the `.yml` with `gen-yaml-folder` / `clean-yaml-folder` and commit both. (If you'd rather keep a flat model, making `ACCESS_GROUP` a `List` also avoids the overwrite, though direction is then no longer distinguishable.)

FYI / optional:

- `ipv6_prefix_lengths` carries the literal string `"None"` for the prefix-less link-local/eui-64 lines. The addresses are real, so this is defensible, but a consumer gets `"None"` rather than an empty value — worth a look before it locks in as the reference output.
- The IPv4 octet separators are unescaped — a bare `.` matches any character. Escaping is more precise (parsed output unchanged):

  ```diff
  -Value List IP_ADDRESSES (\d+.\d+.\d+.\d+)
  -Value List NETMASKS (\d+.\d+.\d+.\d+)
  +Value List IP_ADDRESSES (\d+\.\d+\.\d+\.\d+)
  +Value List NETMASKS (\d+\.\d+\.\d+\.\d+)
  ```

- The IPv6 line still uses `\/+` (one-or-more slashes) where XR emits a single `/`.

Thanks!

jorlandobr:
@matt852 , regarding the None in prefix-less link-local/eui-64 lines, I think is better. It's like an explict reminder for that rarer occurance in the configuration. I'll make the changes and send news files soon.

jorlandobr:
@matt852 , modifications implemented.

matt852:
Thanks @jorlandobr, however it looks like **The latest commits reverted the earlier naming normalization** fixes you implemented. `NETMASKS` went back to `IP_NETMASKS` and `IPV6_PREFIX_LENGTHS` back to `PREFIXV6_LENGTHS`. Is it possible the ACL rework was applied to an older local copy? 

  ```diff
  -Value List IP_NETMASKS (\d+\.\d+\.\d+\.\d+)
  +Value List NETMASKS (\d+\.\d+\.\d+\.\d+)
  ```

  ```diff
  -Value List PREFIXV6_LENGTHS (\d+)
  +Value List IPV6_PREFIX_LENGTHS (\d+)
  ```

  And the matching rules in `Start`:

  ```diff
  -  ^\s+ipv4\s+address\s+${IP_ADDRESSES}\s+${IP_NETMASKS}(?:\s+secondary)?
  +  ^\s+ipv4\s+address\s+${IP_ADDRESSES}\s+${NETMASKS}(?:\s+secondary)?
  ```

  ```diff
  -  ^\s+ipv6\s+address\s+${IPV6_ADDRESSES}(?:\/+${PREFIXV6_LENGTHS})?((?:\s+\S)?)
  +  ^\s+ipv6\s+address\s+${IPV6_ADDRESSES}(?:\/+${IPV6_PREFIX_LENGTHS})?((?:\s+\S)?)
  ```

  Then regenerate the `.yml` with `gen-yaml-folder` / `clean-yaml-folder` so every `ip_netmasks:` key becomes `netmasks:` and every `prefixv6_lengths:` key becomes `ipv6_prefix_lengths:`, and commit both files.

FYI / optional:

- The IPv6 line still uses `\/+` (one-or-more slashes) where XR emits a single `/`; a lone `\/` is more precise (parsed output unchanged):

  ```diff
  -  ^\s+ipv6\s+address\s+${IPV6_ADDRESSES}(?:\/+${IPV6_PREFIX_LENGTHS})?((?:\s+\S)?)
  +  ^\s+ipv6\s+address\s+${IPV6_ADDRESSES}(?:\/${IPV6_PREFIX_LENGTHS})?((?:\s+\S)?)
  ```

jorlandobr:
Indeed, I made the modifications on my local machine :-) Corrected!

ntc_templates/templates/cisco_xr_show_running-config_interface.textfsm
```diff
@@ -0,0 +1,63 @@
...
+  ^\s+flow\s+ipv6\s+monitor\s+${FLOW_MONITOR_V6}\s+sampler\s+${FLOW_SAMPLER_V6}\s+ingress
+  ^\s+flow\s+mpls\s+monitor\s+${FLOW_MPLS}\s+sampler\s+${FLOW_MPLS_SAMPLER}\s+ingress
+  ^\s+encapsulation\s+(?:untagge|dot1q\s+${OUTER_VLAN})?(?:\s+second-dot1q\s+${INNER_VLAN})?
+  ^\s+(mac-accounting|satellite-fabric-link|bfd|remote-ports|rewrite|carrier-delay|lacp|!)
+  ^\s+(bundle|dampening|mpls|shutdown|nv|lldp|l2transport|cdp|negotiation|ethernet|speed|logging)
+  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_IN}\s+ingress
+  ^\s+ipv4\s+access-group\s+${ACCESS_GROUP_OUT}\s+egress
+  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_IN}\s+ingress
+  ^\s+ipv6\s+access-group\s+${ACCESS_GROUP_V6_OUT}\s+egress
+  ^\S+ -> Record
```
mjbear: Should this `Record` line instead be a bit more self-documenting and instead be an exclamation mark?

```suggestion
  ^!$$ -> Record
```

mjbear (approved):

matt852 (approved):

PR #2247, opened 2025-11-27
Replace older toml lib with current tomli

a-detiste:
"tomli" got later merged in the standard library as "tomllib"

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -7,7 +7,7 @@ from invoke import task
 try:
     import tomli
 except ImportError:
-    sys.exit("Please make sure to `pip install toml` or enable the Poetry shell and run `poetry install`.")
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -5,7 +5,7 @@
 from invoke import task
 
 try:
-    import toml
+    import tomli
 except ImportError:
     sys.exit("Please make sure to `pip install toml` or enable the Poetry shell and run `poetry install`.")
```
mjbear: Should probably update the library name in error message too.
_(I'll commit this suggestion. We can always revert it if I'm off-base.)_

```suggestion
    sys.exit("Please make sure to `pip install tomli` or enable the Poetry shell and run `poetry install`.")
```

PR #2250, opened 2025-12-01
Update fortinet get system status for v7.4.8

giovanni-bellini-argo:


giovanni-bellini-argo:
added the test for 7.4.8, tell me if anything is missing

PR #2255, opened 2025-12-10
Add addr6 support to fortinet fnsysctl ifconfig

giovanni-bellini-argo:
Add inet addr6 support to fortinet fnsysctl ifconfig

### DIFF
--- a/ntc_templates/templates/fortinet_fnsysctl_ifconfig.textfsm
+++ b/ntc_templates/templates/fortinet_fnsysctl_ifconfig.textfsm
@@ -36,7 +36,7 @@ Start
   ^\s*TX\s+packets:${TX_PACKETS}\s+errors:${TX_ERRORS}\s+dropped:${TX_DROPPED}\s+overruns:${TX_OVERRUNS}\s+carrier:${TX_CARRIER}\s*$$
   ^\s*collisions:${COLLISIONS}\s+txqueuelen:${TX_QUEUE_LEN}\s*$$
   ^\s*RX\s+bytes:${RX_BYTES}\s+\(.*\)\s+TX\s+bytes:${TX_BYTES}\s+\(.*\)\s*$$
-  # The Interrupt value dosen't provide any usefull info and in later versions it seems to have been removed
+  # The Interrupt value doesn't provide any useful info and in later versions it seems to have been removed
   ^\s*Interrupt\s*:\s*\d+\s*$$
   ^\s*$$ -> Record
   ^. -> Error


giovanni-bellini-argo:
i also encountered an "Interrupt" option on an older version (v5.x) of my firewalls and i'm unsure whether it's still present in newer versions of ifconfig.

i didn't add it in the pull request for the above reason. tell me if u want to see it in the pull request.

ntc_templates/templates/fortinet_fnsysctl_ifconfig.textfsm
```diff
@@ -27,6 +28,7 @@ Value TX_BYTES (\d+)
 Start
   ^\s*${NIC}\s+Link\s+encap:${LINK_ENCAP}(?:\s+HWaddr\s+${HARDWARE_ADDRESS})?\s*$$
   ^\s*inet\s+addr:${IP_ADDRESS}(?:\s+Bcast:${BCAST})?\s+Mask:${MASK}\s*$$
+  ^\s*inet\s+addr6:\s+${IPV6_ADDRESS}\s+prefixlen\s\d+\s*$$
```
mjbear: I changed a whitespace regex to make it repeat so this is somewhat "future proofed" in case something simple such as spacing changes.

```suggestion
  ^\s*inet\s+addr6:\s+${IPV6_ADDRESS}\s+prefixlen\s+\d+\s*$$
```

giovanni-bellini-argo:
> I added patterns to capture the IPv6 prefix length as `PREFIX_LENGTH` from the data model.
> 
> _In terms of normalization, this is one of those situations where more discussion may occur or maybe we implement normalization at a later time for the next major release version._

as for normalization what u want to normalize exactly?
i see both terms `IPV6_ADDRESS` and `PREFIX_LENGTH` in the data_model file.

mjbear:
> i couldn't find in which versions of ifconfig this line is included so i can't state this line isn't included in newer versions, but the only case i found is in a v5.x, as state above.

Nice. Thank you.
It's probably best to add this since you noticed it - that way the template doesn't break if someone has some pretty old systems (or if the Interrupt shows up in the output in a later FortiOS release).

I was going to obfuscate your MAC addresses in the test data (in case you're worried about that), put some of those interfaces in a separate file, and generate the yaml. But I can't push those updates to your branch any longer. _If you'd like me to open a PR against your fork I certainly can._

> > I added patterns to capture the IPv6 prefix length as `PREFIX_LENGTH` from the data model.
> > _In terms of normalization, this is one of those situations where more discussion may occur or maybe we implement normalization at a later time for the next major release version._
> 
> as for normalization what u want to normalize exactly? i see both terms `IPV6_ADDRESS` and `PREFIX_LENGTH` in the data_model file.

Fair question as I wasn't clear.

Whether we want to differentiate IPv6 prefixes from IPv4 with the name.
(Safe to say probably no need to do anything different than we have on this template currently.)

giovanni-bellini-argo:
i agree, i don't see any use of it and it seems to have been removed in later versions (i dind't find any explicit info but in later versions it's nowhere to be seen)

ntc_templates/templates/fortinet_fnsysctl_ifconfig.textfsm
```diff
@@ -27,11 +29,14 @@ Value TX_BYTES (\d+)
...
+  ^\s*inet\s+addr6:\s+${IPV6_ADDRESS}\s+prefixlen\s+${PREFIX_LENGTH}\s*$$
   ^\s*link-local6:\s+${LINK_LOCAL6}\s+prefixlen\s+${LINK_LOCAL6_PREFIXLEN}\s*$$
   ^\s*${MULTICAST}\s+MTU:${MTU}\s+Metric:${METRIC}\s*$$
   ^\s*RX\s+packets:${RX_PACKETS}\s+errors:${RX_ERRORS}\s+dropped:${RX_DROPPED}\s+overruns:${RX_OVERRUNS}\s+frame:${RX_FRAME}\s*$$
   ^\s*TX\s+packets:${TX_PACKETS}\s+errors:${TX_ERRORS}\s+dropped:${TX_DROPPED}\s+overruns:${TX_OVERRUNS}\s+carrier:${TX_CARRIER}\s*$$
   ^\s*collisions:${COLLISIONS}\s+txqueuelen:${TX_QUEUE_LEN}\s*$$
-  ^\s*RX\s+bytes:${RX_BYTES}\s+\(.*\)\s+TX\s+bytes:${TX_BYTES}\s+\(.*\)\s*$$ -> Record
-  ^\s*$$
+  ^\s*RX\s+bytes:${RX_BYTES}\s+\(.*\)\s+TX\s+bytes:${TX_BYTES}\s+\(.*\)\s*$$
+  # The Interrupt value dosen't provide any usefull info and in later versions it seems to have been removed
```
mjbear: Adjusted spelling.

```suggestion
  # The Interrupt value doesn't provide any useful info and in later versions it seems to have been removed
```

PR #2260, opened 2025-12-18
Fix new format of cisco_ios show cdp neigh detail

Ardeck:
… options -   and add tests

## Summary
Update `cisco_ios_show_cdp_neighbors_detail.textfsm` to handle new format and many new fields

## Changes
- Adjust TextFSM regex for 
- Update `ntc_templates/templates/index` for correct mapping
- Add `.raw` and `.yml` test cases under `tests/cisco_ios/show_cdp_neighbors_detail/`

## Motivation
Current template is missing many parameters.

## Validation
- Ran `poetry run invoke tests` locally — all tests pass.
- Verified parsing against provided `.raw` samples.

## Notes
Scope limited to Cisco IOS  show  cdp neighbors detail.


ntc_templates/templates/cisco_ios_show_cdp_neighbors_detail.textfsm
```diff
@@ -1,27 +1,83 @@
 Value Required NEIGHBOR_NAME (\S+)
-Value MGMT_ADDRESS (\d+\.\d+\.\d+\.\d+|\w+\.\w+\.\w+)
-Value PLATFORM (.*)
-Value NEIGHBOR_INTERFACE (.*)
-Value LOCAL_INTERFACE (.*)
-Value NEIGHBOR_DESCRIPTION (.*$)
-Value CAPABILITIES (.+?)
+Value INTERFACE_IP (\d+\.\d+\.\d+\.\d+)
+Value List INTERFACE_IPV6 ([0-9a-fA-F:]+)
+Value MGMT_IP_ADDRESS (\d+\.\d+\.\d+\.\d+)
```
jvanderaa, line 4: I'd like to keep the MGMT_ADDRESS in lock step with LLDP neighbor outputs as well. So that these remain consistent across neighbor templates.

jvanderaa (commented):
I think the rest of these are making sense on the names, we may have to go update some of the other templates as we get this one merged in.

I'm a little worried about the NEIGHBOR_DESCRIPTION getting broken down, will we need to be updating this template quite a bit as new devices that come online since it is all free form fields.

mjbear:
> I am not sure in which version these change was made are the previous template is probably very old.

@Ardeck
Ok, what version and platform/model did you notice the change?
I figure this would be an IOS-XE release?

My reason for asking is to see if I have access to hardware that demonstrates these new format changes.

Thank you for details!

ntc_templates/templates/cisco_ios_show_cdp_neighbors_detail.textfsm
Ardeck: > I'd like to keep the MGMT_ADDRESS in lock step with LLDP neighbor outputs as well. So that these remain consistent across neighbor templates.

I have reverted the breaking changes.  At least the naming is not critical.

tests/cisco_ios/show_cdp_neighbors_detail/cisco_ios_show_cdp_neighbors_detail3.yml
```diff
@@ -0,0 +1,275 @@
...
+    power_request_level: ""
+    serial: ""
+    software: ""
+    software_family: ""
+    software_version: ""
+    unidirectional_mode: "off"
+    version:
+      - "Technical Support: http://www.cisco.com/techsupport"
+      - "Copyright (c) 1986-2010 by Cisco Systems, Inc."
+      - "Compiled Mon 15-Mar-10 23:32 by prod_rel_team"
```
mjbear, line 32: This isn't version information.

ntc_templates/templates/cisco_ios_show_cdp_neighbors_detail.textfsm
```diff
@@ -1,27 +1,95 @@
...
+  # For compatibility with legacy templates
+  ^${NEIGHBOR_DESCRIPTION} -> NewVersion
+  # Match software line with version info for Cisco Devices
+
+NewVersion
+  # Parse directly relevant information of the version of the software
+ ^\s*${SOFTWARE}\s*,\s*${SOFTWARE_FAMILY}\s*,\s*[^,]*\bVersion\s*:?\s*${SOFTWARE_VERSION}\s*,*\s*.*$$
+  ^\s*${SOFTWARE}\s*,*\s*${SOFTWARE_FAMILY}\s*,*s*[^,]*\bVersion\s*:?\s*${SOFTWARE_VERSION}\s*,*\s*.*$$
+  # get whatever is provided
+  ^${VERSION}
```
mjbear, line 85: Something seems a little off.

Plus, there are three capture groups with version in them.

* `VTP_VERSION` stands on its own, so no issue there
* There's `SOFTWARE_VERSION` and `VERSION` that are possibly duplicates
* The [LLDP equivalent](https://github.com/networktocode/ntc-templates/blob/d0bc5a480fe935afb6d9f6806c795a0390aaec90/ntc_templates/templates/cisco_ios_show_lldp_neighbors_detail.textfsm#L15) has capture group names with abbreviations (not sure how much normalization we'd do there -- depending on community preference)

ntc_templates/templates/cisco_ios_show_cdp_neighbors_detail.textfsm
```diff
@@ -1,27 +1,95 @@
...
 
 Start
-  ^Device ID: ${NEIGHBOR_NAME}
-  ^Entry address\(es\)\s*:\s* -> ParseIP
-  ^Platform\s*:\s*${PLATFORM}\s*,\s*Capabilities\s*:\s*${CAPABILITIES}\s+$$
-  ^Platform\s*:\s*${PLATFORM}\s*,\s*Capabilities\s*:\s*${CAPABILITIES}$$
-  ^Interface: ${LOCAL_INTERFACE},  Port ID \(outgoing port\): ${NEIGHBOR_INTERFACE}
-  ^Version : -> GetVersion
+   # Nexus Neighbor provides Serial Number
+  ^Device ID\s*:\s*${NEIGHBOR_NAME}\s*\(${SERIAL}\)\s*$$
```
mjbear, line 32: @Ardeck 
I looked for a Nexus or NX-OS neighbor in the test data, but didn't find one.
Did I miss it?

matt852 (changes requested):
**Breaking Change:** No

Thanks @Ardeck — this is a big usability upgrade for the CDP detail template, and the added fixtures cover a nice range of platforms. A few suggestions before merge:

- **Fix the `,*s*` typo in the second `NewVersion` software rule.** It matches optional literal `s` characters where whitespace was intended — and it's the rule doing the actual software capture today. In `ntc_templates/templates/cisco_ios_show_cdp_neighbors_detail.textfsm`:

  ```diff
  -  ^\s*${SOFTWARE}\s*,*\s*${SOFTWARE_FAMILY}\s*,*s*[^,]*\bVersion\s*:?\s*${SOFTWARE_VERSION}\s*,*\s*.*$$
  +  ^\s*${SOFTWARE}\s*,*\s*${SOFTWARE_FAMILY}\s*,*\s*[^,]*\bVersion\s*:?\s*${SOFTWARE_VERSION}\s*,*\s*.*$$
  ```

- **Remove the unreachable second `^Version` rule in `Start`.** It's identical to the rule above it, so TextFSM can never reach it:

  ```diff
     ^Version\s*:\s*$$ -> GetVersion
  -  ^Version\s*:\s*$$ -> NewVersion
  ```

- **Rename `INTERFACE_IPV6` to `IPV6_ADDRESSES`.** The data model (`docs/dev/data_model.md`) standardizes `IPV6_ADDRESSES` for lists of IPv6 addresses. The Value declaration:

  ```diff
  -Value List INTERFACE_IPV6 ([0-9a-fA-F:]+)
  +Value List IPV6_ADDRESSES ([0-9a-fA-F:]+)
  ```

  The matching rule in `ParseIP`:

  ```diff
  -  ^\s*IPv6 address:\s*${INTERFACE_IPV6}\s*\(.*\)\s*$$
  +  ^\s*IPv6 address:\s*${IPV6_ADDRESSES}\s*\(.*\)\s*$$
  ```

  And in the fixtures:

  ```diff
  -    interface_ipv6: []
  +    ipv6_addresses: []
  ```

  Apply the same key rename to every other `interface_ipv6:` line in all five `.yml` files.

- **Confirm `VTP_VERSION` should not be renamed to `CDP_VERSION`.** I had this in the Claude review output but would like to confirm it is accurate before suggesting you change it: "The rule feeding it matches `advertisement version: 2`, which is the CDP protocol version the neighbor advertises — not anything VTP-related — so the current name will mislead consumers."

If this is accurate, the following changes are needed. Otherwise let me know and I'll update the model.


  ```diff
  -Value VTP_VERSION (\d+)
  +Value CDP_VERSION (\d+)
  ```

  The matching rule in `Start`:

  ```diff
  -  ^advertisement\s+version\s*:\s*${VTP_VERSION}\s*$$
  +  ^advertisement\s+version\s*:\s*${CDP_VERSION}\s*$$
  ```

  And rename every `vtp_version:` key to `cdp_version:` in all five `.yml` files.

- **End the new states with `^. -> Error`.** Per `docs/dev/dev_parser.md`, each state should end with `^. -> Error` so unaccounted-for output fails loudly instead of being silently swallowed (e.g. a record separator reached while still in `ParseMgmtIP` would currently eat the following record). Add as the last rule of `ParseIP`, `NewVersion`, `ParseMgmtIP`, and `PwrAvail` — for example in `ParseIP`:

  ```diff
     ^\s*Platform\s*:\s*${PLATFORM}\s*,\s*Capabilities\s*:\s*${CAPABILITIES}\s*$$ -> Start
  +  ^. -> Error
  ```

  (`GetVersion` doesn't need it since `${NEIGHBOR_DESCRIPTION}` already matches every line.) This passes against all five of your fixtures as-is.

- **`SERIAL` never captures in any fixture.** The `^Device ID\s*:\s*${NEIGHBOR_NAME}\s*\(${SERIAL}\)\s*$$` rule isn't exercised — `serial` is empty in all 23 fixture rows. Could you add a raw sample with a `Device ID: <name>(<serial>)` neighbor (e.g. a Nexus peer), or drop the rule until real output is available?

- **The software fields only populate for legacy two-line version headers.** Because `GetVersion` consumes the first line of the version block as `NEIGHBOR_DESCRIPTION`, the modern single-line header (e.g. your 17.9.5 samples) never reaches the `NewVersion` software rules — `software`, `software_family`, and `software_version` come out empty for exactly the outputs this PR targets, and `software_family` is empty in every row. It's worth either restructuring so those fields populate consistently, or dropping them from this PR until they can be captured reliably.

Thanks!
(reviewed with Claude)

mjbear:
> * **Confirm `VTP_VERSION` should not be renamed to `CDP_VERSION`.** I had this in the Claude review output but would like to confirm it is accurate before suggesting you change it: "The rule feeding it matches `advertisement version: 2`, which is the CDP protocol version the neighbor advertises — not anything VTP-related — so the current name will mislead consumers."
> 
> 
> If this is accurate, the following changes are needed. Otherwise let me know and I'll update the model.
> 
> ```diff
> -Value VTP_VERSION (\d+)
> +Value CDP_VERSION (\d+)
> ```

`advertisement version: 2` is for the CDP version - references below.

```
Field Definitions:

    Advertisement version—The version of CDP being used for CDP advertisements.
```

Reference: [Catalyst 1200 and 1300 reference guide](https://www.cisco.com/c/en/us/td/docs/switches/campus-lan-switches-access/Catalyst-1200-and-1300-Switches/cli/C1200-cli/cdp-commands.html#wp3357581398)

Similar details also show up on another [Cat 1300 command reference guide](https://www.cisco.com/web/fw/tools/cisco-business/emulators/switch/catalyst/c1300-24mgp-4x/html/cat1k/english/1300/t_cdp_neighbors_information.html)


:white_check_mark: I have updated this capture group name from `VTP_VERSION` to `CDP_VERSION` and pushed to this branch.

mjbear:
:white_check_mark: I've also added Error directives to all the states.
:white_check_mark: And also fixed the white space regex typo.

So at this point the following items from Claude Code's suggestions have been resolved:

* VTP_VERSION > CDP_VERSION
* Error directives for all states
* `,*s*` typo fixed to be `,*\s*`

That leaves approximately four remaining items to for @Ardeck to review and action.

ntc_templates/templates/cisco_ios_show_cdp_neighbors_detail.textfsm
mjbear: At commit a840bb5fcb84bcf24e99e2e1b03ab7e0320688eb, both capture groups `MGMT_ADDRESS` and `MGMT_IP_ADDRESS` show up in the template. They capture the same data.

@Ardeck Is the intent to have both?

(I'm figuring we need to remove lines that reference `MGMT_IP_ADDRESS`.)

PR #2261, opened 2025-12-20
Add cisco ios show interface status err-disabled

clement-ldr:
New template to add support for cisco ios `show interface status err-disabled`  🤗


mjbear:
@clement-ldr
Since the raw cli output is basic/simple, it's possible to forego an extra State. The template will still produce the same expected output.

I reduced the redundant lines and I pushed a commit to your branch/PR to simplify the template. 
I hope this helps!

mjbear:
I decided to locally run the project unit tests and it threw an error on the [index order](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#index-file).
They're to be in order by length once they're grouped by platform name.

I've fixed this up, pushed those changes, and kicked off the unit tests on GitHub.

PR #2263, opened 2025-12-20
Add juniper junos show ethernet-switching interfaces

clement-ldr:
New template to add support for juniper junos `show ethernet-switching interfaces`  🤠

### DIFF
--- a/ntc_templates/templates/juniper_junos_show_ethernet-switching_interfaces.textfsm
+++ b/ntc_templates/templates/juniper_junos_show_ethernet-switching_interfaces.textfsm
@@ -12,6 +12,5 @@ Start
   ^${INTERFACE}\s+${STATUS}\s+${VLAN_MEMBER}\s+${TAGGING}\s+${PORT_ERROR}
   ^${INTERFACE}\s+${STATUS}\s+${VLAN_MEMBER}\s+${PORT_ERROR}
   ^\s+${VLAN_MEMBER}\s+${VLAN_ID}\s+${TAGGING}\s+${PORT_ERROR}
-  ^{.*}$$
   ^\s*$$
   ^. -> Error


ntc_templates/templates/juniper_junos_show_ethernet-switching_interfaces.textfsm
```diff
@@ -0,0 +1,17 @@
...
+Value List PORT_ERROR (.*)
+
+Start
+  ^\s*Interface\s+State\s+VLAN\s+members\s+Tag\s+Tagging\s+Blocking
+  ^\S -> Continue.Record
+  ^${INTERFACE}\s+${STATUS}\s+${VLAN_MEMBER}\s+${VLAN_ID}\s+${TAGGING}\s+${PORT_ERROR}
+  ^${INTERFACE}\s+${STATUS}\s+${VLAN_MEMBER}\s+${TAGGING}\s+${PORT_ERROR}
+  ^${INTERFACE}\s+${STATUS}\s+${VLAN_MEMBER}\s+${PORT_ERROR}
+  ^\s+${VLAN_MEMBER}\s+${VLAN_ID}\s+${TAGGING}\s+${PORT_ERROR}
+  ^{.*}$$
```
mjbear: @clement-ldr What is the `^{.*}$$` regex intended to match?

clement-ldr: Thats intended to match the `{master:0}` routing engine info juniper print before the user prompt. 
its part of the prompt but may get mixed with the command body and [somme](https://github.com/networktocode/ntc-templates/blob/master/ntc_templates/templates/juniper_junos_show_arp_no-resolve.textfsm#L11) [templates](https://github.com/networktocode/ntc-templates/blob/286039ef3a4c9b5899ef35eb7e662f44c7328576/ntc_templates/templates/juniper_junos_show_route_summary.textfsm#L17) have it, idk if its really needed.

clement-ldr: For example :
```
user@switch> show ethernet-switching interfaces
Interface    State  VLAN members        Tag   Tagging  Blocking
ae2.0        up     itx-s45-1           10    tagged   unblocked
                    itx-s45-1-adm       20    tagged   unblocked
                    l2rpn               4001  tagged   unblocked

bme0.32770   down   mgmt                      untagged unblocked
me0.0        down   mgmt                      untagged unblocked

{master:0}
user@switch>

{master:0}
user@switch>

{master:0}
user@switch>
```
{master:0} may be something else like {backup}, its more part of the the prompt though so im not sure its needed or if raising an error is better?

mjbear: > Thats intended to match the `{master:0}` routing engine info juniper print before the user prompt. its part of the prompt but may get mixed with the command body and [somme](https://github.com/networktocode/ntc-templates/blob/master/ntc_templates/templates/juniper_junos_show_arp_no-resolve.textfsm#L11) [templates](https://github.com/networktocode/ntc-templates/blob/286039ef3a4c9b5899ef35eb7e662f44c7328576/ntc_templates/templates/juniper_junos_show_route_summary.textfsm#L17) have it, idk if its really needed.

Ah, gotcha.

I wouldn't have guessed that unless I stumbled on another template or its raw cli data. :sweat_smile:

If the word varies from `master` some of the time then we could use a regex.
Would this `^{\S+:\d+}$$` be a viable alternative?

**Update:** From your output above, I wouldn't _think_ we should see `{master:0}` in the output. It's not clear if it appears at the tail end after a command or following a carriage return.

If there are templates without a line/rule for this type of pattern then we probably don't need it. (I haven't checked all the junos templates.)

clement-ldr: There are some templates without it and napalm.cli does not return it (netmiko probably does the same) I think we can remove it fine 👍

Edit:
Removed in https://github.com/networktocode/ntc-templates/pull/2263/commits/b051cc76374c448cae78ab7cb5cc891da0b8ade5

mjbear: > There are some templates without it and napalm.cli does not return it (netmiko probably does the same) I think we can remove it fine 👍

Ah, your mention about NAPALM and Netmiko made me realize that I looked into those lines a while back. For example, Netmiko strips those `{master:0}` strings before returning the output.

I suppose it's ok though if someone is parsing output retrieved using some other library or means (that doesn't remove those strings) then it's a potential wrinkle.

clement-ldr: I think its better to not match the `{master:0}` in the template since you would only have it in a command output if you parse the output wrong. 

The `{master:0}` and `user@switch>` are sent as one block by the switch, they are both part of the switch command line prompt which the template shouldn't match.

PR #2264, opened 2026-01-05
Update cisco_nxos_show_module for N9K linecard statuses

mtinberg:
Add test output from N9K-C9500 family device which has additional status 'active', 'standby' on the system controller and also include 'testing' and 'initializing' status from early boot (although I didn't retain a capture of that output) as we are using this template to detect when a N9K has completed the reboot process after firmware updates, which is some time after SSH logins become available.


mjbear:
@mtinberg
I made some tweaks to your PR to reduce the duplication in STATUS capture group.
If there's anything you might not agree with drop the question here for discussion.
(My updates are suggestions and can be changed before we merge this PR.)

Thank you!

PR #2269, opened 2026-01-26
Fix AOSCX sh int Sync devices from network

philsummers:
(aoscx show interface)

Added 2 x rate-limit directives to allow Sync Devices From Network job to complete successfully.

Added Leader-Follower directive to allow Sync Devices From Network job to complete successfully.

Tested on aoscx 10.16.1006

### DIFF
--- a/ntc_templates/templates/aruba_aoscx_show_interface.textfsm
+++ b/ntc_templates/templates/aruba_aoscx_show_interface.textfsm
@@ -82,11 +82,9 @@ Start
   ^\s*rate-limit\s+unknown-unicast
   ^\s*rate-limit\s+broadcast
   ^\s*rate-limit\s+multicast
-  ^\s*rate-limit\s+icmp
   ^\s*L3\s+Counters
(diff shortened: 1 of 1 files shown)


mjbear:
Hi @philsummers ! :slightly_smiling_face:

> Raw data has been added to the PR and I have used the helper script to generate the yaml.

Good.

> Tests have been run and pass (do you need evidence of this?)

No, I don't need evidence of the tests passing.
If the CI tests run and succeed on GitHub then we're good. :grinning:
(I'll review and then initiate a GH test run.)

There have been cases where PR contributions had the tests fail -- we want contributors to catch those items early.

> I have a couple of other changes almost ready to go (re sflow and rate limiting in the show_interfaces textfsm) just need to find time to grab the raw data etc.
>
> Would you like to wait for me to roll them in to this PR or just submit another PR when ready? It's likely I'll come across more changes needed as I'm currently onboarding around 250 aruba 6300m's.

Please include only one template per PR.

Thank you for your contribution!

PR #2270, opened 2026-01-28
Fix IOS show env power parser

hvanderheide:
On a non-PoE switch, "PoE Pwr" will display "n/a", which won't match "\w". Changed to match non-whitespace "\S". Tested on a bunch of Cisco C9300s

Also:

* Added Error directive
* Updated index pattern (which allows this to support older switches too)
* Added support for some older cli output


mjbear:
> On a non-PoE switch, "PoE Pwr" will display "n/a", which wonn't match "\w". Changed to match non-whitespace "\S". Tested on a bunch of Cisco C9300s

Hi @hvanderheide

Thank you for your submission and fix.

There's a little bit more needed to qualify this PR ready for review and merging.

* Would you please add [raw cli output for project unit tests](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#tests)?
* Please generated structured YAML off of the raw output using the [helper scripts](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_parser.md#development-helper-scripts).
* And run the [test cases via invoke](https://github.com/networktocode/ntc-templates/blob/master/docs/dev/dev_environment.md#invoke).

If there are any questions, please ask. Thank you!

hvanderheide:
> There's a little bit more needed to qualify this PR ready for review and merging.

Hi @mjbear, thanks for helping me through the process. I've added a test case with both the raw output and a YAML file. The test seems to succeed.

mjbear (dismissed):

mjbear:
@itdependsnetworks @jmcgill298 @jvanderaa @matt852
Could I get a second set of eyes on this PR?
(I made changes to the index which should be fine based on research, but it never hurts to have other reviewers!)

jvanderaa:
I'm tempted that we shouldn't reduce the command length matching part. There are likely other platforms that may have a command issue there. @itdependsnetworks / @jmcgill298 I'm going to need some of your thoughts there.

jmcgill298 (approved):

PR #2273, opened 2026-02-04
Update XR sh eigrp neigh timestamp regex

GutturalOnion:
(cisco_xr show eigrp neighbors)

Fixed Regex for timestamp matching to account for single-digit dates.


matt852:
@GutturalOnion Are you still working on this PR?

PR #2286, opened 2026-02-17
Fix mikrotik_routeros_system_resource_print

kwood92:
Build time regex to support RouterOS v7.13+


ntc_templates/templates/mikrotik_routeros_system_resource_print.textfsm
```diff
@@ -5,7 +5,7 @@ Value UPTIME_HOURS (\d+)
 Value UPTIME_MINUTES (\d+)
 Value UPTIME_SECONDS (\d+)
 Value VERSION (.+)
-Value BUILD_TIME ([A-Za-z]{3}\/\d{1,2}\/\d{4} \d{2}:\d{2}:\d{2})
+Value BUILD_TIME (\d{4}-\d{2}-\d{2}\s*\d{2}:\d{2}:\d{2}|[A-Za-z]{3}\/\d{1,2}\/\d{4} \d{2}:\d{2}:\d{2})
```
mjbear: @kwood92

By the output it looks like there's always a space between date format and time.

We can handle that either with a single literal space _or_ a regex `\s+` for one or more spaces.

I adjusted this to have the single literal space like the previous part of that regex. (However if someone thinks it is better to go the white space regex route we could go that route instead.) 

```suggestion
Value BUILD_TIME (\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}|[A-Za-z]{3}\/\d{1,2}\/\d{4} \d{2}:\d{2}:\d{2})
```

ntc_templates/templates/mikrotik_routeros_system_resource_print.textfsm
```diff
@@ -5,7 +5,7 @@ Value UPTIME_HOURS (\d+)
 Value UPTIME_MINUTES (\d+)
 Value UPTIME_SECONDS (\d+)
 Value VERSION (.+)
-Value BUILD_TIME ([A-Za-z]{3}\/\d{1,2}\/\d{4} \d{2}:\d{2}:\d{2})
+Value BUILD_TIME (\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}|[A-Za-z]{3}\/\d{1,2}\/\d{4} \d{2}:\d{2}:\d{2})
```
mjbear: We can reduce duplication since the time regex is the same between the two build time stamps.
(Wish I'd have recognized that sooner. :face_exhaling:)

Tested, working, adding this. Thank you.

```suggestion
Value BUILD_TIME ((\d{4}-\d{2}-\d{2}|[A-Za-z]{3}\/\d{1,2}\/\d{4}) \d{2}:\d{2}:\d{2})
```

mjbear (dismissed):

kwood92:
Thanks @mjbear 
I've commited some additional changes that fix lines containing extra (unneeded) whitespace introduced in newer versions of RouterOS 7

ntc_templates/templates/mikrotik_routeros_system_resource_print.textfsm
```diff
@@ -4,12 +4,12 @@ Value UPTIME_DAYS (\d+)
 Value UPTIME_HOURS (\d+)
 Value UPTIME_MINUTES (\d+)
 Value UPTIME_SECONDS (\d+)
-Value VERSION (.+)
-Value BUILD_TIME ([A-Za-z]{3}\/\d{1,2}\/\d{4} \d{2}:\d{2}:\d{2})
-Value FACTORY_SOFTWARE (\S+)
+Value VERSION (\S+(?:\s+\S+)*)
+Value BUILD_TIME ((\d{4}-\d{2}-\d{2}|[A-Za-z]{3}\/\d{1,2}\/\d{4}) \d{2}:\d{2}:\d{2})
+Value FACTORY_SOFTWARE (\S+(?:\s+\S+)*)
```
mjbear, line 9: The `factory software` data provided does not require a change to the regex as there only appears to be one word (not two words/strings separated with a space).
Do you have data that shows a version separated with a space?

ntc_templates/templates/mikrotik_routeros_system_resource_print.textfsm
```diff
@@ -18,9 +18,9 @@ Value TOTAL_HDD_SPACE (\d+\.\d{1}(K|M|G)iB)
 Value WRITE_SECT_SINCE_REBOOT (\d+)
 Value WRITE_SECT_TOTAL (\d+)
 Value BAD_BLOCKS (\d{1,3}(\.\d+)?%)
-Value ARCHITECTURE_NAME (\S*)
-Value BOARD_NAME (.+)
-Value PLATFORM (.+)
+Value ARCHITECTURE_NAME (\S+(?:\s+\S+)*)
```
mjbear, line 21: None of the `architecture name` in the provided data has a space and a second word. There doesn't appear to be a reason to change the regex unless there's test data we haven't seen.

ntc_templates/templates/mikrotik_routeros_system_resource_print.textfsm
```diff
@@ -18,9 +18,9 @@ Value TOTAL_HDD_SPACE (\d+\.\d{1}(K|M|G)iB)
...
+Value BOARD_NAME (\S+(?:\s+\S+)*)
+Value PLATFORM (\S+(?:\s+\S+)*)
```
mjbear, line 23: `platform` only shows up as a single word in the test data. There doesn't appear to be a reason to change the regex. Is there more test data?

kwood92:
@mjbear The example data is in tests/mikrotik_routeros/system_resource_print/mikrotik_routeros_system_resource_print_6.raw

All the lines aside from build-time are padded out with spaces to the end of the build-time string

For a visual reference, see the screenshot below:
image


As for the comments about `platform`, `architecture name` & `factory software`, right now I do not have any examples, it's just defensive changes in case Mikrotik changes their minds again.

mjbear:
> @mjbear The example data is in tests/mikrotik_routeros/system_resource_print/mikrotik_routeros_system_resource_print_6.raw
> 
> All the lines aside from build-time are padded out with spaces to the end of the build-time string
> 
> For a visual reference, see the screenshot below: image

I agree, the furthest right whitespace regex (at the end of the line) may as well just be `\s*`. Those elsewhere within the line can be `\s+` as long as there are 1+ spaces there reliably.

> As for the comments about `platform`, `architecture name` & `factory software`, right now I do not have any examples, it's just defensive changes in case Mikrotik changes their minds again.

@jvanderaa @jmcgill298
Thoughts on anticipating future changes when we don't have test data for these situations today (on this PR)?

PR #2287, opened 2026-02-18
Add mikrotik_routeros_interface_lte_apn_print_terse

kwood92:
Adds support for /interface lte print terse

### DIFF
--- a/ntc_templates/templates/mikrotik_routeros_interface_lte_apn_print_terse.textfsm
+++ b/ntc_templates/templates/mikrotik_routeros_interface_lte_apn_print_terse.textfsm
@@ -1,8 +1,8 @@
 Value INDEX (\d+)
-Value FLAGS (\*)
+Value FLAGS ([X\*]+)
 Value NAME (\S+)
 Value APN (\S+)
-Value IP_TYPE (auto|ipv4|ipv6)
+Value IP_TYPE (\S+)
 Value USE_PEER_DNS (yes|no)
 Value USE_NETWORK_APN (yes|no)
 Value ADD_DEFAULT_ROUTE (yes|no)
@@ -10,6 +10,6 @@ Value DEFAULT_ROUTE_DISTANCE (\d+)
 Value AUTHENTICATION (\S+)
 
 Start
- ^${INDEX}\s+${FLAGS}\s+name=${NAME}\s+apn=${APN}\s+use-peer-dns=${USE_PEER_DNS}\s+use-network-apn=${USE_NETWORK_APN}\s+add-default-route=${ADD_DEFAULT_ROUTE}\s+default-route-distance=${DEFAULT_ROUTE_DISTANCE}\s+ip-type=${IP_TYPE}\s+authentication=${AUTHENTICATION}\s*$$
+ ^${INDEX}\s+(${FLAGS}\s+)?name=${NAME}\s+apn=${APN}\s+use-peer-dns=${USE_PEER_DNS}\s+use-network-apn=${USE_NETWORK_APN}\s+add-default-route=${ADD_DEFAULT_ROUTE}\s+default-route-distance=${DEFAULT_ROUTE_DISTANCE}\s+ip-type=${IP_TYPE}\s+authentication=${AUTHENTICATION}\s*$$
   ^\s*$$
   ^. -> Error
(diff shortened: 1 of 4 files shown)


ntc_templates/templates/mikrotik_routeros_interface_lte_apn_print_terse.textfsm
```diff
@@ -0,0 +1,13 @@
...
+Value APN (\S+)
+Value IP_TYPE (auto|ipv4|ipv6)
+Value USE_PEER_DNS (yes|no)
+Value USE_NETWORK_APN (yes|no)
+Value ADD_DEFAULT_ROUTE (yes|no)
+Value DEFAULT_ROUTE_DISTANCE (\d+)
+Value AUTHENTICATION (\S+)
+
+Start
+ ^${INDEX}\s*${FLAGS}\s*name=${NAME}\s*apn=${APN}\s*use-peer-dns=${USE_PEER_DNS}\s*use-network-apn=${USE_NETWORK_APN}\s*add-default-route=${ADD_DEFAULT_ROUTE}\s*default-route-distance=${DEFAULT_ROUTE_DISTANCE}\s*ip-type=${IP_TYPE}\s*authentication=${AUTHENTICATION}\s*$$
```
mjbear: From my time in the past using Mikrotik gear, there's always a space between the units of data on a line. The whitespace regex for many of them should be `\s+`.

```suggestion
 ^${INDEX}\s+${FLAGS}\s+name=${NAME}\s+apn=${APN}\s+use-peer-dns=${USE_PEER_DNS}\s+use-network-apn=${USE_NETWORK_APN}\s+add-default-route=${ADD_DEFAULT_ROUTE}\s+default-route-distance=${DEFAULT_ROUTE_DISTANCE}\s+ip-type=${IP_TYPE}\s+authentication=${AUTHENTICATION}\s*$$
```

ntc_templates/templates/mikrotik_routeros_interface_lte_apn_print_terse.textfsm
```diff
@@ -0,0 +1,13 @@
...
+Value APN (\S+)
+Value IP_TYPE (auto|ipv4|ipv6)
+Value USE_PEER_DNS (yes|no)
+Value USE_NETWORK_APN (yes|no)
+Value ADD_DEFAULT_ROUTE (yes|no)
+Value DEFAULT_ROUTE_DISTANCE (\d+)
+Value AUTHENTICATION (\S+)
+
+Start
+ ^${INDEX}\s+${FLAGS}\s+name=${NAME}\s+apn=${APN}\s+use-peer-dns=${USE_PEER_DNS}\s+use-network-apn=${USE_NETWORK_APN}\s+add-default-route=${ADD_DEFAULT_ROUTE}\s+default-route-distance=${DEFAULT_ROUTE_DISTANCE}\s+ip-type=${IP_TYPE}\s+authentication=${AUTHENTICATION}\s*$$
```
jvanderaa, line 13: We need the error part yet on this one.

jvanderaa (changes requested):

ntc_templates/templates/mikrotik_routeros_interface_lte_apn_print_terse.textfsm
kwood92: Apologies, that missing error directive has been added in my last commit

matt852 (changes requested):
**Breaking Change:** No

Thanks @kwood92 — this is in great shape, and the earlier whitespace/error-rule fixes look good. Two suggestions before merge, both to make the template parse real-world APN tables and not just the builtin `default` entry:

- **`FLAGS` only matches the default `*` and is required.** As written, a user-added APN (no flag) or a disabled one (`X`) won't parse. The other mikrotik `_terse` templates capture flags as a multi-char, optional column — recommend matching that.

  In `mikrotik_routeros_interface_lte_apn_print_terse.textfsm`, the Value declaration:
  ```diff
  -Value FLAGS (\*)
  +Value FLAGS ([X\*]+)
  ```
  And the matching rule in `Start`, making the flag column optional:
  ```diff
  - ^${INDEX}\s+${FLAGS}\s+name=${NAME}\s+apn=${APN}...
  + ^${INDEX}\s+(${FLAGS}\s+)?name=${NAME}\s+apn=${APN}...
  ```

- **`IP_TYPE` can't match `ipv4-ipv6`.** RouterOS supports a dual-stack `ip-type=ipv4-ipv6`, which the closed `(auto|ipv4|ipv6)` alternation would fail on. Simplest is to broaden it (the existing `ipv4` sample still parses unchanged):
  ```diff
  -Value IP_TYPE (auto|ipv4|ipv6)
  +Value IP_TYPE (\S+)
  ```
  (If you'd rather keep the alternation for validation, add `ipv4-ipv6` to it and include raw rows that exercise each option.)

Thanks!

kwood92:
@matt852 Thanks for the feedback. I have added the requested changes.

matt852 (approved):

PR #2291, opened 2026-03-04
Update cisco_ios_show_sdwan_control_connections

ppetilli:


mjbear (approved):
That's interesting Cisco changed the table header formatting.

Your fix takes care of it by being backwards compatible between existing test data as well as your newly added test data.

Good job. Thank you!

PR #2294, opened 2026-03-09
Add Prun state to alcatel sros show service sap-using

jorlandobr:
Found that there is a "Prun" (pruned) state that wasn't covered on the original template.

Added raw and test files.


ntc_templates/templates/alcatel_sros_show_service_sap-using.textfsm
```diff
@@ -4,8 +4,8 @@ Value Required INGRESS_QOS (\S+)
 Value Required INGRESS_FILTER (\S+)
 Value Required EGRESS_QOS (\d+)
 Value Required EGRESS_FILTER (\S+)
-Value Required ADMIN_STATE (Up|Down)
-Value Required OPER_STATE (Up|Down)
+Value Required ADMIN_STATE (Up|Down|Prun)
```
mjbear: @jorlandobr 
I only see "Prun" on the operational state.

Do you have live device test data to support the admin state also showing Prun?

Thank you!

.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ concurrency:  # Cancel any existing runs of this workflow for this same PR
 on: # yamllint disable-line rule:truthy rule:comments
   push:
     branches:
-      - "master"
+      - "pruned"
```
mjbear: I can imagine you wanted the CI to run the tests on your forked repo as you added commits. :slightly_smiling_face:

We don't want to update the CI config for the upstream repository though.
(I'm changing the branch name back and committing this suggestion.)

```suggestion
      - "master"
```

ntc_templates/templates/alcatel_sros_show_service_sap-using.textfsm
jorlandobr: @mjbear

On live data, only on operational state.

mjbear: > @mjbear
> 
> On live data, only on operational state.

@jorlandobr 
Could we get raw output from a test device to show the "admin state" in a `Prun` state?

(We don't want to change the template for conditions or output that doesn't actually exist. :slightly_smiling_face:)

jorlandobr: I couldn't... even the PRUN on a operational state was caught by chance...

mjbear: > I couldn't... even the PRUN on a operational state was caught by chance...

Caught by chance! :grinning:

Ok, unfortunately I can't find any list of the Admin/Opr states in the few [Nokia docs](https://documentation.nokia.com/acg/23-7-2/books/md-cli-part-ii/m1135-vlan-ranges-md-cli.html#ariaid-title6) I found that have the plain `show service sap-using` command listed.

If we don't have test data for the Admin state as `Prun`, should risk adding something made up/fabricated?
_(my thought is not unless we can find it in docs or a live example)_

jorlandobr: Yes, by chance! We never caught that state before. A colleague asked for help since the data collection was failling on this router and I caught that line. Made the changes and here we are :-) And a few days after even that one router didn't had the status anymore.

So, since I don't have real data, elliminate that change and keep only for operational state.

mjbear: > So, since I don't have real data, elliminate that change and keep only for operational state.

Agreed
I'll make that modification

mjbear (approved):

PR #2295, opened 2026-03-10
Add aethra_atosnt show ip routemap work

ppetilli:
adding an additional vendor and os version (aethra atosnt) with a first command "show ip routemap work"


ntc_templates/templates/index
```diff
@@ -11,6 +11,8 @@
 #
 Template, Hostname, Platform, Command
 
+aethra_atosnt_show_ip_routemap_work.textfsm, .*, aethra_atosnt, sh[[ow]] ip routemap work
```
mjbear: @ppetilli 
Do the words `ip`, `routemap`, and `work` need to be completely typed out?

If there's a shorthand that also works we could account for that too (like is done for `show`), but only if it is a supported CLI syntax.
(**Edit:** I didn't find an example in the [atosnt manual](https://arcutronix.com/customer/10_Manuals/ATOSNTUserManual%20%206.1.12.pdf).)

First I've heard of Aethra. It doesn't look like [Netmiko](https://ktbyers.github.io/netmiko/PLATFORMS.html) has details on Aethra equipment, but [Scrapli](https://scrapli.github.io/scrapli_community/reference/aethra/atosnt/aethra_atosnt/) does.

ntc_templates/templates/aethra_atosnt_show_ip_routemap_work.textfsm
```diff
@@ -0,0 +1,15 @@
+Value Filldown MAP_NAME (\S+)
+Value Required ACTION (permit|deny)
+Value Required SEQUENCE (\d+)
+Value TYPE (\S+)
+Value VALUE (.+)
+
+Start
+  ^LIST OF MAPS -> Maps
```
mjbear: Matching the white space with a regex for future proofing.

```suggestion
  ^LIST\s+OF\s+MAPS -> Maps
```

tests/aethra_atosnt/show_ip_routemap_work/aethra_atosnt_show_ip_routemap_work.raw
```diff
@@ -0,0 +1,18 @@
...
+LIST OF COMMUNITY-SETS
+COMMUNITY-SET NAME
+LOCALPREF_135       1111:135
+
+LIST OF EXT-COMMUNITY-SETS
+Empty list
+
+LIST OF CLASSIFIERS
+adver_subs
+  permit 10     match-ip              1.1.1.1/29
```
mjbear, line 11: Do you have other examples of adver_subs where the action isn't permit?

For test coverage it would be good to have varied output/conditions.

ntc_templates/templates/index
ppetilli: Indeed commands need to be typed out completely. even show (i will correct the code) 
The router I tested this on is a TMC-R3305 (from telco systems: https://www.telco.com/wp-content/uploads/2021/03/TMC-R3305_Datasheet.pdf). This device runs ATOSNT operating system from Aethra. (as far as i could find).

tests/aethra_atosnt/show_ip_routemap_work/aethra_atosnt_show_ip_routemap_work.raw
ppetilli: @mjbear I have added the example.
br

tests/aethra_atosnt/show_ip_routemap_work/aethra_atosnt_show_ip_routemap_work.raw
```diff
@@ -0,0 +1,20 @@
+Show of SOMEDEVICE-73493 ip routemap
+LIST OF COMMUNITY-SETS
+COMMUNITY-SET NAME
+LOCALPREF_135       1111:135
+
+LIST OF EXT-COMMUNITY-SETS
+Empty list
+
+LIST OF CLASSIFIERS
+adver_subs
```
mjbear, line 10: @ppetilli 
Could there be an instance where the classifier name is something other than `adver_subs`?

ppetilli: Hi @mjbear,
You were right to question this. I verified it directly on a device and confirmed that the classifier name can be up to 16 characters.
***********************************
add ip routemap CLASSIFIER ?

add command parameters:
  classifier name  [Any value (max 16 chars)]
***********************************
I’ve updated the code accordingly to align with this behavior.
Br,
Pieter

mjbear: @ppetilli 
On line 14 (in the Start state) of the textfsm template, is a single classifier `adver_subs`.

If there's ever a different classifier name (if it's possible) then the template is not likely to match the classifier (regarding line 10 of this raw file).

I'd suggest another classifier name be added to the device configuration and update the corresponding raw output file. Thoughts?

(I don't have Aethra atosnt equipment, so I could be mistaken.)

ppetilli: @mjbear 
i must have done something wrong when commiting my last changes as i didn't intend to have adver_subs in the starte state. I checked the manual and tested this on a device.   ^\S{1,16}$$ would be the better choice.
Commited (correct) changes.

mjbear: I took a look and although that `^\S{1,16}$$` regex is pretty loose ... it won't inadvertently match anything else in the test data (other than those classifiers). Once another pattern triggers the state transition to `Maps`, at that point this regex isn't a concern anymore (since it's not in the `Maps` state).

Thank you @ppetilli!

ntc_templates/templates/aethra_atosnt_show_ip_routemap_work.textfsm
```diff
@@ -0,0 +1,23 @@
...
+Value VALUE (.+)
+
+Start
+  ^Show\s+of\s+\S+\s+ip\s+routemap$$
+  ^LIST\s+OF\s+(EXT-)?COMMUNITY-SETS$$
+  ^COMMUNITY-SET\s+NAME$$
+  ^\S+\s+\d+:\d+$$
+  ^Empty\s+list$$
+  ^LIST\s+OF\s+CLASSIFIERS$$
+  ^\S{1,16}$$
```
mjbear, line 15: Adding a comment for good measure.
(This will remind us it's for the classifiers if we have to work on it later.) :wink:

```suggestion
  # classifier names
  ^\S{1,16}$$
```

mjbear (approved):

PR #2311, opened 2026-05-04
hp_procurve_show_interfaces_brief: support stacked port names

julmanglano:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT
hp_procurve, `show interfaces brief`

##### SUMMARY
The `PORT` capture group was anchored to `[a-zA-Z0-9\-]+`, which excludes
the `/` character used in stacked-switch port naming.

On a stacked HP/Aruba ProCurve switch, ports are named `1/1`, `1/2`,
…, `2/1`, etc. None of those names match the existing regex, so the
`showintbrief2` state hits its catch-all `^. -> Error` rule and parsing
aborts on the very first interface.

This PR broadens the regex to `[a-zA-Z0-9\-/]+`, allowing the slash.
No other capture groups change, no schema change, and the five existing
fixtures (standalone-switch outputs) continue to pass.

The new `show_interfaces_brief2.{raw,yml}` fixture covers an 84-row
stacked output with a mix of `1/N`, `2/N`, and trunk-named ports.


mjbear:
@julmanglano 
Thank you for the PR!

I made a couple of minor changes.

* Reduced some of the duplication from the test data file
* Renamed the raw test data name be more descriptive (not a requirement, but several other files had names we purposefully set in past PRs so I adjusted this one)

mjbear (approved):

PR #2316, opened 2026-05-08
aruba_aoscx_show_interface: capture Filtered counters and tolerate L3/VRF/subinterface output

julmanglano:
##### ISSUE TYPE
 - Bugfix Pull Request
 - Enhancement Pull Request

##### COMPONENT
aruba_aoscx, `show interface`

##### SUMMARY
Three changes to `aruba_aoscx_show_interface`:

1. **New `Filtered` counters.** Adds `RX_FILTERED` / `TX_FILTERED`
   capture groups + a matching `^\s*Filtered` rule in the
   `Statistic` state. AOS-CX emits these on switches running L3
   ACLs / classifier policies; the existing template ignored them.

2. **L3 / aggregate / subinterface tolerance.** Adds three
   no-op match lines so previously-fatal output no longer hits
   `^. -> Error`:
   - `^\s*Subinterface\s+count` — emitted on parent interfaces
     that carry subinterfaces.
   - `^\s*active-gateway\s` — emitted on interfaces with
     active-gateway/VRRP-style virtual IPs.
   - `^\s*VRF\s+name\s+is\s+\S+` — emitted on interfaces bound to
     a non-default VRF.

3. **Two new fixtures.**
   - `show_interface5.{raw,yml}` — exercises the `Filtered`
     counters, an L3-routed interface block, an admin-down
     header variant, and a `No XCVR` transceiver state.
   - `show_interface6.{raw,yml}` — two loopbacks in non-default
     VRFs (`VRF_RED` / `VRF_BLUE`), exercising the new VRF
     swallow rule.

##### BREAKING CHANGE
The two new `Value` declarations `RX_FILTERED` and `TX_FILTERED`
add `rx_filtered` / `tx_filtered` keys to every parsed record.
Existing fixtures (`show_interface{,1-4}.yml`) have been updated
with empty-string defaults to reflect the new schema; consumers
already defending against unknown keys are unaffected, but anyone
diffing against a fixed key set will see two new keys.

##### NOT INCLUDED
VRF *capture* (turning the VRF line into a `Value` rather than a
swallow) is deferred — it would require updating every existing
fixture again and the user does not currently have a captured
data sample for every interface variant. The swallow rule is
sufficient to unblock parsing on VRF-bound interfaces; promoting
to a capture group can be done in a follow-up.


matt852 (approved):
Breaking change, so holding off on merging until ready for a new release.

mjbear:
> ##### BREAKING CHANGE
> 
> The two new `Value` declarations `RX_FILTERED` and `TX_FILTERED` add `rx_filtered` / `tx_filtered` keys to every parsed record. Existing fixtures (`show_interface{,1-4}.yml`) have been updated with empty-string defaults to reflect the new schema; consumers already defending against unknown keys are unaffected, but anyone diffing against a fixed key set will see two new keys.

Would this be a breaking change though?

It's backwards compatible (something that used an old template would be oblivious to a new key). It's not like a key name changed or the data type (scalar or string vs list, etc).

julmanglano:
Thanks @mjbear — agreed, "breaking" was being over-cautious. The only consumer that'd notice is one snapshot-testing against a fixed key set; anyone indexing by key name is unaffected since no existing key was renamed, retyped, or had its value format changed.

Will skip the breaking label on similar additive PRs going forward.

@matt852 — happy to drop the release hold then?

mjbear:
> Thanks @mjbear — agreed, "breaking" was being over-cautious. The only consumer that'd notice is one snapshot-testing against a fixed key set; anyone indexing by key name is unaffected since no existing key was renamed, retyped, or had its value format changed.
> 
> Will skip the breaking label on similar additive PRs going forward.
> 
> @matt852 — happy to drop the release hold then?

I feel the breaking change label isn't needed on this PR.

Letting some time pass so @matt852 has time to review.

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
```diff
@@ -36,6 +37,7 @@ Value TX_MCAST_PACKETS (\d+)
 Value TX_BROADCAST_PACKETS (\d+)
 Value TX_ERRORS (\d+)
 Value TX_DROPPED (\d+)
+Value TX_FILTERED (\d+)
 Value TX_CRC_FCS (\S+)
 Value TX_COLLISION (\d+)
```
matt852, line 42:
```suggestion
Value TX_COLLISION (\d+)
Value VRF (\S+)
```

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
```diff
@@ -92,6 +95,8 @@ Start
   ^\s*Rx -> Rx
   ^\s*Rate\s+RX\s+TX\s+Total -> Rate
   ^\s*Statistic\s+RX\s+TX\s+Total -> Statistic
+  ^\s*active-gateway\s
+  ^\s*VRF\s+name\s+is\s+\S+
```
matt852:
```suggestion
  ^\s*VRF\s+name\s+is\s+${VRF}
```

matt852 (changes requested):
I am good with dropping the breaking change label. It came from a misunderstanding of one of the existing docs by the AI reviewer agent.

After refining the agentic reviewer further, there are a couple new changes suggested (though not required), along with a test coverage update finding, generated by Claude:

**Two of the three new tolerance rules ship without raw fixture coverage.** `^\s*VRF\s+name\s+is\s+\S+` is exercised by show_interface6.raw, but the new `^\s*Subinterface\s+count` and `^\s*active-gateway\s` rules have no `.raw` line in this PR that exercises them. Project policy is that every template change needs corresponding test data — if you can capture a real-output sample of a parent interface with subinterfaces and an interface with `active-gateway` (even a small one) and add it as a `.raw` + `.yml` pair, that closes the loop. Alternatively, drop those two rules until you have such output captured and add them in a follow-up.

julmanglano:
Thanks @matt852 —

Bundled both the required coverage and the suggested VRF capture in one push:

- `show_interface7` — `vlan2` with `active-gateway` lines (exercises the active-gateway swallow rule).
- `show_interface8` — `1/1/10` parent with `Subinterface count: 1` (exercises the Subinterface swallow rule).
- Promoted the VRF tolerance rule from swallow to capture (`Value VRF (\S+)` + rule update). Every parsed record now has a `vrf` key — empty for default-VRF interfaces, populated for the loopbacks in `show_interface6` (`VRF_RED`, `VRF_BLUE`).

All 9 fixtures regenerated; parser tests + meta tests green locally.

Pushed.

matt852 (approved):
Thanks @julmanglano — looks good to merge.

PR #2322, opened 2026-06-02
Fix alphanum transceiver Rev in arista eos sh inv

surudhi:
## What

`arista_eos_show_inventory` silently drops every transceiver whose hardware `Rev` is not purely numeric, and mis-parses rows whose serial is numeric. The `Transceiver` state records via the following where `VID` was defined numeric only.
`^\s+${PORT}\s+${DESCR}\s+${PID}\s+${SN}\s+${VID} -> Record`

  1. **Dropped rows.** Any pluggable with an alpha `Rev` (DACs, AOCs, and most optics — `A`, `A3`, `A4`) fails the record regex and is omitted entirely. 
  2. **Column shift.** When the *serial* is numeric (e.g. Acacia `DP04QSDD-...`, serial `253255691`), the greedy `DESCR` + numeric-only `VID` forced TextFSM to bind the serial as `VID`, shifting `PID`/`SN`/`VID` left by on column.

## Fix

Broaden `VID` to `(\S+)`. The greedy `DESCR` then back-tracks to leave the last three whitespace-separated tokens as `PID`/`SN`/`VID`, so alpha-rev rows record and columns align. Change is scoped to this template only.

## Testing

- Added fixture `arista_eos_show_inventory2.{raw,yml}` covering alpha-rev DAC, AOC, and optic rows plus the numeric-serial (Acacia) shift case.
- Existing `arista_eos_show_inventory` and `arista_eos_show_inventory1` fixtures still pass unchanged (their revs are numeric).


mjbear:
> `arista_eos_show_inventory` silently drops every transceiver whose hardware `Rev` is not purely numeric, and mis-parses rows whose serial is numeric. The `Transceiver` state records via the following where `VID` was defined numeric only.

Hi @surudhi
Good catch. Thank you for the PR!

This template also doesn't have [Error directives](https://github.com/networktocode/ntc-templates/blob/master/docs/user/faq.md#why-is-there-a-requirement-to-use-error-in-every-template) in the sections like it ~probably~ should. (Something probably there from before it was a project requirement.)

I suppose we keep this PR scoped to just the `VID` since adding in the Error directive part could be quite a few modifications.

If I take on a PR to add Error directives, would you be willing to test it on Arista gear? :slightly_smiling_face:

surudhi:
> > `arista_eos_show_inventory` silently drops every transceiver whose hardware `Rev` is not purely numeric, and mis-parses rows whose serial is numeric. The `Transceiver` state records via the following where `VID` was defined numeric only.
> 
> Hi @surudhi Good catch. Thank you for the PR!
> 
> This template also doesn't have [Error directives](https://github.com/networktocode/ntc-templates/blob/master/docs/user/faq.md#why-is-there-a-requirement-to-use-error-in-every-template) in the sections like it ~probably~ should. (Something probably there from before it was a project requirement.)
> 
> I suppose we keep this PR scoped to just the `VID` since adding in the Error directive part could be quite a few modifications.
> 
> If I take on a PR to add Error directives, would you be willing to test it on Arista gear? 🙂

Thanks @mjbear ! Agreed on keeping this PR scoped to just the `VID` change and leave the Error directive work as a separate effort.

And yes, I'm glad to test an Error-directive PR on Arista gear. I've got show inventory access across a good spread of EOS hardware.

mjbear (approved):

PR #2328, opened 2026-07-02
Fix cisco_ios show lldp neighbors blank-capability parsing

### LINKED ISSUE #2327: Bug-cisco_ios_show_lldp_neighbors
```
Capability codes:
    (R) Router, (B) Bridge, (T) Telephone, (C) DOCSIS Cable Device
    (W) WLAN Access Point, (P) Repeater, (S) Station, (O) Other

Device ID           Local Intf     Hold-time  Capability      Port ID
switch-01-2960      Gi2/0/6        105        B               Gi2/0/51
switch-01-2960      Gi3/0/6        94         B               Gi1/0/51
ABC-Ultra-210W      Gi2/0/16       110                        Port 6
ABC-Ultra-210W      Gi2/0/16       102                        Port 3
ABC-Ultra-210W      Gi2/0/16       112                        Port 6
crq-switch-02c-1    Te3/1/2        116        B,R             port-002-00000
tg-DT-argh-25       Gi5/0/44       2770                       c843.09a2.0f03


Total entries displayed: 7
```

Sample output it creates
```
    {
        "NEIGHBOR_NAME": "ABC-Ultra-210W",
        "LOCAL_INTERFACE": "Gi2/0/16",
        "CAPABILITIES": "Port",
        "NEIGHBOR_INTERFACE": "6"
    },
```

You can see its messing up because of the lack of a capability.  Not really sure the best way to fix this. I can make the textfsm work but via AI i have no idea if the changes its making are good or bad so i did not try to make a PR for this. 

This was on a WS-C3750X-48P 
running: Version 15.0(2)SE10a

olasupo:
Fixes #2327

### Problem

`cisco_ios_show_lldp_neighbors` mis-parses a neighbor row when the **Capability** column is blank **and** the Port ID contains a space (for example `Port 6`). Because `CAPABILITIES` was `(\S*)`, it greedily consumed the first word of the Port ID:

| Raw row | Parsed (before) |
|---|---|
| `ABC-Ultra-210W  Gi2/0/16  110  <blank>  Port 6` | `capabilities="Port", neighbor_interface="6"` |

A blank-capability row whose Port ID is a single token (e.g. `Gi0/1` or a MAC) already parsed correctly — only multi-word Port IDs were affected. Reported against a WS-C3750X running 15.0(2)SE10a.

### Fix

- Constrain `CAPABILITIES` to the documented LLDP capability-code letters (`[BRCWPSTO,]+`) and make the field optional. It can no longer match a Port ID word such as `Port`, because the lowercase continuation breaks the match while the capability codes are always uppercase.
- Allow `NEIGHBOR_INTERFACE` to hold a multi-word Port ID.

After the fix the same row parses as `capabilities="", neighbor_interface="Port 6"`.

### Tests

Added `cisco_ios_show_lldp_neighbors3.raw`/`.yml` from the reporter's capture, covering blank capabilities with single-word, multi-word (`Port 6`) and MAC-address Port IDs, plus normal and multi-code (`B,R`) rows. The two existing fixtures are unchanged and still pass, and the full test suite passes.


matt852 (changes requested):
**Breaking Change:** No

Thanks @olasupo. This is a clean fix with great fixture coverage of the blank-capability edge cases. One suggestion before merge:

- **Add `^. -> Error` termination (and the supporting header/blank-line rules) to both states.** Our parser conventions ask that each state end with `^. -> Error` so every line of raw output is accounted for, and since this PR already reworks both `LLDP` rules it's a natural moment to bring the template up to convention. In `ntc_templates/templates/cisco_ios_show_lldp_neighbors.textfsm`, the `Start` state:

  ```diff
   Start
     ^Device.*ID -> LLDP
     # Capture time-stamp if vty line has command time-stamping turned on
     ^Load\s+for\s+
     ^Time\s+source\s+is
  +  ^Capability codes:
  +  ^\s+\(
  +  ^\s*$$
  +  ^. -> Error
  ```

  And the `LLDP` state (the `^Total\s+entries\s+displayed` and `^\s*$$` rules must sit above the `^${NEIGHBOR_NAME}` rule, which otherwise matches nearly any line):

  ```diff
   LLDP
     ^${NEIGHBOR_NAME}\s*${LOCAL_INTERFACE}\s+\d+\s+(?:${CAPABILITIES}\s+)?${NEIGHBOR_INTERFACE}\s*$$ -> Record
  +  ^Total\s+entries\s+displayed
  +  ^\s*$$
     ^${NEIGHBOR_NAME}
     ^\s+${LOCAL_INTERFACE}\s+\d+\s+(?:${CAPABILITIES}\s+)?${NEIGHBOR_INTERFACE}\s*$$ -> Record
  +  ^. -> Error
  ```

  All existing and new fixtures pass unchanged with these additions.

olasupo:
Thanks @matt852 — good call, done in 51dc609.

Both states now terminate with `^. -> Error`, with the supporting rules added:

- **`Start`:** `^Capability\s+codes:`, `^\s+\(`, and `^\s*$$` (the banner/legend and blank line ahead of the `Device ID` header).
- **`LLDP`:** `^Total\s+entries\s+displayed` and `^\s*$$` placed **above** the greedy `^${NEIGHBOR_NAME}` rule so they're matched explicitly instead of being swallowed by it.

All existing fixtures plus the new blank-capability fixture pass unchanged, and I confirmed the `Start` `^. -> Error` actually trips when fed an unexpected banner line — so every raw line is genuinely accounted for now. Full test suite is green.

matt852 (approved):

PR #2335, opened 2026-07-28
Add Fsas Si-R show vrrp brief

caribouHY:
I Added a template for Fasa Si-R  `show vrrp brief` .


mjbear:
> I Added a template for Fasa Si-R `show vrrp brief` .

Hi @caribouHY
Rather than parse a "brief" command, is there more verbose VRRP command we could instead template?

The thought is we maintain one template that contains all the VRRP information and one template to maintain in the long run (and not two -- verbose and brief).

caribouHY:
Thanks for the suggestion. I’ll create a template for the more verbose VRRP command instead, and remove this PR.

PR #2336, opened 2026-07-29
Add extra fields to aruba aoscx show interface

Ardeck:
Enhance aruba_aoscx_show_interface parser with the capture of additional fields

## Summary

Enhance the Aruba AOS-CX `show_interface` parser by capturing additional fields that are already present in the command output.

Breaking change : 
- prefix length              -> split from ip_address
- secondary_ip_prefix_length              -> split from secondary_ip_address

Reason : standardization, IP_address is a valid IPv4 address and not a CIDR notation

Added fields:
- active_gateway_ip
- active_gateway_ip_mac
- active_gateway_source_mac
- leader_follower_mode
- mdi_mode
- subinterface_count

Also improved IPv4 address matching to tolerate trailing whitespace.

## Validation

- Regenerated all associated YAML test fixtures
- Verified parsing across all available Aruba AOS-CX samples
- Test suite passed successfully


matt852 (changes requested):
## Recommendation: Changes Required
**Breaking Change:** Yes

- `ntc_templates/templates/aruba_aoscx_show_interface.textfsm` narrows `Value IP_ADDRESS` from `(\S+)` to `(\d+\.\d+\.\d+\.\d+)`, so the existing `ip_address` key changes value format from `10.1.2.1/24` to `10.1.2.1`.
- 10 existing `ip_address` values change across `tests/aruba_aoscx/show_interface/show_interface2.yml`, `show_interface3.yml`, `show_interface4.yml` (5), `show_interface6.yml` (2), and `show_interface7.yml`.

Thanks @Ardeck — nice to see these fields picked up, and thanks for regenerating all nine fixtures. A few things to address before this can merge:

- **Please flag the breaking change in the PR description.** Splitting the prefix out of `ip_address` changes the value of an existing key for every routed interface, which breaks downstream consumers that feed that value straight into something like `ipaddress.ip_interface()`. The description explains the mechanism ("netmask -> split from ip_address") but doesn't call it out as breaking — could you add that explicitly? It'll need to go out with a major release alongside the other queued breaking changes.

- **`NETMASK` should be `PREFIX_LENGTH`.** The captured value is a CIDR prefix length, not a netmask, and every other template doing this same `IPv4 address x.x.x.x/NN` split uses `PREFIX_LENGTH` — `allied_telesis_awplus_show_interface`, `cisco_nxos_show_interface`, `cisco_ios_show_interfaces`, `ruckus_fastiron_show_interfaces`, plus `aruba_aoscx_show_ip_route_all-vrfs` and `aruba_aoscx_show_interface_mgmt` on this same platform.

  In `ntc_templates/templates/aruba_aoscx_show_interface.textfsm`, the Value declaration:

  ```diff
  -Value NETMASK (\d+)
  +Value PREFIX_LENGTH (\d+)
  ```

  And the matching rule in the `Start` state:

  ```diff
  -  ^\s*IPv4\s+address\s+${IP_ADDRESS}(/${NETMASK})?\s*$$
  +  ^\s*IPv4\s+address\s+${IP_ADDRESS}(/${PREFIX_LENGTH})?\s*$$
  ```

  Then in the fixtures:

  ```diff
  -    netmask: "24"
  +    prefix_length: "24"
  ```

  Apply the same key rename to every other `netmask:` line in all nine fixture files, or just regenerate with `gen-yaml-folder` / `clean-yaml-folder`.

- **Please expand the `GW_` abbreviations.** These are the only `GW_`-prefixed values in the library — the project convention is full names. `ACTIVE_GATEWAY_*` also keeps them distinct from the default-gateway sense `IPV4_GATEWAY` has in `aruba_aoscx_show_interface_mgmt`.

  The Value declarations:

  ```diff
  -Value GW_IP (\S+)
  -Value GW_IP_MAC (\S+)
  -Value GW_SOURCE_MAC (\S+)
  +Value ACTIVE_GATEWAY_IP (\S+)
  +Value ACTIVE_GATEWAY_MAC_ADDRESS (\S+)
  +Value ACTIVE_GATEWAY_SOURCE_MAC_ADDRESS (\S+)
  ```

  And the rules (this also picks up the whitespace point below):

  ```diff
  -  ^\s*active-gateway L3 source mac ${GW_SOURCE_MAC}
  -  ^\s*active-gateway ip mac ${GW_IP_MAC}
  -  ^\s*active-gateway ip ${GW_IP}
  +  ^\s*active-gateway\s+L3\s+source\s+mac\s+${ACTIVE_GATEWAY_SOURCE_MAC_ADDRESS}
  +  ^\s*active-gateway\s+ip\s+mac\s+${ACTIVE_GATEWAY_MAC_ADDRESS}
  +  ^\s*active-gateway\s+ip\s+${ACTIVE_GATEWAY_IP}
  ```

  In `tests/aruba_aoscx/show_interface/show_interface7.yml`:

  ```diff
  -    gw_ip: "192.0.2.254"
  -    gw_ip_mac: "02:00:00:aa:bb:cc"
  -    gw_source_mac: "bc:d7:a5:aa:bb:cc"
  +    active_gateway_ip: "192.0.2.254"
  +    active_gateway_mac_address: "02:00:00:aa:bb:cc"
  +    active_gateway_source_mac_address: "bc:d7:a5:aa:bb:cc"
  ```

  Apply the same key rename to every other `gw_ip:` / `gw_ip_mac:` / `gw_source_mac:` line across all nine fixture files.

- **Literal spaces in the new rules — could you switch these to `\s+`?** Future-proofing against CLI spacing drift, matching the rest of the template. The `Subinterface count` line also has a trailing space (harmless, since TextFSM strips rule lines, but worth removing while you're in there).

  ```diff
  -  ^\s*Subinterface count\s*:\s+${SUBINTERFACE_COUNT} 
  +  ^\s*Subinterface\s+count\s*:\s+${SUBINTERFACE_COUNT}
  ```

  ```diff
  -  ^\s*MDI mode\s*:\s*${MDI}
  -  ^\s*Leader-follower\s+mode:\s+${LEADER_FOLLOWER_MODE}
  +  ^\s*MDI\s+mode\s*:\s*${MDI}
  +  ^\s*Leader-follower\s+mode\s*:\s+${LEADER_FOLLOWER_MODE}
  ```

Two more that would be good to fold in while this is open:

- **`MDI` → `MDI_MODE`.** The raw label is `MDI mode:`, the value is a mode, your own sibling field uses the suffix (`LEADER_FOLLOWER_MODE`), and `hp_procurve_show_interfaces_brief` already uses `MDI_MODE`.

  ```diff
  -Value MDI (\S+)
  +Value MDI_MODE (\S+)
  ```

  ```diff
  -  ^\s*MDI\s+mode\s*:\s*${MDI}
  +  ^\s*MDI\s+mode\s*:\s*${MDI_MODE}
  ```

  Apply the matching `mdi:` → `mdi_mode:` key rename across all nine fixture files.

- **`secondary_ip_address` still carries the prefix.** After this change the template emits `ip_address: "172.23.2.1"` + a separate prefix, but `secondary_ip_address: ["172.25.2.1/30"]` — two address formats out of one parse. Since `ip_address` is already breaking here, splitting the secondary in the same release avoids a second breaking change later.

  ```diff
  -Value List SECONDARY_IP_ADDRESS (\S+)
  +Value List SECONDARY_IP_ADDRESS (\d+\.\d+\.\d+\.\d+)
  +Value List SECONDARY_PREFIX_LENGTH (\d+)
  ```

  ```diff
  -  ^\s*IPv4\s+address\s+${SECONDARY_IP_ADDRESS}\s+secondary
  +  ^\s*IPv4\s+address\s+${SECONDARY_IP_ADDRESS}(/${SECONDARY_PREFIX_LENGTH})?\s+secondary
  ```

  I checked this against the fixtures and the list alignment holds on the multi-secondary interface (`1/1/45` → `['172.22.19.1', '172.22.22.1']` / `['30', '30']`).

Optional / FYI: the `(/${NETMASK})?` optional group has no test data for the bare-address case. I think the optional form is the right call — a required `/` would push an unmasked line into `^. -> Error` — but a line in the description noting that AOS-CX always emits the mask would head off the question.

Thanks!
(review generated with Claude)

matt852:
Additionally, with the `netmask -> split from ip_address` changes breaking the current `ip_address` field, what's the purpose for this?

mjbear:
In addition to converting to unix file format, I made several white space regex related commits.

mjbear:
> [!NOTE]
> While we're at it, we could normalize `INTERFACE_DESC` to `DESCRIPTION` (since we already have a breaking change).

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
```diff
@@ -40,46 +46,49 @@ Value TX_DROPPED (\d+)
...
+Value SUBINTERFACE_COUNT (\d+)
 
 Start
   ^\s*(Interface|Aggregate)\s+\S+\s+is -> Continue.Record
   ^\s*(Interface|Aggregate)\s+${INTERFACE}\s+is\s+${LINK_STATUS}
-  ^\s*Admin\s*state\s*is\s*${LINK_ADMIN}
-  ^\s*Link\s*transitions\s*:\s*${LINK_TRANSITIONS}
+  ^\s*Admin\s+state\s+is\s+${LINK_ADMIN}
+  ^\s*Link\s+transitions\s*:\s+${LINK_TRANSITIONS}
   ^\s*Description\s*:\s*${INTERFACE_DESC}
```
matt852:
```suggestion
  ^\s*Description\s*:\s*${DESCRIPTION}
```
This will also need the YAML test files regenerated with the updated `description` variable.

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
```diff
@@ -89,35 +98,37 @@ Start
   ^\s*rate-limit\s+broadcast
   ^\s*rate-limit\s+multicast
   ^\s*L3\s+Counters
-  ^\s*MDI mode
-  ^\s*Leader-follower\s+mode
-  ^\s*Rate collection interval
+  ^\s*MDI\s+mode\s*:\s+${MDI_MODE}
+  ^\s*Leader-follower\s+mode:\s+${LEADER_FOLLOWER_MODE}
```
matt852, line 102:
```suggestion
  ^\s*MDI\s+mode\s*:\s+${MDI_MODE}
  ^\s*MDI\s+mode
  ^\s*Leader-follower\s+mode\s*:\s+${LEADER_FOLLOWER_MODE}
  ^\s*Leader-follower\s+mode
```
Add a fallback rule in case either value is empty, or if Leader-follower mode has a space before the colon, they fail to parse where it used to work.

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
```diff
@@ -89,35 +98,37 @@ Start
...
+  ^\s*Rate\s+collection\s+interval
   ^\s*Energy-Efficient
   ^\s*Rx -> Rx
   ^\s*Rate\s+RX\s+TX\s+Total -> Rate
   ^\s*Statistic\s+RX\s+TX\s+Total -> Statistic
-  ^\s*active-gateway\s
+  ^\s*active-gateway\s+L3\s+source\s+mac\s+${ACTIVE_GATEWAY_SOURCE_MAC}
+  ^\s*active-gateway\s+ip\s+mac\s+${ACTIVE_GATEWAY_IP_MAC}
+  ^\s*active-gateway\s+ip\s+${ACTIVE_GATEWAY_IP}
```
matt852, line 110:
```suggestion
  ^\s*active-gateway\s+ip\s+${ACTIVE_GATEWAY_IP}
  ^\s*active-gateway
```
Fallback rule to keep IPv6 forms parsing correctly.

matt852 (changes requested):
A couple other suggestions. YAML test files will need to be regenerated if the `INTERFACE_DESC` --> `DESCRIPTION` change is accepted (as suggested by @mjbear)

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
mjbear: I changed the description capture group name in the template and regenerated the structured test files (YAML).

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
mjbear: Do we have test data for both of these lines being blank?

In the present data, MDI mode is either `MDIX` or `none`.

Maybe we need more data on the Leader-follower mode.

ntc_templates/templates/aruba_aoscx_show_interface.textfsm
mjbear: Similar to MDI and Leader-follower modes -- do we have test data to support this?

It would seem when set, active-gateway is present in the output and if not it is absent from the output. (Though our test data output is likely from different versions and maybe the behavior has changed across versions.)

@Ardeck Would you please confirm this?

mjbear: > Similar to MDI and Leader-follower modes -- do we have test data to support this?
> 
> It would seem when set, active-gateway is present in the output and if not it is absent from the output. (Though our test data output is likely from different versions and maybe the behavior has changed across versions.)
> 
> @Ardeck Would you please confirm this?

@Ardeck Could I get some confirmation to the above question?

mjbear:
_(akin to a "note to self" :sweat_smile:)_

I'd venture PR #2360 could be merged before this one (since it doesn't include breaking changes like this does).
Heads up in case of any merge conflicts or the need to rebase.

lboue:
@mjbear I've added the tests requested in the other PR:
- https://github.com/networktocode/ntc-templates/pull/2360

PR #2337, opened 2026-07-29
Add hp_comware_display_version

Ardeck:
This PR adds a new TextFSM parser for `display version` on HP / HPE Comware platforms (Comware v5 and v7).

#### Supported Hardware & Topologies

* **Chassis / Modular Switches:** Parses multi-slot chassis (e.g., HP 7500 series) with `MPU`/`LPU` boards, properly extracting individual slot uptime, hardware types, and memory breakdown across chassis/slots.
* **Standalone / Fixed-Configuration Switches:** Correctly parses single-unit devices (e.g., HP 5800 / 5130 series) without relying on explicit `Slot X:` headers.

#### Parsing Highlights

* Captures full `UPTIME` and `SLOT_UPTIME` strings without truncation.
* Uses state transitions (`Start` $\rightarrow$ `SlotSection`) to handle slot flushing cleanly via `Continue.Record`, preventing missing or overwritten slot records (e.g., Slot 0).
* Captures global software metadata (`SOFTWARE_VERSION`, `RELEASE`, `MODEL`, `BOOT_IMAGE`, `SYSTEM_IMAGE`, etc.) alongside hardware slot attributes (`DRAM`, `NOR_FLASH`, `NAND_FLASH`, `BOARD_TYPE`, `BOOTROM_VERSION`).

---

### Motivation

HP Comware devices lacked a standardized `display version` parser in `ntc-templates`. Adding this template enables structured parsing of hardware inventory, software release details, and device/slot uptimes for network automation workflows using Nornir, Ansible, or Netmiko.

---

### Testing

* [x] Test `.raw` output samples added under `tests/hp_comware/display_version/` (covering both chassis and standalone switches).
* [x] Test `.yml` files generated via `invoke gen-yaml-folder`.
* [x] Ran local test suite (`invoke pytest`) — **100% passing**.
* [x] Validated YAML syntax (`invoke yamllint`).
* [x] Registered command in `ntc_templates/templates/index`.
* [x] Tested on 900+ devices 

---

### Checklist

* [x] Template added in `ntc_templates/templates/`
* [x] Template registered in `ntc_templates/templates/index`
* [x] Test cases (`.raw` and `.yml`) added in `tests/hp_comware/display_version/`
* [x] All tests passing locally


matt852 (changes requested):
## Recommendation: Changes Required
**Breaking Change:** No

Thanks @Ardeck — this is a lot of ground covered (Comware v5 standalone, v7 standalone, and multi-chassis all in one template), and the three fixtures are genuinely good coverage. Two things to address before merge:

- **The two `Compiled` rules collide, so `system_compile_date` is never populated.** `^\s+Compiled\s+${BOOT_COMPILE_DATE}` and `^\s+Compiled\s+${SYSTEM_COMPILE_DATE}` are byte-identical, and TextFSM stops at the first match — so the boot rule consumes both lines. `boot_compile_date` ends up holding the *system* image's date and `system_compile_date` stays empty. Fixture 2 hides it because both images share a timestamp, but on a device where they differ the output is wrong. Splitting the two blocks into their own states fixes it.

  In `hp_comware_display_version.textfsm`, the `Start` state:

  ```diff
     ^Last reboot reason\s*:\s*${LAST_REBOOT_REASON}
  -  ^Boot image:\s+${BOOT_IMAGE}
  -  ^Boot image version:\s+${BOOT_IMAGE_VERSION},\s+Release\s+${BOOT_RELEASE}
  -  ^\s+Compiled\s+${BOOT_COMPILE_DATE}
  -  ^System image:\s+${SYSTEM_IMAGE}
  -  ^System image version:\s+${SYSTEM_IMAGE_VERSION},\s+Release\s+${SYSTEM_RELEASE}
  -  ^\s+Compiled\s+${SYSTEM_COMPILE_DATE}
  +  ^Boot image:\s+${BOOT_IMAGE} -> BootImage
  +  ^System image:\s+${SYSTEM_IMAGE} -> SystemImage
     # Standalone switch rules (e.g. HP 5800)
  ```

  And two new states, inserted just above `SlotSection`:

  ```diff
  +BootImage
  +  ^Boot image version:\s+${BOOT_IMAGE_VERSION},\s+Release\s+${BOOT_RELEASE}
  +  ^\s+Compiled\s+${BOOT_COMPILE_DATE} -> Start
  +  ^System image:\s+${SYSTEM_IMAGE} -> SystemImage
  +
  +SystemImage
  +  ^System image version:\s+${SYSTEM_IMAGE_VERSION},\s+Release\s+${SYSTEM_RELEASE}
  +  ^\s+Compiled\s+${SYSTEM_COMPILE_DATE} -> Start
  +
   SlotSection
  ```

  And the fixture it corrects, in `tests/hp_comware/display_version/hp_comware_display_version2.yml`:

  ```diff
  -    system_compile_date: ""
  +    system_compile_date: "Oct 12 2023 11:00:00"
  ```

- **Literal single spaces in the match patterns should be `\s+`.** Comware pads these columns differently across releases, so `BOARD TYPE:` / `Uptime is` / `Boot image:` etc. are brittle as written. This one doesn't change any parsed output — no fixture regeneration needed.

  In the `Start` state:

  ```diff
  -  ^.*Comware Software,\s+Version\s+${SOFTWARE_VERSION},\s+Release\s+${RELEASE}
  -  ^${MODEL}\s+uptime is\s+${UPTIME}
  -  ^Last reboot reason\s*:\s*${LAST_REBOOT_REASON}
  -  ^Boot image:\s+${BOOT_IMAGE}
  -  ^Boot image version:\s+${BOOT_IMAGE_VERSION},\s+Release\s+${BOOT_RELEASE}
  +  ^.*Comware\s+Software,\s+Version\s+${SOFTWARE_VERSION},\s+Release\s+${RELEASE}
  +  ^${MODEL}\s+uptime\s+is\s+${UPTIME}
  +  ^Last\s+reboot\s+reason\s*:\s*${LAST_REBOOT_REASON}
  +  ^Boot\s+image:\s+${BOOT_IMAGE}
  +  ^Boot\s+image\s+version:\s+${BOOT_IMAGE_VERSION},\s+Release\s+${BOOT_RELEASE}
  ```

  ```diff
  -  ^System image:\s+${SYSTEM_IMAGE}
  -  ^System image version:\s+${SYSTEM_IMAGE_VERSION},\s+Release\s+${SYSTEM_RELEASE}
  +  ^System\s+image:\s+${SYSTEM_IMAGE}
  +  ^System\s+image\s+version:\s+${SYSTEM_IMAGE_VERSION},\s+Release\s+${SYSTEM_RELEASE}
  ```

  ```diff
  -  ^BootRom Version is ${BOOTROM_VERSION}
  +  ^BootRom\s+Version\s+is\s+${BOOTROM_VERSION}
  ```

  And in the `SlotSection` state:

  ```diff
  -  ^Uptime is ${SLOT_UPTIME}
  -  ^BOARD TYPE:\s+${BOARD_TYPE}
  +  ^Uptime\s+is\s+${SLOT_UPTIME}
  +  ^BOARD\s+TYPE:\s+${BOARD_TYPE}
  ```

  ```diff
  -  ^(?:Bootrom|Bootware) Version:\s+${BOOTROM_VERSION}
  -  ^BootRom Version is ${BOOTROM_VERSION}
  -  ^Patch Version\s*:\s*${PATCH_VERSION}
  -  ^Reboot Cause\s*:\s*${REBOOT_CAUSE}
  +  ^(?:Bootrom|Bootware)\s+Version:\s+${BOOTROM_VERSION}
  +  ^BootRom\s+Version\s+is\s+${BOOTROM_VERSION}
  +  ^Patch\s+Version\s*:\s*${PATCH_VERSION}
  +  ^Reboot\s+Cause\s*:\s*${REBOOT_CAUSE}
  ```

A few more worth picking up in the same pass:

- **The `PreStart` state never runs.** TextFSM always begins in `Start`, so neither `PreStart` rule can fire — the `HP Comware Platform Software` banner is just falling through unmatched in `Start` today. Deleting the state produces identical output.

  ```diff
  -PreStart
  -  ^HPE?\s+(?:HP\s+)?Comware\s+(?:Platform\s+)?Software,\s+Version\s+${SOFTWARE_VERSION},\s+Release\s+${RELEASE} -> Start
  -  ^HPE?\s+(?:HP\s+)?Comware\s+Platform\s+Software -> Start
  -
   Start
  ```

- **`SOFTWARE_VERSION` → `VERSION` for cross-vendor consistency.** Eleven of the 22 existing `show version` / `display version` templates use bare `VERSION` for the primary software version and none use `SOFTWARE_VERSION`. `cisco_ios_show_version` is the exact parallel — it pairs `VERSION` with `RELEASE`, same as here.

  ```diff
  -Value Filldown SOFTWARE_VERSION (\S+)
  +Value Filldown VERSION (\S+)
  ```

  Apply the same rename to the rule that references it, and to every `software_version:` line in all three fixture files (`invoke gen-yaml-folder` will regenerate them for you).

- **Some useful fields in the raw output aren't captured yet.** `Release Version:`, `PCB N Version:` / `Hardware Version is`, `CPLD N Version:`, the processor type and count (`with 1 BCM1125H Processor` — dropped entirely inside `SlotSection`), the `[SubSlot N]` module lines, and `Config Register points to ...`. Worth grabbing now since adding capture groups later is a breaking change. I left this without a diff because how to decompose these (per-index `List` values vs. flat) is your call.

- **No state ends with `^. -> Error`.** Both nearest siblings (`huawei_vrp_display_version`, `hp_procurve_show_version`) have one, and it's what would have surfaced the dropped fields above. This needs to land together with the capture work — adding it as-is would fail all three fixtures — so pair it with explicit ignore rules for the lines you decide not to capture (`^Copyright`, `^Config Register`, `^\[SubSlot`, `^\s*$$`).

Thanks!
(review generated with Claude)

Ardeck:
I tried to fix everything but this is dying hardware mostly, but that are still running. The most difficult part is the moving format accross platform and version.  The last commit handlings errors is also checked on a few hundreds of device and should catch most of the fields possible.

PR #2343, opened 2026-08-11
Fix alcatel_sros gnss ports with less columns

bennnnnnnn:
Small fix to support gnss pseudo ports


mjbear (approved):

tests/alcatel_sros/show_port/show_port.raw
```diff
@@ -233,6 +233,8 @@ Id            State      State   MTU  MTU  Bndl Mode Encp Type   MDIMDX
 10/2/c22      Down       Down                             conn
 10/2/c23      Down       Down                             conn
 10/2/c24      Down       Down                             conn
+A/1           Up    Yes  Up      1514 1514    - netw null xcme   MDI
+A/gnss        Down  No   Down                             gnss
```
matt852: Are the trailing white spaces at the end of line 237 intentional or accidental? It looks like the only line in this file with trailing white space.

matt852 (commented):
Looks good overall. 1 nit question posted on white space.

tests/alcatel_sros/show_port/show_port.raw
bennnnnnnn: Good catch! Copy/paste error 😅 I've corrected it!

PR #2152, opened 2025-06-30
Support Alcatel service service-using names with spaces

jorlandobr:
Some of our service-names contains spaces. Modified the regex for these cases and added test case.

Ignore other Cisco PRs, they are already addressed, but still appearing...

PR #2344, opened 2026-08-11
Fix alcatel_sros sh svc id int - names w/ spaces

jorlandobr:
Found interface names with spaces. Modified template and added test cases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ concurrency:  # Cancel any existing runs of this workflow for this same PR
 on: # yamllint disable-line rule:truthy rule:comments
   push:
     branches:
+      - "interface-name-with-spaces"
       - "master"
     tags:
       - "v*"
--- a/ntc_templates/templates/alcatel_sros_show_service_id_interface.textfsm
+++ b/ntc_templates/templates/alcatel_sros_show_service_id_interface.textfsm
@@ -1,4 +1,4 @@
-Value Required INTERFACE_NAME (\S.+\S)
+Value Required INTERFACE_NAME (\S+(?:\s+\S+)*)
 Value Required ADMIN_STATE (Up|Down)
 Value Required OPER_STATE_V4 (Up|Down)
 Value Required OPER_STATE_V6 (Up|Down)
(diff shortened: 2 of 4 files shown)


mjbear:
@jorlandobr
I pushed a change to the interface_name regex that simplifies the regex pattern.

Please feel free to test against more live gear.
Thank you! :slightly_smiling_face:

jorlandobr:
@mjbear , the modification works ok! :-)

matt852:
**Breaking Change:** No

Thanks @jorlandobr — one suggestion and two questions:

- **Use the regex form this vendor already uses for names with spaces.** `(\S.+\S)` needs at least 3 characters, so a 1- or 2-character interface name (`l3`, `e1`, `a`) now fails to parse instead of matching as it did before (as per your past PR #2152)

  In `ntc_templates/templates/alcatel_sros_show_service_id_interface.textfsm`, the Value declaration:

  ```diff
  -Value Required INTERFACE_NAME (\S.+\S)
  +Value Required INTERFACE_NAME (\S+(?:\s+\S+)*)
  ```

- **Optional — the `-` placeholder is captured as an address.** Was this intentional? The `Loopback 10 loop 2` row has no address, so the parser returns `ip_address: ["-"]`. Adding an ignore rule for that row returns an empty list instead:

  In the `Interface` state:

  ```diff
     ^${INTERFACE_NAME}\s+${ADMIN_STATE}\s+${OPER_STATE_V4}\/${OPER_STATE_V6}\s+${TYPE}\s+${PORT_SAP_ID}
  +  ^\s+-\s+-\s*$$
     ^\s+${IP_ADDRESS}\s+${PFX_STATE}
  ```

  And the matching fixture entry in `show_service_id_interface_with_spaces.yml`:

  ```diff
  -    ip_address:
  -      - "-"
  +    ip_address: []
       oper_state_v4: "Down"
       oper_state_v6: "Down"
  -    pfx_state:
  -      - "-"
  +    pfx_state: []
  ```

- **Optional — one question on the new `.raw` file.** The existing `show_service_id_interface.raw` starts with a blank line and the new one starts at the `===` banner — was that blank line absent in the output you captured?

Thanks!
(review written by Claude, modified by me)

jorlandobr:
@matt852 , missed the small named interfaces. I am on vacation and don't have acces in this period to live equipment, so I used just the raw data that I have, just a small sample.

About the placeholder, I prefer to capture the output as it is and treat it later in my programs. I don't now... maybe for some cases could make a difference (for the user) having an empty output and one that has a marker 

The empty line at the beggining was a mistake when copy/pasting the output for testing the template

matt852:
@jorlandobr Sounds good, feel free to push the updates once your back from vacation. Thank you for your submission!

jorlandobr:
@matt852 ,
added "e1" named interface, reverted regex, modified .raw and .yml files. Passed tests :+1:

PR #2345, opened 2026-08-14
Fix RouterOS 7 interface print terse

bile0026:
Updates mikrotik_routeros_interface_print_terse_without-paging.textfsm, so it parses output from newer RouterOS versions instead of erroring:

Skip the `Flags:` and `Columns:` header lines that RouterOS 7 emits before the interface rows (previously these hit `^. -> Error`).
Add three new optional values captured from the interface rows: `IFNAME, IFINDEX, and INTERFACE_ID` (from `ifname=`, `ifindex=`, and `id=`).
Add a trailing `(\s+\S+=\S+)*` catch-all so any additional unmodeled `key=value` attributes don't fail the match.
All new values are optional, so existing output continues to parse unchanged.


tests/mikrotik_routeros/interface_print_terse_without-paging/mikrotik_routeros_interface_print_terse_without-paging3.raw
mjbear, line 1: @bile0026 
Is this ***terse*** output since it contains "flags" and "columns" lines?

matt852:
**Breaking Change:** No

- **Please post the unmodified `/interface print terse` output.** In the repo's real RouterOS capture (`mikrotik_routeros_interface_print_detail_01.raw`) `ifname` is the kernel name — `ifname="eth0"` for `name="ether1"`, `ifname="ppp25889"` for `name="pppoe-out1"` — and `ifindex` differs from `id`, but in the new fixture `ifname` matches `name` and `ifindex` matches `id` in every row.

- **Which RouterOS 7 version prints `Flags:` and `Columns:` before the terse rows?** No other terse fixture in the repo has them, and every fixture that does is the tabular `print` form followed by a `#` header row.

- **Capture the `switch=` field.** The new catch-all swallows `switch=switch1` on row 4, so it never reaches the parsed output.

  In the Value block:

  ```diff
   Value INTERFACE_ID (\d+)
  +Value SWITCH (\S+)
   Value LAST_LINK_DOWN_TIME (\S+\s\S+)
  ```

  And the tail of the `Start` rule:

  ```diff
  -  ...(\s+link-downs=${LINK_DOWNS})?(\s+\S+=\S+)*\s*$$ -> Record
  +  ...(\s+link-downs=${LINK_DOWNS})?(\s+switch=${SWITCH})?(\s+\S+=\S+)*\s*$$ -> Record
  ```

- **Rename `IFNAME` and `IFINDEX`.**  As per `docs/dev/data_model.md`

  ```diff
  -Value IFNAME (\S+)
  -Value IFINDEX (\d+)
  +Value INTERFACE_NAME (\S+)
  +Value INTERFACE_INDEX (\d+)
  ```

  And in the `Start` rule:

  ```diff
  -  ...(\s+ifname=${IFNAME})?(\s+ifindex=${IFINDEX})?...
  +  ...(\s+ifname=${INTERFACE_NAME})?(\s+ifindex=${INTERFACE_INDEX})?...
  ```

  Then rename the `ifname:` and `ifindex:` keys in all three `.yml` files.

- **Decide whether `mikrotik_routeros_interface_print_detail.textfsm` should expose these fields too.** It matches `ifname=`, `ifindex=`, and `id=` without capturing them, so the two interface templates would return different schemas for the same data.

Thanks!
(review written by Claude, modified by me)

tests/mikrotik_routeros/interface_print_terse_without-paging/mikrotik_routeros_interface_print_terse_without-paging3.raw
bile0026: Still working some of this out. I believe there might be different results based on the type of device or the version of routeros. Trying to narrow this down.

PR #2347, opened 2026-08-17
Test space consistency and unused states

matt852:
This PR adds a new pytest file with a few new checks:

* test_template_compiles - Tests that all submitted templates successfully compile.
* test_rule_indentation - TextFSM compiles with 1 space, 2 spaces, or a tab. This rule enforces 2 spaces as a standard.
* test_no_unused_states - Verifies that every defined state is reachable from another rule.

Additionally, this PR cleans up multiple files not currently using the standard 2 spaces, as well as ones that have unreachable "Done" states at the end of them.


mjbear (approved):
Testing whether the template compiles (is valid TFSM) is handy.
Good work.

PR #2355, opened 2026-08-27
Fix nxos LLDP detail when Port id is not present

jacqueswindt-gif:
### Problem

On NX-OS releases that do not emit a `Port id:` line, `show lldp neighbors detail` parses to **zero rows**. Not a partial result — the whole command comes back empty, and nothing raises.

`NEIGHBOR_INTERFACE` is `Required` and is sourced from:

```
  ^[Pp]ort\s[idID]{2}\:\s${NEIGHBOR_INTERFACE}$$
```

That rule is anchored at the start of the line, so `Local Port id:` correctly does *not* match it (it is caught by the rule below). On a release that emits `Local Port id:` and `Port Description:` but no bare `Port id:` line, `NEIGHBOR_INTERFACE` is therefore never filled — and because it is `Required`, every block is discarded.

### Where this was seen

Two Cisco Nexus 3064s, same estate:

| release | `show lldp neighbors detail` | parsed |
|---|---|---|
| **6.0(2)U2(9Z)** | 250 lines, 18 neighbours | **0 rows** |
| 7.0(3)I4(9) | 208 lines, 14 neighbours | 14 rows |

On the 6.x switch all 18 blocks carry `Local Port id:` and `Port Description:`, and **none** carries `Port id:`.

The silence is the awkward part: an empty parse from a switch with 18 LLDP neighbours is indistinguishable from a switch that genuinely has none, so a consumer reports "no LLDP adjacency" and is believed.

### Fix

Drop `Required` from `NEIGHBOR_INTERFACE`. A block then records with the fields the device did send, instead of the command yielding nothing.

`NEIGHBOR_NAME` and `LOCAL_INTERFACE` stay `Required`, so they still guard against recording junk.

### Verification

- Added `cisco_nxos_show_lldp_neighbors_detail_no_port_id.{raw,yml}` — the block shape that release emits. It parses to 0 rows before this change and 2 rows after.
- Both existing `show_lldp_neighbors_detail` fixtures parse **byte-identically** before and after; releases that already worked are unaffected.
- Full suite green: `1896 passed` for the parsed-reference tests and `2929 passed` for index/testcase/capture-group.

### Not included, deliberately

`Port Description:` is present in this output and currently discarded, and on this release it happens to hold the remote interface name. Mapping it to `NEIGHBOR_INTERFACE` would be wrong in general — in the existing `..._detail2.raw` fixture, `Port id: Ethernet1/52` sits beside `Port Description: SW-NEXUS-UPLINK`, where the description is a description. Adding a separate `PORT_DESCRIPTION` value would be a reasonable follow-up but changes the parsed schema for every consumer, so it is left out of a bug fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mjbear:
The majority of the test data only covers Cisco LLDP neighbors.
This could be the case that Mikrotik decided to do things a bit different in the LLDP PDUs (since [`Port Id` is a mandatory LLDP TLV](https://en.wikipedia.org/wiki/Link_Layer_Discovery_Protocol#Frame_structure)).

It's too bad we don't have test data with other vendors' NOSes as neighbors (Juniper, Arista, HPE, etc) as a comparison. That could be something to add in the future.

It seems as if Mikrotik may be using port description in lieu of port id. Best case we can capture the description as well. Since it's a dictionary and they are referenced by key name this should not be a data structure issue.

Other existing LLDP templates capture the neighbor port description so the capture group name is normalized (matches the other templates).

I pushed some commits to this branch, including port description and to add an Error directive.

PR #2356, opened 2026-08-27
Fix nxos LLDP neighbors with spaces in sys name

jacqueswindt-gif:
### Problem

`show lldp neighbors` silently drops any neighbour whose LLDP system name contains a space.

`NEIGHBOR_NAME` is `(\S+)`, so on a row like:

```
Office Access Point  Eth1/5          120        BR          wlan0
```

the record rule cannot match — `Office` is taken as the name, `Access` as the local interface, and the `\s+\d+` that follows then meets `Point` instead of a hold-time. The bare-name rule `^${NEIGHBOR_NAME}$$` does not match either, because the line is not a single token. The row matches nothing and is discarded.

Every other neighbour parses normally, so the result is a short list rather than an error. On the switch where this was found, 17 of 18 neighbours came back and the missing one was a device named with a space.

### Fix

Two changes, and both are needed:

1. `NEIGHBOR_NAME` becomes `(\S+(?:\s\S+)*?)` — one or more whitespace-separated tokens, non-greedy. The record rules anchor on the hold-time digits that follow the local interface, which is what stops the name swallowing the rest of the line.

2. A rule consuming `Total entries displayed` is added at the top of the `LLDP` state.

The second is not cosmetic. `Total entries displayed: 2` is also a line of words, so with a permissive name it matches the bare-name rule, fills `NEIGHBOR_NAME` *after* the last `Record`, and TextFSM's implicit EOF rule then emits it as an extra row. Change 1 alone regresses **both** existing fixtures for this template; consuming that line first is what makes the permissive name safe.

A non-capturing group is used so `test_unused_capture_group` stays satisfied.

### Verification

- Added `cisco_nxos_show_lldp_neighbors_name_with_space.{raw,yml}`. Before this change it parses to 1 row (the spaced name is dropped); after, 2.
- Both existing fixtures — including `..._multiline.raw`, which exercises the wrapped-long-name path this shares a rule with — parse **byte-identically** before and after: 3 rows and 15 rows respectively.
- Full suite green: `4824 passed` across the parsed-reference, index-order, testcase-exists and capture-group tests.

### Note

The wrapped-name rules are untouched. A device ID long enough for NX-OS to print it on its own line still parses through `^${NEIGHBOR_NAME}$$` followed by the indented continuation rule, which is what the `multiline` fixture covers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mjbear:
I pushed a handful of commits to adjust literal white spaces, neighbor name  regex, and add in Error directives.

~The Error directive part helped me find a bug when long names cause the data to be on a different line than the name. If desired, I can remove some of these commits and create a new PR.
(@jacqueswindt-gif If that's your preference, respond here. Thank you.)~

> [!IMPORTANT]
> After thinking about it a bit, the two-line data sets within the `LLDP` State should be their own PR. I'll revert that commit and leave the rest.

PR #2358, opened 2026-09-04
Fix wrapped supported/advertising fields in mikrotik_routeros_interface_ethernet_monitor_name_once

k-ribot:
## Summary
- `supported:`, `advertising:` and `link-partner-advertising:` on Mikrotik
  RouterOS can wrap onto multiple bare-value lines instead of one comma
  separated line, when the terminal is too narrow. The template only handled
  the single-line case.
- Converts these three fields to `List` values and adds dedicated FSM states
  so wrapped continuation lines are correctly captured into the right list.
- Adds two new raw/yml test cases (5 and 6) reproducing the wrapped format.


mjbear:
@k-ribot
What's the process like to accept code suggestions through the unyc GH organization?
Are you able to click accept on my suggestions to have them added to your PR?

PR #2360, opened 2026-09-07
Fix aruba_aoscx show interface with L3 counters

lboue:
## Description

Fix parsing of `show interface` output from ArubaOS-CX devices containing L3 counter sections.

On ArubaOS-CX `FL.10.05.0030`, the `show interface` command can return the following structure:

```text
Rx
        L3:
             0 packets, 0 bytes
Tx
        L3:
             0 packets, 0 bytes
```

- https://github.com/networktocode/ntc-templates/issues/2359

PR #2365, opened 2026-09-16
Fix missing indentation in show_interface9.raw (aruba_aoscx)

lboue:
Following feedback from @matt852 on #2360 — the .raw test file for show_interface9 was missing the original terminal indentation, since the output was reformatted (flush-left) when it was added.

This PR restores the unmodified terminal output, including leading whitespace, for the vlan1, vlan10, and lag40 interface blocks.

- No changes to the .textfsm template itself
- No changes to the parsed .yml expected output — the parsing logic is indentation-tolerant, so behavior is unaffected
- Fixes test fixture accuracy only, per maintainer request

Closes/follows up on feedback from #2360.


matt852:
Thank you!

matt852 (approved):

PR #2186, opened 2025-09-06
Fix arista_eos sh ip bgp sum for PfxAdv column

itsMrRobot:
Encountered PfxAdv option when working on Lab cEOS devices running 4.34.2F-43232954.4342F, added line 21 to account for that field as previous version encountered error without it

CLI output from device:
image

Error received without line:
image

Results after adding line to local template file:
image

PR #2366, opened 2026-09-20
Fix arista_eos sh ip bgp sum pfxadv column header

mjbear:
succeeds #2186 (incomplete PR closed by contributor)

* Adds support for PfxAdv column header
* Adds test data with PfxAdv column header present
   * Utilized Arista EOS output found online [[1]](https://labguides.testdrive.arista.com/2025.3/campus/l3ls-evpn/#lab-tasks) [[2]](https://arista.my.site.com/AristaCommunity/s/article/Example-Demonstration-1-Conditional-Advertisement-Withdrawal-of-Local-Networks-to-Backup-ISP-Session-Based-on-Primary-ISP-Session-Status-UP-DOWN)


ntc_templates/templates/arista_eos_show_ip_bgp_summary.textfsm
```diff
@@ -19,6 +19,7 @@ Start
   ^\s*Neighbor\s+V\s+AS\s+MsgRcvd\s+MsgSent\s+InQ\s+OutQ\s+Up/Down\s+State\s+PfxRcd\s*$$
   ^\s*Neighbor\s+V\s+AS\s+MsgRcvd\s+MsgSent\s+InQ\s+OutQ\s+Up/Down\s+State\s+PfxRcd\s+PfxAcc\s*$$
```
matt852, line 18: Can you add this line directly below line 20? (I'm having trouble with the spacing with the "suggested change" feature):

```diff
  ^\s*Neighbor\s+V\s+AS\s+MsgRcvd\s+MsgSent\s+InQ\s+OutQ\s+Up/Down\s+State\s+PfxRcd\s+PfxAcc\s+PfxAdv\s*$$
```

A device on the same EOS release with no neighbor descriptions emits the header without the leading Description column, and that line still fails to parse.

matt852 (commented):
Looks good, with one optional line to add to catch configs with no neighbor descriptions.

ntc_templates/templates/arista_eos_show_ip_bgp_summary.textfsm
mjbear: > A device on the same EOS release with no neighbor descriptions emits the header without the leading Description column, and that line still fails to parse.

I'm open to this suggestion and will incorporate it.

I discovered that it's possible for a lowercase `m` (maint mode) near the beginning of the line ([example](https://www.arista.com/en/um-eos/eos-maintenance-mode#xx1183612)) so that's something else to be factored in. :zany_face:

mjbear: @matt852
On the topic of your suggestion, does PfxAdv ever show up without Description?

---

I added support for maintenance mode - found online output to use for test data from a release _without pfxadv_. I had to approximate the output when description/pfxadv is present. :shrug:

PR #2367, opened 2026-09-22
arista_eos: add description field support to show_ipv6_bgp_summary template

alex4lbin:
## Problem
Parsing failed for Arista EOS `show ipv6 bgp summary` output when the header included an unexpected `Description` field.

## Solution
- added support for the `Description` field to `arista_eos_show_ipv6_bgp_summary.textfsm`
- added raw output containing `Description` field and the corresponding .yaml file
- updated existing .yaml files to account for the newly added field

## Testing
- ran `invoke tests` locally and all tests passed


mjbear:
Hi @alex4lbin
Thank you for your PR.

There are a couple of updates I'd like to make as a maintainer. It seems like this PR currently may not allow maintainer edits. Could you check that setting?

alex4lbin:
Hi @mjbear
Looks like I accidentally disabled it. Should be fixed now.

